PR #31, opened 2016-05-06
Link to latest draft to README via GutHub pages

adrianhopebailie:


domenic:
There's already a link at the top of the page.

PR #42, opened 2017-02-09
Prefer dictionary to boolean parameters, and other advice related to booleans and dictionaries.

### LINKED ISSUE #28: Prefer dictionary over a boolean parameter
We should have this as advice. And we should note that boolean dictionary members need to default to false, although this would be redundant with IDL.

This ensures the API is:
- More easily extensible
- Less ambiguous when used (what is true?)

dbaron:
This fixes #28, although I'm a little worried that it crams too much advice into a single section.  Thoughts?

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -212,6 +212,40 @@ examples that violate the above rules are <code>XMLHttpRequest</code> and
 initialisms, even if they are repeated.
 </div>
 
+<h3 id="prefer-dict-to-bool">Prefer dictionary parameters over boolean parameters or other unreadable parameters</h3>
+
+APIs should generally prefer dictionary parameters
+(with named booleans in the dictionary)
+over boolean parameters.
+This makes the code that calls the API
+<a href="https://ariya.io/2011/08/hall-of-api-shame-boolean-trap">much more readable</a>.
+It also makes the API more extensible in the future,
+particularly if multiple booleans are needed.
+
+<p class="example">For example,
+<code highlight="js">new Event("exampleevent", { bubbles: true, cancelable: false})</code>
+is much more readable than
+<code highlight="js">new Event("exampleevent", true, false)</code>.
+
+Furthermore,
+the booleans in dictionaries need to be designed so that they all default to false.
+If booleans default to true, then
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -212,6 +212,32 @@ examples that violate the above rules are <code>XMLHttpRequest</code> and
 initialisms, even if they are repeated.
 </div>
 
+<h3 id="prefer-dict-to-bool">Prefer dictionary parameters over boolean parameters or other unreadable parameters</h3>
+
+APIs should generally prefer dictionary parameters over booleans.
```
domenic: "Over boolean parameters" is helpful since we actually keep using booleans inside the dictionary :)

index.bs
```diff
@@ -212,6 +212,32 @@ examples that violate the above rules are <code>XMLHttpRequest</code> and
...
+<code highlight="js">new Event("exampleevent", { bubbles: true, cancelable: false})</code>
+is much more readable than
+<code highlight="js">new Event("exampleevent", true, false)</code>.
+
+Furthermore,
+the booleans in dictionaries need to be designed so that they all default to false.
+If booleans default to true, then
+<a href="https://lists.w3.org/Archives/Public/public-script-coord/2013OctDec/0302.html">users of the API will find unexpected JavaScript behavior</a> since <code highlight="js">{ passive: false }</code> and <code highlight="js">{ passive: undefined }</code> will produce different results.
+But at the same time, it's important to avoid naming booleans in negative ways,
+because then code will have confusing double-negatives.
```
domenic: Hmm. These are obviously in conflict. An example of a workaround might help.

index.bs
dbaron: Replaced with: APIs should generally prefer dictionary parameters (with named booleans in the dictionary) over boolean parameters.

index.bs
dbaron: Added: These pieces of advice may sometimes conflict, but the conflict can be avoided by using opposite words without negatives, such as “repeat” versus “once”, “isolate” versus “connect”, or “private” versus “public”.

PR #43, opened 2017-02-09
Add advice for event handler attributes. Fix #29

### LINKED ISSUE #29: Clarify if event handler attributes need definition (and if so, under what conditions)
For more context, see https://github.com/w3c/media-source/issues/66 for more context, especially the chain of comments up to the reply by @domenic (https://github.com/w3c/media-source/issues/66#issuecomment-215540130).

Under what conditions should event handler attributes be defined in a spec? Should they be for only "declarative" objects (like `<video>`), or for any object that inherits from EventTarget, or some other conditions?

Note that both MSE and EME specs currently operate under the assumption that event handlers only need definition for declarative objects. For example, "MediaSource.onsourceopen" is not defined, since MediaSource objects can only be created by script. Script must explicitly listen for "sourceopen" event firing on a MediaSource to handle that event.

@ddorwin, @foolip, @jdsmith3000, @paulbrucecotton for getting updates on the status of this issue.

travisleithead:


index.bs
```diff
@@ -246,6 +247,30 @@ For example,
...
+
+Because there are two techniques for registering events on objects inheriting from `EventTarget`,
+authors may be tempted to omit the corresponding event handler IDL attributes. They may assume that
+event handler IDL attributes are a legacy registration technique or are simply not needed given that
+`addEventListener` is available as an alternative. However, it is important to continue to define
+event handler IDL attributes because:
+
+* they preserve consistency in the platform (e.g., several media-related event handler IDL
+    attributes are defined on the `Window` and `Document` interfaces despite the fact that
+    implementations will never dispatch those events to those objects)
```
domenic: I think this parenthetical is a bit unrelated to the two bullet points.

Instead I'd suggest moving it to a separate paragraph afterward. Maybe something like:

> Note that for HTML and SVG elements, it is traditional to add the event handler IDL attributes to `{{GlobalEventHandlers}}`.

index.bs
```diff
@@ -246,6 +247,30 @@ For example,
 is more readable than
 <code highlight="js">window.scrollBy(50, 0)</code>.
 
+<h3 id="always-add-event-handlers">Always add event handler attributes</h3>
+
+For an object that inherits from `EventTarget`, there are two techniques available for registering
```
domenic: Should be `{{EventTarget}}`

index.bs
```diff
@@ -246,6 +247,30 @@ For example,
...
+an event handler (e.g., an event named "somethingchanged"):
+
+1. `addEventListener` which allows authors to register for the event using the event's name
+    (i.e., <code highlight="js">someobject.addEventListener("somethingchanged", myhandler)</code>) and
+2. `onsomethingchanged` IDL attributes which allow one event handler to be directly assigned to the
+    object (i.e., `someobject.onsomethingchanged`).
+
+Because there are two techniques for registering events on objects inheriting from `EventTarget`,
+authors may be tempted to omit the corresponding event handler IDL attributes. They may assume that
```
domenic: `<a>event handler IDL attributes</a>`

index.bs
```diff
@@ -246,6 +247,30 @@ For example,
 is more readable than
 <code highlight="js">window.scrollBy(50, 0)</code>.
 
+<h3 id="always-add-event-handlers">Always add event handler attributes</h3>
+
+For an object that inherits from `EventTarget`, there are two techniques available for registering
+an event handler (e.g., an event named "somethingchanged"):
+
+1. `addEventListener` which allows authors to register for the event using the event's name
```
domenic: `{{EventTarget/addEventListener()}}` here and below.

travisleithead:
Thanks for the feedback @Domenic. (I should know better about the linking--I do it all the time in the HTML spec... :-)

domenic:
Thanks for doing this! It's been on my to-do list for months and months.

PR #45, opened 2017-02-10
Use bikeshed's syntax highlighting throughout.

dbaron:
Set the default highlight language to js, remove some use of explicit highlight="js", and add markup to annotate the things that are not JS (because they're HTML, WebIDL, or event names).

Note that it would be possible to go in the other direction and use highlight="js" everywhere.  The editing errors resulting from that are a little less serious (no highlight rather than incorrect highlight), perhaps, but it's a decent amount more markup.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -78,7 +78,7 @@ strategies to ossify, preventing improvement in one of the most dynamic areas of
 virtual machine technology.
 
 In particular, this means that you can't expose any API that acts as a weak reference, e.g. with a
-property that becomes <code>null</code> once garbage collection runs. Such freeing of memory must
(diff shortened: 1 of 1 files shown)


dbaron:
The alternative approach is https://github.com/dbaron/design-principles/commit/c9692383994dae2b45ca8225aa374e003c6a3d7e ; I'm actually thinking maybe that way would be better.  Thoughts?

domenic:
I think I like the explicit version better. Especially since some cases should probably be Bikeshed autolinks anyway (e.g. `{{HTMLHRElement}}` instead of `<code highlight="js">HTMLHRElement</code>`), so the need for it goes down if we do things properly.

dbaron:
Updated with the explicit version (and merged).

PR #47, opened 2017-02-11
Describe tradeoffs between live and static objects.

### LINKED ISSUE #34: should discuss reasons to provide live vs. static objects
The design principles document should probably talk about the reasons to have live vs. static objects in API design.  (The classic example of this is <code>getElementsByTagName()</code> returning a live <code>NodeList</code> that changes over time as the document changes vs. <code>querySelectorAll()</code> returning a static <code>NodeList</code> that represents the result of the query at the time of the function call.)

Some initial thoughts:

Advantages of live objects include:
- performance when the things provided by the object are expensive to compute up front (e.g., <code>getComputedStyle</code>)
- providing a sensible API when the object only exists to "namespace" a group of related functionality (but, beyond that, is really just an extension of the object it's attached to) (at a stretch, maybe <code>window.history</code> is an example here?)
- providing a sensible API when the object needs to have setters that mutate the state that it represents (e.g., <code>window.indexedDB</code>)

Advantages of static objects include:
- performance when the things provided by the object are expensive to keep up-to-date for the lifetime of the object (e.g., <code>querySelectorAll</code>)
- matching expectations of API users who expect the result of a method not to change under them

I'm sure there's a good bit more to list here, though, and I'm not particularly confident that the points above are in sensible categories.

It's certainly then worth linking to [Attributes should behave like data properties](https://w3ctag.github.io/design-principles/#attributes-like-data) to point out that static objects that can change over time need to be the result of methods rather than attributes.

(It may also be worth discussing static vs. frozen.)

dbaron:
Fixes #34.

This could use some substantive review to see if others agree with this advice, and with how I've described things.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -130,6 +130,70 @@ much like a data property as possible. Specific guidance in this regard includes
     true. (This will not always be the case, e.g. if a normalization or type conversion step is
     necessary, but should be held as a goal for normal code paths.)
 
+<h3 id="live-vs-static">Consider whether objects should be live or static</h3>
+
+Objects returned from functions, attribute getters, etc.,
+can either be live or static.
+A <dfn>live object</dfn> is one that continues to reflect
+changes made after it was returned to the caller.
+A <dfn>static object</dfn> is one that reflects
+the state at the time it was returned.
+
+Objects that are the way state is mutated are generally live.
+For example, DOM nodes are returned as live objects,
+since they are the API through which attributes are set and other changes
+to the DOM are made.
+They also reflect changes to the DOM made in other ways
+(such as through user interaction with forms).
+
+Objects that represent a collection
+that might change over time
+(and that are not the way state is mutated)
+should generally be returned as static objects.
+This is because it is confusing to users of the API when
+a collection changes while being iterated.
+Because of this,
+it is generally considered a mistake that methods like
+{{Document/getElementsByTagName()}} return live objects;
+{{ParentNode/querySelectorAll()}} was made to return static objects
+as a result of this experience.
+On the other hand, even though {{URLSearchParams}} represents a collection,
+it should be live because the collection is mutated through that object.
+
+Note:  It's possible that some of this advice should be reconsidered
+for <a>maplike</a> and <a>setlike</a> types,
+where iterators have reasonable behavior
+for mutation that happens during iteration.
+This point likely needs further discussion,
+and perhaps further experience of use of these types.
+
+It's also worth considering the implications of having
+live versus static objects for the speed of implementations of the API.
+When the data needed by an object are expensive to compute up-front,
+there is an advantage for that object to be live so that the results
+can be computed lazily, such as for {{Window/getComputedStyle()}}.
+On the other hand,
+if the data needed by an object are expensive to keep up-to-date,
+such as for the {{NodeList}} returned from {{ParentNode/querySelectorAll()}},
+then providing a static object avoids
+having to keep the object updated until it is garbage collected
+(which may be substantially after its last use).
+
+Likewise, the choice of live versus static objects
+can influence the memory use of an API.
+If each call of a method returns a new static object,
+and the objects are large,
+then substantial amounts of memory can be wasted
+until the next garbage collection.
+
+The choice of whether an object is live or static may also
+influence whether it should be returned
+from an attribute getter or from a method.
+See [[#attributes-like-data]].
+In particular,
+if a result that changes frequently is returned as a static object,
+it should probably be returned from a method rather than an attribute getter.
+
 <h3 id="casing-rules">Use casing rules consistent with existing APIs</h3>
 
 Although they haven't always been uniformly followed, through the history of web platform API


index.bs
```diff
@@ -130,6 +130,71 @@ much like a data property as possible. Specific guidance in this regard includes
...
+A <dfn>static object</dfn> is one that reflects
+the state at the time it was returned.
+
+Objects that are the way state is mutated are generally live.
+For example, DOM nodes are returned as live objects,
+since they are the API through which attributes are set and other changes
+to the DOM are made.
+They also reflect changes to the DOM made in other ways
+(such as through user interaction with forms).
+Even though the nodes themselves are live,
```
domenic: This sentence feels a bit out of place to me, but maybe it is clearing up real potential confusion... not sure.

index.bs
```diff
@@ -130,6 +130,71 @@ much like a data property as possible. Specific guidance in this regard includes
...
+This is because it is confusing to users of the API when
+a collection changes while being iterated.
+Because of this,
+it is generally considered a mistake that methods like
+{{Document/getElementsByTagName()}} return live objects;
+{{ParentNode/querySelectorAll()}} was made to return static objects
+as a result of this experience.
+
+Note:  It's possible that some of this advice should be reconsidered
+for <a>Iterable</a>, <a>Maplike</a>, and <a>Setlike</a> types
```
domenic: Nit: I don't think these should be capitalized.

index.bs
```diff
@@ -130,6 +130,71 @@ much like a data property as possible. Specific guidance in this regard includes
...
+which can be iterated using iterators that have
+well-defined behavior for mutation that happens during iteration.
+(Is this true for all three?)
```
domenic: It's certainly well-defined. Both for these and for cases like live NodeLists. I think the issue is that it might be confusing to deal with, regardless.

index.bs
```diff
@@ -130,6 +130,71 @@ much like a data property as possible. Specific guidance in this regard includes
...
+This point likely needs further discussion,
+and perhaps further experience of use of these types.
+
```
domenic: An interesting case to mention might be that the former advice seems to override the latter. For example, for URLSearchParams, which is live but iterable.

index.bs
```diff
@@ -130,6 +130,71 @@ much like a data property as possible. Specific guidance in this regard includes
...
+It's also worth considering the implications of having
+live versus static objects for the speed of implementations of the API.
+When the data needed by an object are expensive to compute up-front,
+there is an advantage for that object to be live so that the results
+can be computed lazily, such as for {{Window/getComputedStyle()}}.
+On the other hand,
+if the data needed by an object are expensive to keep up-to-date,
+such as for {{ParentNode/querySelectorAll()}},
```
domenic: Editorial: it's a little unclear what the object is here, so I'd say "for the NodeList returned by querySelectorAll()"

index.bs
```diff
@@ -130,6 +130,71 @@ much like a data property as possible. Specific guidance in this regard includes
...
+until the next garbage collection.
+
+The choice of whether an object is live or static may also
+influence whether it should be returned
+from an attribute getter or from a method.
+See [[#attributes-like-data]].
+In particular,
+if a result that changes frequently is returned as a static object,
+it should probably be returned from a method rather than an attribute getter.
+
```
domenic: Maybe not as part of this PR or even this section, but a common point of confusion is the case of "read only lists" (where only the implementation can change them). Currently we tend to recommend people return static snapshots as arrays (for methods, when they change frequently) or frozen arrays (for attributes, when they change infrequently). It's not the best solution, compared to some kind of real read-only array type, but we tend to be going in that direction. Notably we are moving away from creating various fake-array types like NodeList or StyleSheetList for this purpose.

index.bs
dbaron: Er, I probably should have said reasonable rather than well-defined.  @bzbarsky was talking yesterday about how iteration of ES maps is defined by describing the map as an array (in the spec, not in implementations) so that things added get added to the end so that in-progress iteration will find them, and so that elements that are removed leave conceptual "holes" so as not to mess up in-progress iteration.  It wasn't 100% clear to me whether all of this definition applies to maplike as well (as opposed to actual maps).

But if there's a widely expected behavior for mutation while iterating that isn't completely crazy and doesn't require developers to figure out how to adjust an index (or to make sure to iterate from length-1 to 0 and only touch elements already reached), it seems like it might be a little more reasonable to expose that behavior, particularly if (as @bzbarsky pointed out yesterday) iteration is possible *only* through iterators and not through indexed getters.

index.bs
dbaron: Yeah, I was thinking of having a separate paragraph about how live and static objects can basically point to each other arbitrarily, and live vs. static is a distinction made about a single object rather than a part of the graph, but decided that was too much, and sort of half tacked it on to the end here.  Maybe I should just drop it.  Or maybe expand it?

index.bs
dbaron: Which advice overrides which?

annevk: I think the point is that if you have an object with a method that returns a collection, that collection should be static. But if you have an object that represents the collection, its iterator should be live.

index.bs
travisleithead: I would drop it. It seems to go without saying that I can create arrays of arbitrary objects and that this wrapper in no way changes my expectations about the live-ness of the things in contains.

index.bs
bzbarsky: Maplikes are implemented on top of `Map`, so they do in fact get the `Map` iteration behavior.  Same for setlikes and `Set`.  indexed-getter things use the `Array` iterator, which has all sorts of weirdness on mutation (it just increments an index and gets things at that value; the index is NOT changed on mutation).  The `iterable` iterator acts more like the `Array` iterator than `Map` or `Set`.

index.bs
domenic: What I meant was that objects that are collections that are supposed to be mutated need to be live. URLSearchParams are such an example.

dbaron:
Took a few days to get back to this one, but I believe I've now updated this to address the comments above.

dbaron:
@domenic do you prefer to merge the pull requests yourself, or do you mind if I do it at this point?

domenic:
Looks great, thanks for the ping!

PR #58, opened 2017-04-28
Add advice for event design.

### LINKED ISSUE #12: Events should be used for notification of actions, not actions themselves
@annevk sums this up [on public-webapps](http://lists.w3.org/Archives/Public/public-webapps/2014AprJun/0510.html), pointing out that dispatching a ClipboardEvent should not actually copy things to the user's clipboard.

He might add this advice to the DOM standard itself. But if he doesn't, then we should be sure to point it out at some point.

slightlyoff:
Fixes #12

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -342,21 +342,31 @@ So, if the object inherits from {{EventTarget}}, add a corresponding
 [=event handler IDL attributes=] on the {{GlobalEventHandlers}} interface, instead of directly on
 the relevant element interface(s).</p>
 
+<h2 id="event-design">Event Design</h2>
+
+<h3 id="events-are-for-notification">Events are for notification</h3>
+
+Try to design DOM events to deliver after-the-fact notifications of changes. It may be tempting to try to trigger side-effects from the action of <code>dispatchEvent</code>, but in general this is <a href="https://lists.w3.org/Archives/Public/public-webapps/2014AprJun/0510.html">discouraged</a>.
+
+<h3 id="favor-asynchronous-events">Favor asynchornous events</h3>
+
+A few events in the platform are specified to dispatch synchronously. These events cause problems for engines and performance issues in applications. If you feel you may ned a synchronous event in your design, please reconsider and ask the TAG for help in redesigning the API.
+
+<h3 id="state-and-subclassing">State and Event subclasses</h3>
+
+It's tempting to create subclasses of `Event` for all event types. This is frequently overkill. Consider subclassing `Event` when adding unique methods and large amounts of state. In all other cases, using a "vanilla" event with state captured in the `.target` object.
+
 <h2 id="types-and-units">Types and Units</h2>
 
 <h3 id="numeric-types">Use numeric types appropriately</h3>
 
-[[!WEBIDL]], because of its OMG IDL heritage, has a plethora of numeric types. However, it is very
+[[!WEBIDL]] contains many numeric types. However, it is very
 rare that its more specific ones are actually appropriate.
 
-Remember that JavaScript itself only has one numeric type, Number: IEEE 754 double-precision
-floating point, including ±0, ±Infinity, and NaN (although thankfully only one). The Web IDL
-"types" are simply coercion rules that apply when accepting an argument or triggering a setter. For
-example, a Web IDL <code>unsigned short</code> roughly says: "when someone passes this as an
-argument, take it modulo 65535 before doing any further processing". That is very rarely a useful
-thing to do.
+JavaScript has only one numeric type, Number: IEEE 754 double-precision
+floating point, including ±0, ±Infinity, and NaN (although thankfully only one). The Web IDL "types" are coercion rules that apply when accepting an argument or triggering a setter. For example, a Web IDL <code>unsigned short</code> roughly says: "when someone passes this as an argument, take it modulo 65535 before doing any further processing". That is very rarely a useful thing to do.
 
-Instead, you probably want to stick with one of:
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -338,21 +338,31 @@ So, if the object inherits from {{EventTarget}}, add a corresponding
 [=event handler IDL attributes=] on the {{GlobalEventHandlers}} interface, instead of directly on
 the relevant element interface(s).</p>
 
+<h2 id="event-design">Event Design</h2>
+
+<h3 id="events-are-for-notification">Events are for notification</h3>
+
+Try to design DOM events to deliver after-the-fact notifications of changes. It may be tempting to try to trigger side-effects from the action of <code>dispatchEvent</code>, but in general this is <a href="https://lists.w3.org/Archives/Public/public-webapps/2014AprJun/0510.html">discouraged</a>.
+
+<h3 id="favor-asynchronous-events">Favor asynchornous events</h3>
```
dbaron: s/chorn/chron/

index.bs
```diff
@@ -338,21 +338,31 @@ So, if the object inherits from {{EventTarget}}, add a corresponding
...
+
+A few events in the platform are specified to dispatch synchronously. These events cause problems for engines and performance issues in applications. If you feel you may ned a synchronous event in your design, please reconsider and ask the TAG for help in redesigning the API.
```
dbaron: s/ned/need/

Or actually even "may ned" -> "need".


I also wonder if it's worth mentioning the pattern that IntersectionObserver uses here, although I'm not sure what to call it.

index.bs
```diff
@@ -338,21 +338,31 @@ So, if the object inherits from {{EventTarget}}, add a corresponding
...
+
+<h3 id="state-and-subclassing">State and Event subclasses</h3>
+
+It's tempting to create subclasses of `Event` for all event types. This is frequently overkill. Consider subclassing `Event` when adding unique methods and large amounts of state. In all other cases, using a "vanilla" event with state captured in the `.target` object.
```
dbaron: probably "overkill" -> "unnecessary".

index.bs
travisleithead: Do we want to go into the why these are problems? E.g., synchronous events cause re-entrancy (e.g., in a focus event handler another element is focused...) which can cause unexpected results in author code and are tricky to spec correctly. Also Sync mutation events were triggered at a time in which an operation of the DOM was partially-complete, resulting in reentrancy, breaking internal invariants, and causing platform reliability issues.

index.bs
```diff
@@ -338,21 +338,31 @@ So, if the object inherits from {{EventTarget}}, add a corresponding
 [=event handler IDL attributes=] on the {{GlobalEventHandlers}} interface, instead of directly on
 the relevant element interface(s).</p>
```
travisleithead: Wondering if the previous section should be relocated down into your new section...

index.bs
```diff
@@ -338,21 +338,31 @@ So, if the object inherits from {{EventTarget}}, add a corresponding
...
 rare that its more specific ones are actually appropriate.
 
-Remember that JavaScript itself only has one numeric type, Number: IEEE 754 double-precision
-floating point, including ±0, ±Infinity, and NaN (although thankfully only one). The Web IDL
-"types" are simply coercion rules that apply when accepting an argument or triggering a setter. For
-example, a Web IDL <code>unsigned short</code> roughly says: "when someone passes this as an
-argument, take it modulo 65535 before doing any further processing". That is very rarely a useful
-thing to do.
+JavaScript has only one numeric type, Number: IEEE 754 double-precision
+floating point, including ±0, ±Infinity, and NaN (although thankfully only one). The Web IDL "types" are coercion rules that apply when accepting an argument or triggering a setter. For example, a Web IDL <code>unsigned short</code> roughly says: "when someone passes this as an argument, take it modulo 65535 before doing any further processing". That is very rarely a useful thing to do.
```
annevk: Aren't the other NaNs exposed through typed array objects?

index.bs
```diff
@@ -338,21 +338,31 @@ So, if the object inherits from {{EventTarget}}, add a corresponding
 [=event handler IDL attributes=] on the {{GlobalEventHandlers}} interface, instead of directly on
 the relevant element interface(s).</p>
 
+<h2 id="event-design">Event Design</h2>
+
+<h3 id="events-are-for-notification">Events are for notification</h3>
+
+Try to design DOM events to deliver after-the-fact notifications of changes. It may be tempting to try to trigger side-effects from the action of <code>dispatchEvent</code>, but in general this is <a href="https://lists.w3.org/Archives/Public/public-webapps/2014AprJun/0510.html">discouraged</a>.
```
annevk: It's more than discouraged as it requires changing the DOM Standard each time it's done.

index.bs
slightlyoff: Done

index.bs
slightlyoff: Done

index.bs
slightlyoff: Done

index.bs
slightlyoff: Agree; moved

index.bs
slightlyoff: Added a note to this effect

index.bs
slightlyoff: Only by comparison operations, right?

index.bs
slightlyoff: Will update. Thanks.

PR #60, opened 2017-04-28
Add links to advice on related topic of specification writing.

dbaron:
This addresses a part of https://github.com/w3ctag/design-reviews/issues/136 .

This isn't *obviously* a thing that belongs in this document, but I tend to think it may be useful.


dbaron:
I brought this up to a slightly larger subset of the TAG and they suggested moving to the end of the document but were otherwise happy with it, so I've updated the PR to do that.

PR #62, opened 2017-04-29
Suggest use of promises, and point to promises-guide.

dbaron:
In discussion before dinner we realized that the design-principles document didn't point to the promises-guide.  This adds a brief section on when promises may be needed, and then points to the promises guide for more.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -315,6 +315,33 @@ For example,
 is more readable than
 <code highlight="js">window.scrollBy(50, 0)</code>.
 
+<h3 id="promises">Design asynchronous APIs using Promises</h3>
+
+Asynchronous APIs should generally be designed using promises
+rather than callback functions.
+This is the pattern that we've settled on for the Web platform,
+and having APIs consistently use promises means that the APIs are
+easier to use together (such as by chaining promises).
+This pattern also tends to produce cleaner code than the use
+of APIs with callback functions.
+
+Furthermore, you should carefully consider whether an API might need
+to be asynchronous before making it a synchronous API.
+An API might need to be asynchronous if:
+  * some implementations may (in at least some cases) wish to prompt the user
+    or ask the user for permission before allowing use of the API,
+  * implementations may need to consider information that might be stored
+    on disk in order to compute the result,
+  * implementations may need to interact with the network before returning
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -315,6 +315,30 @@ For example,
 is more readable than
 <code highlight="js">window.scrollBy(50, 0)</code>.
 
+<h3 id="promises">Use Promises</h3>
+
+Asynchronous APIs should generally be designed using promises
+rather than callback functions,
+because APIs using promises are generally easier to use
+than APIs using callback functions.
```
domenic: IMO the more important reason is because this is the pattern we've settled on for the web platform as a whole. Consistency is important, and you can compose more easily with other web platform APIs (e.g. creating promise chains).

index.bs
```diff
@@ -315,6 +315,30 @@ For example,
 is more readable than
 <code highlight="js">window.scrollBy(50, 0)</code>.
 
+<h3 id="promises">Use Promises</h3>
```
domenic: I'd title this "Designing asynchronous APIs" perhaps since a large chunk of it is about the general question of sync-vs.-async.

index.bs
```diff
@@ -315,6 +315,30 @@ For example,
...
+to be asynchronous before making it a synchronous API.
+An API might need to be asynchronous if:
+  * some implementations may (in at least some cases) wish to prompt the user
+    or ask the user for permission before allowing use of the API,
+  * implementations may need to consider information that might be stored
+    on disk in order to compute the result,
+  * implementations may need to interact with the network before returning
+    the result, or
+  * implementations may wish to do work on another thread before returning
+    the result.
```
domenic: Maybe add "or communicate with an out-of-process subsystem" since that case is pretty common these days.

index.bs
dbaron: I'm almost tempted to make it "Design asynchronous APIs" instead 😄.  I guess I'll stick with "Designing", though.

dbaron: Actually, I think what fits best with the rest of the headers is "Design asynchronous APIs using Promises", which appropriately falls between the other two ideas in terms of how much of a command it is.

PR #66, opened 2017-05-04
Switch to container-based virtualization for Travis CI.

dbaron:
This should have much faster boot time, which should be the main time
constraint.  See https://docs.travis-ci.com/user/ci-environment/ .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 script: bash ./deploy-gh-pages.sh
 env:
   global:


dbaron:
The execution was a bit slower, compare [build without](https://travis-ci.org/w3ctag/design-principles/builds/228926477) that took 21 seconds, versus a [build with](https://travis-ci.org/w3ctag/design-principles/builds/228928448) that took 39 seconds.  But I think it did start a good bit faster, so I think it's a sensible switch.

PR #75, opened 2017-08-23
Describe when features should be limited to secure contexts.

### LINKED ISSUE #32: How should Secure Context-only APIs fail in insecure contexts?
Via an issue from @sleevi and @mikewest, it seems we should have uniform advice for spec authors (and an IDL recipe) for how to handle cases where an API is only partially usable, as is the case for Secure Context-only APIs.

@mikewest notes that the platform's behavior is not uniform today. We should catalog and attempt to drive uniformity.

/cc @domenic @torgo

dbaron:
This pull request is intended to fix #32.

I expect this text to be somewhat controversial and need a good bit of review and polishing.  However, it seemed like a good way to start would be to write something down (thanks to @mikewest for the [reminder](https://twitter.com/mikewest/status/900090659641491456)), and we can try to make progress from there.

I'd also note that the place I put the new section in the document didn't feel particularly obvious, and is probably worth thinking about during the review.

I'll try to remember to address feedback as additional commits, with the intention of squashing the later commits in at the end.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -345,6 +345,65 @@ and on when to use promises and when not to use promises,
 see <strong><a href="https://www.w3.org/2001/tag/doc/promises-guide">Writing
 Promise-Using Specifications</a></strong>.
 
+<h3 id="secure-context">Limit new and distinct APIs to secure contexts</h3>
+
+Many new APIs should be available only in
+<a href="https://w3c.github.io/webappsec-secure-contexts/">secure contexts</a>.
+This restriction exists for two reasons.
+First, it helps encourage Web content and applications
+to migrate to secure contexts.
+Second, it can restrict new APIs where authentication, integrity, or confidentiality
+are important to prevent substantial increases to the privacy or security risks of using the Web.
+For more detail, see the W3C TAG Finding on
+<a href="https://www.w3.org/2001/tag/doc/web-https">Securing the Web</a>.
+
+When the new feature is defined in
+<a href="https://heycam.github.io/webidl/">WebIDL</a>,
+specification authors can limit a feature to secure contexts
+by marking new interfaces and methods with the
+<code>[<a href="https://w3c.github.io/webappsec-secure-contexts/#integration-idl">SecureContext</a>]</code> extended attribute
+on interfaces, namespaces, or their members (such as methods and attributes).
+Similar ways of marking features as limited to secure contexts should be added
+to other points where the Web platform is extended over time
+(for example, the definition of a new CSS property).
+However, for some times of extension points (e.g., new DOM events),
+limitation to secure contexts should just
+be defined in normative prose in the specification.
+
+When deciding whether a feature should be limited to secure contexts,
+we suggest considering the following factors:
+
+ : Does the feature have a clear and understandable boundary?
+
+ :: A feature that is limited to secure contexts
+    should be recognizable by the developers who use it as
+    distinct from features that are available in non-secure contexts.
+    This will reduce developer confusion about where the boundaries are.
+    For example, a new CSS property is a distinct feature,
+    whereas the ability to omit the commas in the CSS ''rgb()'' function is not.
+    New features that are clearly distinct
+    should generally be limited to secure contexts.
+    However, there are many cases where whether a feature is clearly distinct
+    is a matter of judgment.
+
+
+ : Does the feature depend on being in a secure context?
+
+ :: If a feature depends on
+    the expectations of authentication, integrity, or confidentiality
+    that are met only in secure contexts,
+    then it should be limited to secure contexts,
+    even if it is not a clearly distinct feature as described in the previous point.
+    For example, a feature that communicates with USB devices
+    if those devices have allowed
+    Web content from the site's origin
+    to communicate those USB devices
+    depends on the authentication of the origin
+    and the integrity of the data
+    sent to the USB device,
+    since sending untrusted data to a USB device could damage that device
+    or compromise computers that the device connects to.
+
 <h2 id="event-design">Event Design</h2>
 
 <h3 id="always-add-event-handlers">Always add event handler attributes</h3>


index.bs
```diff
@@ -345,6 +345,65 @@ and on when to use promises and when not to use promises,
...
+First, it helps encourage Web content and applications
+to migrate to secure contexts.
+Second, it can restrict new APIs where authentication, integrity, or confidentiality
+are important to prevent substantial increases to the privacy or security risks of using the Web.
+For more detail, see the W3C TAG Finding on
+<a href="https://www.w3.org/2001/tag/doc/web-https">Securing the Web</a>.
+
+When the new feature is defined in
+<a href="https://heycam.github.io/webidl/">WebIDL</a>,
+specification authors can limit a feature to secure contexts
```
annevk: can or should?

index.bs
```diff
@@ -345,6 +345,65 @@ and on when to use promises and when not to use promises,
...
+by marking new interfaces and methods with the
+<code>[<a href="https://w3c.github.io/webappsec-secure-contexts/#integration-idl">SecureContext</a>]</code> extended attribute
+on interfaces, namespaces, or their members (such as methods and attributes).
+Similar ways of marking features as limited to secure contexts should be added
+to other points where the Web platform is extended over time
+(for example, the definition of a new CSS property).
+However, for some times of extension points (e.g., new DOM events),
```
annevk: I can't parse this sentence.

index.bs
```diff
@@ -345,6 +345,65 @@ and on when to use promises and when not to use promises,
...
+limitation to secure contexts should just
+be defined in normative prose in the specification.
```
annevk: If the new event comes with a new interface it's quite easy to restrict it though. Maybe you should clarify this by talking about "dispatching an event" instead.

index.bs
```diff
@@ -345,6 +345,65 @@ and on when to use promises and when not to use promises,
...
+    However, there are many cases where whether a feature is clearly distinct
+    is a matter of judgment.
+
+
+ : Does the feature depend on being in a secure context?
+
+ :: If a feature depends on
+    the expectations of authentication, integrity, or confidentiality
+    that are met only in secure contexts,
+    then it should be limited to secure contexts,
```
annevk: must?

index.bs
```diff
@@ -345,6 +345,65 @@ and on when to use promises and when not to use promises,
...
 
+<h3 id="secure-context">Limit new and distinct APIs to secure contexts</h3>
+
+Many new APIs should be available only in
+<a href="https://w3c.github.io/webappsec-secure-contexts/">secure contexts</a>.
+This restriction exists for two reasons.
+First, it helps encourage Web content and applications
+to migrate to secure contexts.
+Second, it can restrict new APIs where authentication, integrity, or confidentiality
+are important to prevent substantial increases to the privacy or security risks of using the Web.
```
mikewest: I would suggest reversing these. At least from Chrome's perspective, the latter has been the the overriding concern internally.

annevk: But the first is the main goal. If you make the second the main goal, it's easier for folks to weasel their way into an exception.

index.bs
jonathanKingston: s/times/types/ ?

index.bs
dbaron: I think this one actually meant can; this was a description of facts, not a conformance requirement.

dbaron:
OK, I think I've addressed the feedback so far.  I think it's worth another round of review at this point.

I worry a little bit that I may have come down a little too hard in terms of requiring that all parts of a new feature be hidden.  Maybe it's ok to just hide the major pieces (and primary detection points) such that it's not usable and not detected as present.  But it wasn't obvious to me how to fix that in my current wording...

index.bs
```diff
@@ -345,6 +345,90 @@ and on when to use promises and when not to use promises,
...
+    if those devices have allowed
+    Web content from the site's origin
+    to communicate with those USB devices
+    depends on the authentication of the origin
+    and the integrity of the data
+    sent to the USB device,
+    since sending untrusted data to a USB device could damage that device
+    or compromise computers that the device connects to.
+
+Specification authors can most features defined in
```
annevk: can limit*

travisleithead:
I also think the intro is too heavy-handed regarding requiring **all new features** to be in secure contexts. I think the subsequent sections are actually pretty good, outlining some principles that help shape the decision in a more nuanced way (and tend to disagree with the **all** aspect laid down before).

Perhaps this could be softened by suggesting that a feature **should** be considered for Secure Context by default given the principles [described after that].

From my implementer's hat, I'm particularly sensitive to plumbing yet another **mode** through the platform (speaking as a veteran of IE's document modes). We already have quirks/almost standards/standards mode, and now we'd have Secure Context as well... For APIs exposed to script, I'm OK with having this mode because the "switch" is processed once (when the type system is initialized for a given script engine) and never consulted again. Plumbing the switch into the formatting/layout system or event system is far-less clean and has performance overhead.

Applying this to CSS also seems a little questionable to me.

I understand the desire to put new Houdini features behind secure context. Those features tend to rely on JavaScript APIs for initialization. If the JS APIs are put behind secure context, then surely that cuts-off access to the related CSS properties? 

Another example: for new conceptual CSS layout types, would we really want to prevent non-HTTPS sites from adopting a new kind of layout? A new layout seems to have no legitimate ties to "security" except that we want to use it as "bait" to switch the web to HTTPs...

Must we recommend that even new CSS properties should have secure context applied? 

There may be a similar argument for events--though events can provide output that might leak sensitive information, and so this argument seems weaker to me.

annevk:
> Must we recommend that even new CSS properties should have secure context applied?

If we can, why not?

What's the rationale for continued support for insecure contexts?

cynthia:
> What's the rationale for continued support for insecure contexts?

While a bit of a corner case - internal/isolated network services, in-development, prototypes come to mind. Working on secure context only features using a local server isn't that user friendly, since (AFAIK, correct me if I am wrong) you can't use acme to get a free certificate.

(Wonder what kind of rules need to be bent to get a CA to issue a certificate that only works on localhost origins..)

plinss:
> On Sep 12, 2017, at 7:27 PM, cynthia <notifications@github.com> wrote:
> While a bit of a corner case - internal/isolated network services, in-development, prototypes come to mind. Working on secure context only features using a local server isn't that user friendly, since (AFAIK, correct me if I am wrong) you can't use acme to get a free certificate.
> 
To get an acme certificate you have to have a real domain name reachable from the internet and either access to setting a DNS TXT record or a publicly reachable HTTP or HTTPS server. It’s doable, but may not be easy for most people behind consumer grade firewalls without access to a real server.

It’s also possible to get acme authorization on a publicly reachable machine, then use that authorization to get a cert on a different machine that isn’t publicly reachable (I do this regularly via my own acmebot). You still have to have a real domain name for the local machine (but that can be faked via a hosts file).
> (Wonder what kind of rules need to be bent to get a CA to issue a certificate that only works on localhost origins..)
> 

Never gonna happen (I hope).

That all said, I believe localhost is (or will soon be) considered a secure context without HTTPS. You can also generate a self-signed certificate with any common/alt name you want and tell your browser to trust it for local testing. So this shouldn’t be a burden for developers.

mikewest:
> > Must we recommend that even new CSS properties should have secure context applied?
>
> If we can, why not?
>
> What's the rationale for continued support for insecure contexts?

FWIW, I'm with Anne. If we can reasonably limit a given feature, we should.

At a minimum, I believe that stance should be our default. We can evaluate arguments on a case-by-case basis when that stance leads to results we're unhappy with. But folks should expect to have to make those arguments.

> That all said, I believe localhost is (or will soon be) considered a secure context without HTTPS.

`http://localhost/` will be considered a secure context in the spec as soon as we can convince IETF folks to accept https://tools.ietf.org/html/draft-west-let-localhost-be-localhost. Chrome's out ahead of this a little bit (https://groups.google.com/a/chromium.org/d/msg/blink-dev/RC9dSw-O3fE/E3_0XaT0BAAJ), and I hope other vendors will do the same.

> You can also generate a self-signed certificate with any common/alt name you want and tell your browser to trust it for local testing. So this shouldn’t be a burden for developers.

Exactly this. Browsers should also do a better job of allowing developers to treat a given origin as "secure enough" for development purposes. Chrome has command-line flags, but we should really embed it into devtools somehow.

martinthomson:
@mikewest, there isn't a lot of visibility of [draft-west-let-localhost-be-localhost](https://tools.ietf.org/html/draft-west-let-localhost-be-localhost).  I think that we're somewhat stalled on that.  I don't know where you were discussing it though.  There are alternatives though on the browser side.  Maybe we should discuss that some more; I've some ideas here that I need to look into first.

mikewest:
@martinthomson: Just finished a [call for adoption in DNSOP](https://www.ietf.org/mail-archive/web/dnsop/current/msg20963.html), which I think was successful though the chairs haven't yet confirmed their view. The main hangup is the fork in https://tools.ietf.org/html/draft-west-let-localhost-be-localhost-06#section-4.2. Strong opinions on both sides. *shrug*

martinthomson:
Thanks for the pointer.  I predict that it will be adopted but that you will eventually regret ever attempting this.  It will be an RFC worth having though.  Congratudolences.

triblondon:
At first I though that we in the TAG were being congratudoliated, but it was @mikewest that had that honour.

index.bs
```diff
@@ -345,6 +345,81 @@ and on when to use promises and when not to use promises,
...
+it is not possible for developers to detect whether a feature is present,
+limiting the feature to secure contexts
+might cause problems
+for libraries that may be used in either secure or non-secure contexts.
+
+If a feature depends on
+the expectations of authentication, integrity, or confidentiality
+that are met only in secure contexts,
+then it must be limited to secure contexts,
+even if the other factors above could justify exposing it in non-secure contexts.
```
mikewest: I don't see it as the feature depending on authentication, integrity, or confidentiality, but instead the feature posing some risk to user privacy or security which is mitigated only by requiring authentication, integrity, and confidentiality. I mean, at some level, _all_ features depend on the page's integrity, right? :)

WDYT about something like "If a feature poses a risk to user privacy or security which can be mitigated by requiring authentication, integrity, and confidentiality, the feature must be limited to secure contexts, ..."?

mikewest:
@dbaron: Friendly ping. We're having some conversations about this internally in Chrome, and a clear position statement from the TAG would be helpful in that discussion. :) /cc @slightlyoff

Anything I can do to help out?

dbaron:
Yeah, I managed to miss the github email from your review comments 6 days ago.  I'd been hoping to (a) polish the text I had so far, which was a bit rough, (b) go through the blink-api-owners thread and now also (c) go through your comments above.   Not going to happen today, but hopefully sometime later this week.

index.bs
```diff
@@ -345,6 +345,81 @@ and on when to use promises and when not to use promises,
...
+
+The most common justification that we expect
+for supporting a new capability in non-secure contexts
+is that the new capability is not recognizably separate
+from an existing feature that is available in non-secure contexts.
+In this case, limiting it to secure contexts could cause
+developer confusion about where the boundaries are.
+We also don't want to increase
+the complexity of implementations of Web technology
+by requiring tests for secure contexts in too many *types* of places.
```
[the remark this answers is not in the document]
dbaron: What I mean by this is that for `[SecureContext]` annotations in WebIDL, the annotation presumably gets stored as a bitflag or similar, and gets checked in a small set of places that implement `[SecureContext]`.  Likewise for CSS properties, engines presumably have a set of data about each property, to which a secure-contexts-only bit could be added, and likewise tested in a small number of places.  But it seems preferable to avoid littering IsSecureContext() tests through the CSS parser (or other language parsers), and I think this preference likely aligns with the previous justification.

index.bs
```diff
@@ -345,6 +345,81 @@ and on when to use promises and when not to use promises,
...
 
+<h3 id="secure-context">Limit new features to secure contexts</h3>
+
+New capabilities added to the Web should be available only in
+<a href="https://w3c.github.io/webappsec-secure-contexts/">secure contexts</a>.
+Exposing them in non-secure contexts
+is discouraged and requires strong justification.
+The TAG is interested in hearing about and discussing cases
+where it is unclear whether exposing the capability
+in non-secure contexts is justifiable.
```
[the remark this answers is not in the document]
dbaron: How about just:
> The TAG is interested in hearing about and working to resolve any cases where exposure in non-secure contexts is being seriously considered.

index.bs
dbaron: To avoid the "poses a risk" in your wording that makes it sound like the feature is problematic, I think I'm going to try:
> If a feature would pose a risk to user privacy or security  without the authentication, integrity, or confidentiality that is present only in secure contexts, then the feature must be limited to secure contexts, ...

index.bs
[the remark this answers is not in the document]
dbaron: Or, simpler:
> The TAG is interested in hearing about cases 
where exposing new features in non-secure contexts is being considered.

(I'm pulling the "working to resolve" because I suspect that much of what we hear about might qualify as "not a new feature".)

dbaron:
I think the above addresses @mikewest's review comments above and most of my own concerns about polish.  I still need to go through the blink-api-owners thread.  (Maybe I'm doing these things in the wrong order...)

dbaron:
So I didn't see a lot in [the blink-api-owners-discuss thread](https://groups.google.com/a/chromium.org/d/topic/blink-api-owners-discuss/Gmxej1Ryj8c/discussion) that seemed relevant to any particulars of the text; it's largely about the high-level tradeoff between cost for developers caused by these changes (and how to measure the risk / amount of that) versus the (not explicitly discussed) risk to both users and developers of using non-secure transport.  It seemed (unless I'm missing something) largely relevant to the question of how strong a recommendation to make, and not to the details of how we make that recommendation.

index.bs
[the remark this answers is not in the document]
mikewest: For WebIDL, Chrome checks when generating bindings for a given context (e.g. everything is wrapped up in the `exposed` checks). For deprecations or places where `[SecureContext]` isn't relevant, we do inline checks at the entry point to the API (e.g. https://cs.chromium.org/chromium/src/third_party/WebKit/Source/modules/geolocation/Geolocation.cpp?rcl=2bd5f03512bcb0b0632366109612ea4e9c4b7ce2&l=220).

I think I agree with the thrust of your comments here, but I still don't really understand what this paragraph is telling feature designers. Does it boil down to "Use `[SecureContext]` when possible?"?

index.bs
```diff
@@ -345,6 +345,81 @@ and on when to use promises and when not to use promises,
...
+they are limited to secure contexts.
+Putting this another way,
+a feature that is limited to secure contexts should, in non-secure contexts,
+be indistinguishable from a feature that is not implemented.
+However, if, for some reason
+(a reason that itself requires serious justification),
+it is not possible for developers to detect whether a feature is present,
+limiting the feature to secure contexts
+might cause problems
+for libraries that may be used in either secure or non-secure contexts.
```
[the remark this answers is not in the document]
dbaron: I moved this into https://github.com/w3ctag/design-principles/pull/82 and revised the text.

index.bs
```diff
@@ -345,6 +345,81 @@ and on when to use promises and when not to use promises,
...
+to secure contexts
+by using the
+<code>[<a href="https://w3c.github.io/webappsec-secure-contexts/#integration-idl">SecureContext</a>]</code> extended attribute
+on interfaces, namespaces, or their members (such as methods and attributes).
+Similar ways of marking features as limited to secure contexts should be added
+to other major points where the Web platform is extended over time
+(for example, the definition of a new CSS property).
+However, for some types of extension points (e.g., dispatching an event),
+limitation to secure contexts should just
+be defined in normative prose in the specification.
```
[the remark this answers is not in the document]
dbaron: I reordered the paragraphs.

index.bs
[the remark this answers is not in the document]
dbaron: I think it boils down to "use `[SecureContext]` and equivalent things for other languages like CSS", and maybe be a little more hesitant about secure context restrictions that would need to be in other places.

ojanvafai:
There's a tradeoff we're making between secure context growth and growth of the web in general. Presumably we all care strongly about both of those things.

Requiring SecureContexts means that the majority of libraries and third party embeds won't use the API because a significant fraction of their clients are on insecure contexts. In some cases, they will use the API with a polyfill fallback, but my experience is that they only do this when the wins are astoundingly significant, which is quite rare. It's not worth the code bloat and implementation complexity for them in these cases.

A significant fraction of the web today is libraries and third party embeds. So, requiring SecureContext for new features does hurt adoption of those features. To the extent that using those features makes web content better (which is presumably why we're shipping them), it hurts the quality of web content.

We can't properly evaluate what we're trading off without data in my opinion. What percentage of domains are secure? What percentage of new domains are secure? For popular libraries and third party embeds, what percentage of their page loads are secure?

index.bs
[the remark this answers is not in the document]
jonathanKingston: Can we make it a requirement for specs to have web platform tests so that all browsers are consistent in how these features are hidden/fail?

Such that *"For deprecations or places where [SecureContext] isn't relevant, we do inline checks at the entry point to the API"* behaves consistently?

travisleithead:
I am still not satisfied with the message here. As I read this, I'm seeing a mixed message and I'm honestly not convinced there's a away to have both messages really make sense. Message 1: *every new feature* must be in secure context. Message 2: there are known legitimate reasons why some new features wouldn't make sense in secure context.

To be clear: I'm fully on board with the stated rationale for putting all new features into secure context: yes getting to 100% encrypted endpoints on the web is the right goal. (Though I do have some misgivings about the whole digital certificate identity trust model system in general--it's not a utopia we are building upon.)

I think we are getting stuck wanting a simple message for spec authors that is not as simple as it seems. If spec authors and implementers take a religious approach to this, we will undoubtedly see the complications that the guideance already cautions us about: `"We also don't want to increase the complexity of implementations of Web technology by requiring tests for secure contexts in too many *types* of places"`. When I asked around on the Edge team, it was noted that with Houdini, adding secure context checks would be even more complicated with new kinds of value types possible through [Custom properties and values](https://drafts.css-houdini.org/css-properties-values-api/) and the [Typed OM](http://www.w3.org/TR/css-typed-om-1/) which would need versioning of internal data structures that we absolutely would not want to do.


Let me posit a new approach to this guideance: instead of laying down an absolute (all new features in secure context or face the wrath of TAG) with noted exceptions, let's get more clear on what the principle behind the exceptions are and use that as the guideance. After dwelling on this for a bit, the principle that I think captures the exceptions is [Coupling](https://en.wikipedia.org/wiki/Coupling_%28computer_programming%29). When a new feature is added to the platform there's an implied degree of coupling that the feature has with the existing platform--that coupling ranges from very loose (e.g., relatively isolated complete features like Web USB/Bluetooth/Web Crypto/IndexedDB, etc.) to relatively tight (e.g., [ParentNode.append](https://dom.spec.whatwg.org/#dom-parentnode-append), [EventListenerOptions](https://dom.spec.whatwg.org/#dictdef-eventlisteneroptions), new extensions to IndexedDB, etc.). I think the general guideance can be summarized as "use secure context for all new features that have loose-coupling to the rest of the platform. Don't use secure context for new features with tight-coupling to existing features that aren't already in secure context." Of course, there's a range of gray in the middle of these end points, but I don't think we're going to escape that, and should just expect that spec authors will need to think a little (and TAG is a helpful resource for them). The case brought up in the proposed guideance (i.e., with new CSS properties vs. new syntax parsing for commas) falls somewhere in the middle. I think some examples really help make the principle clear and we should add them.

I believe an added benefit of this approach is that it reduces some of the potential integration pain and feature-detect issues that will arise as library code blends with local code. I don't think it makes a lot of sense, for example, to put new IndexedDB or DOM convenience APIs into secure contest, requiring permanent feature detection by libraries that want to use them (vs. feature detects for old browsers that don't support them, which detects can eventually be removed once browsers update to support it interoperably--which in theory is much faster than the time it will take to get the web to 100% secure context only). There's also the problem of keeping track of what *parts of a feature* (again coming back to coupling) are available or not in secure context. At least for loosely-coupled features, it's much easier to reason about.

Finally, a nit on the wording: "feature" and "capability" are both used but are relatively undefined. I infer based on the context that a "features" is larger than a "capability" but would prefer to either define them a bit more rigorously, or just stick with one word or the other :)

annevk:
> which would need versioning of internal data structures that we absolutely would not want to do

Can you elaborate on that? I agree that's not desirable, but I don't really see why that would be the case. It would just affect whether there's a JavaScript API exposing the internal data structure. E.g., had we restricted `fetch()`, that would not mean the underlying fetch primitive would have to be versioned somehow.

And if specification authors make incorrect judgments, then implementers, the W3C TAG, and others will surely advice them otherwise. I'd rather have them be conservative though than figure out what coupling means to them.

(I'm also not entirely convinced by the examples. It seems reasonable for new DOM operations to become restricted even if they're just "sugar".)

travisleithead:
>> When I asked around on the Edge team, it was noted that with Houdini, adding secure context checks would be even more complicated with new kinds of value types possible through Custom properties and values and the Typed OM which would need versioning of internal data structures that we absolutely would not want to do.
>
> Can you elaborate on that? I agree that's not desirable, but I don't really see why that would be the case. It would just affect whether there's a JavaScript API exposing the internal data structure. E.g., had we restricted fetch(), that would not mean the underlying fetch primitive would have to be versioned somehow.

Spent some more time getting clarity on that point--it turns out it was based on someone's assumption that the secure context state-of-the-page was mutable--once we got that sorted out, the internal versioning problem went away :-) (as an aside, Houdini features *are* going to add a lot of complexity to the layout pipeline, just not something that is dramatically impacted by secure context).

> And if specification authors make incorrect judgments, then implementers, the W3C TAG, and others will surely advice them otherwise.

Agree--and this is important no matter how this advice lands.

> I'd rather have them be conservative though than figure out what coupling means to them.

I think there's room for conservatism in the guideance. I just disagree that it takes the form of an absolute statement along the line of "all new features..."

dbaron:
So I think @travisleithead's https://github.com/w3ctag/design-principles/pull/75#issuecomment-347964832 is calling for an approach closer to where I started out (see the initial commit in https://github.com/w3ctag/design-principles/pull/75/commits/64e63a4d18a0376fe4e204586a4219c7ddb75002) than where I ended up after feedback from others (within and outside the TAG).  And I fully admit it might be a mixed message because I probably wasn't fully convinced by that feedback (and the consensus in the discussion when we discussed this on a TAG teleconference) but tried nevertheless to reflect it.

I need to do a little more thinking on how to sort this out (sorry, I managed to miss the github email on this until @annevk pointed it out to me today).

dbaron:
@annevk and I just talked about this a little more in person.

* I expressed the opinion that I really don't want to give advice that we want people to ask for exceptions to, because I expect whether people ask for exceptions is more a function of their personality rather than the real need for an exception; if we believe there are exceptions, we should document them, even if we can't document them exactly
* @annevk argued that we're better off taking a hard line at first, because it's easy (compatibility-wise) to relax the restriction later if we decide that the boundaries end up being too confusing
* I pointed out that some of the boundaries (for new capabilities that integrate tightly with an existing feature) might require significant implementation complexity for adding the checks, and @annevk agreed that cases where the boundary of the new API is complex enough that there would be significant implementation complexity to add the checks would justify not having SecureContext checks

dbaron:
... so @annevk was suggesting a compromise of mostly sticking with the harder line on secure contexts, but with more of an exception for cases where limiting a feature to secure contexts would lead to implementation complexity (so that people don't waste too much time trying to follow the strict advice).

hober:
I don't support limiting features to secure contexts *solely as a carrot to encourage HTTPS adoption*. Limiting new features that have privacy or security implications to secure contexts is sensible of course, but absent such reasons the implementation and authoring costs of fragmenting the platform generally outweigh the benefit of paternalism here. I'm glad @dbaron's planning to expand the exception around implementation complexity; I would go farther and carve out a similar exception for authoring complexity (again, when the feature does not have privacy or security concerns).

pinobatch:
And in addition to "implementation complexity" for web browser developers, another aspect of "implementation complexity" is complexity for server owners. It's one thing for the operator of a web server reachable through the Internet to build a secure context. It's a bit more difficult for, say, the operator of a consumer-grade router, printer, or network-attached storage (NAS) device on a private home LAN. I estimate that it would include at least the annual cost of a real domain name, plus the annual cost of a domain-validated certificate should the single point of failure that is Let's Encrypt go under, plus one or more *For Dummies* books.

othermaciej:
@hober's comment above is the general consensus at Apple. We think features should be restricted to secure contexts only when there is a privacy or security reason to do so for that specific feature. And it definitely should not be done for features where it would require constantly checking the isSecure bit during parsing of languages like CSS, WebAssembly or JavaScript.

While we agree with the goal of getting more of the web onto HTTPS, we don't think forking the web platform is an acceptable cost for doing so.

natewiebe13:
Provided `localhost` is considered a secure context, making secure contexts a requirement for new features is a welcome change. Working locally, you can still prototype features without having to setup a certificate. Once you deploy the code to a server, there should be no problem as setting up encryption using `certbot` is __very__ simple. I would argue that using consumer grade devices that have their own webserver isn't a cause for concern, as this is targeting new features, not ones that have already shipped. For example, most of these devices are using very outdated techniques since they are trying to support a large number of browsers (usually as old as IE9). None I've seen are using flexbox, CSS grid, or any web technology developed within the past few years.

Lastly, since certificates are essentially required for HTTP2. If we can get the web on HTTPS, it will make the transition to HTTP2 even easier.

pinobatch:
@natewiebe13 It's not always practical to test on `localhost` on mobile.

littledan:
I'm not sure how or whether this recommendation should be applied to programming languages like JavaScript and WebAssembly. TC39 has been trying to avoid making more JavaScript language modes with our "1JS" policy. Just from a parsing perspective, for example, adding another parameter to the grammar with new constructs banned adds significant complexity to the language definition, implementations and the testing matrix.

JavaScript library functions have not added I/O capabilities. The specification is currently not organized to give some global objects some library capabilities and others not; all global objects get all of the library. A change to this policy is possible, but it might relate to the in-development Realms specification (cc @erights). The most prominently security-relevant recent feature is SharedArrayBuffer, but the TC39 decision (taking into account apparent cross-browser consensus) has been to not remove SharedArrayBuffer from the ECMAScript specification.

annevk:
> TC39 has been trying to avoid making more JavaScript language modes with our "1JS" policy.

The difference is that this mode will go away long term, whereas class/module/non-strict/strict are likely to stick around.

littledan:
@annevk Many also have the goal that most/all JS code will be be able to transition to modules and in strict mode in the future.

annevk:
Most seems reasonable, but all seems far less realistic given event handlers, legacy scripts, etc.

littledan:
@annevk I'm not going to argue about whether non-secure contexts will go away, but, just saying this recommendation could be taken to affect JavaScript significantly in a way we're not currently working towards, or alternatively, it could be interpreted to be out of scope.

Is anyone who is working on this policy interested in presenting it to TC39? (or at least discuss it in a bug on [the ECMA-262 repo](https://github.com/tc39/ecma262/issues/)). Although the next meeting might be a little hard for some to attend physically (it's in London), it's possible to call into a VC. If you're associated with the W3C but not Ecma, it's possible to attend as a liaison or invited expert, just a matter of moving a few papers.

littledan:
@annevk To clarify, for JavaScript, do you think this policy should apply to things in TC39's ECMAScript standard itself, or only to Web APIs defined in other specifications on top of it?

littledan:
I [asked](https://github.com/tc39/ecma262/issues/1093) TC39 if they would like to adopt this policy, and so far the response has been widespread skepticism.

daurnimator:
There is a good discussion around this issue in this hackernews thread: https://news.ycombinator.com/item?id=16337998

pinobatch:
@natewiebe13 wrote:
>most of these devices are using very outdated techniques since they are trying to support a large number of browsers

Let me describe the NAS case in more detail. The Secure Contexts spec proposes requiring a secure context for the [Fullscreen API](https://fullscreen.spec.whatwg.org/) to make [phishing by spoofing the operating system UI](https://feross.org/html5-fullscreen-api-attack/) more difficult. A NAS on a home LAN might use the Fullscreen API to let users view videos stored on its drive. What certificate should the NAS's web server use once that becomes the case?

annevk:
@littledan I think it depends on the feature and the impl complexity, but if a major new feature like modules would come along today, that seems like something we should restrict.

pinobatch:
@annevk Plex can pay for the DigiCert partnership with from revenue from users who subscribe to Plex Pass ($5 per month, $40 per year, or $120 lifetime). Many developers of free software server applications have no analogous revenue source.

annevk:
You could build equivalent infrastructure on top of Let's Encrypt, no?

astoeckel:
Certainly, you could get a Let's Encrypt wildcard certificate for `*.<USER>.domain` and setup a DNS server that resolves `<IP>.<USER>.domain` with `<IP>`, and hand out the certificate private key to the user. However, I think that this solution is mildly insane and still doesn't solve the overall problem:

* This is significant overhead. We're talking about requiring infrastructure for every (free) software project that allows you to run a HTTP server on your local network. Why should a local server depend on *any* remote infrastructure. You may still want to use this software in ten years, when this infrastructure no longer exists.
* If the IP-based certificate provider is a publicly available service, this approach would be *less* secure than just connecting to a local HTTPS server via its IP address. The certificate is not signifying the authenticity of the server/device you're connecting to, but merely that the certificate has been issued to a person having control over `*.domain`. Furthermore, you would be trusting a (potentially) third party in actually returning the correct IP in the DNS query. But hey, you got a "green lock symbol" in the browser, so it must be secure, right?
* In the Let's Encrypt case the certificate needs to be renewed every three months. Believe it or not, there are regions in the world where no permanent access to the global internet exists; instead, access to information (think dumps of Wikipedia, etc.) are provided in a local network. There is absolutely no reason to artificially restrict these web applications to an old subset of JavaScript (by not providing a user-friendly way to trust a local HTTPS connection).

My proposal would be to only introduce secure contexts for new JavaScript features if at the same time there is a standardised way to get a "trust certificate on first use" mechanism (think SSH) in the browser, that does not scare users away. I've attached my humble attempt at a mockup of what something like that may look like in the UA (comparison of what it looks like right now, and what it may look like).

mokup svg

dbaron:
After the face-to-face discussion yesterday, the TAG concluded we couldn't come to consensus on strong advice about limiting features to secure contexts.  @slightlyoff drafted some text that we could have consensus on, which I've now merged with the text that was here.  Given how long the history is here, I decided to create the new proposal as a separate pull request in #89 rather than continuing to revise this one.

fantasai:
FWIW, I would like to throw my (fairly insubstantial) weight 100% behind @hober’s [comment](https://github.com/w3ctag/design-principles/pull/75#issuecomment-358126860) here. Authoring complexity matters, release-date–based modes are capricious, and CSS authors often do not have the influence over server configs to escape this trap, making it particularly vicious to impose on them.

Imho, W3C TAG should be recommending _against_ policies like “all new Web platform features added after X date are HTTPS-only, because we want to have more HTTPS carrots”. Such is not a policy crafted in service of good technical architecture: it is a _marketing project_ being implemented as technical architecture. I don't believe marketing is a good basis for making decisions about the architectural foundations of the Web platform, and in this case I do consider it harmful, for the various reasons described by others in this thread.

PR #80, opened 2017-10-07
Classes should have constructors whenever possible

domenic:
The long-awaited text on why constructors are good.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -111,95 +111,6 @@ and by name association.
 
 Please consult widely on names in your APIs.
 
-<h3 id="attributes-like-data">Attributes should behave like data properties</h3>
-
-[[!WEBIDL]] attributes are reified in JavaScript as accessor properties, i.e. properties with
-separate getter and setter functions which can react independently. This is in contrast to the
-"default" style of JavaScript properties, data properties, which do not have configurable behavior
-but instead can simply be set and retrieved, or optionally marked read-only so that they cannot be
-set.
-
-Data property semantics are what are generally expected by JavaScript developers when interfacing
-with objects. As such, although getters and setters allow infinite customizability when defining
-your Web IDL attributes, you should endeavor to make the resulting accessor properties behave as
-much like a data property as possible. Specific guidance in this regard includes:
-
-* Getters must not have any (observable) side effects.
-* Getters should not perform any expensive operations. (A notable failure of the platform in this
-    regard is getters like <code highlight="js">offsetTop</code> performing layout; do not repeat this mistake.)
-* Ensure that your attribute's getter returns the same object each time it is called, until some
-    occurrence in another part of the system causes a logical "reset" of the property's value. In
-    particular, <code highlight="js">obj.attribute === obj.attribute</code> must always hold, and so returning a
-    new value from the getter each time is not allowed.
-* Whenever possible, preserve values given to the setter for return from the getter. That is,
-    given <code highlight="js">obj.attribute = x</code>, a subsequent <code highlight="js">obj.attribute === x</code> should be
-    true. (This will not always be the case, e.g. if a normalization or type conversion step is
-    necessary, but should be held as a goal for normal code paths.)
-
-<h3 id="live-vs-static">Consider whether objects should be live or static</h3>
-
-Objects returned from functions, attribute getters, etc.,
-can either be live or static.
-A <dfn>live object</dfn> is one that continues to reflect
-changes made after it was returned to the caller.
-A <dfn>static object</dfn> is one that reflects
-the state at the time it was returned.
-
-Objects that are the way state is mutated are generally live.
-For example, DOM nodes are returned as live objects,
-since they are the API through which attributes are set and other changes
-to the DOM are made.
-They also reflect changes to the DOM made in other ways
-(such as through user interaction with forms).
-
-Objects that represent a collection
-that might change over time
-(and that are not the way state is mutated)
-should generally be returned as static objects.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -318,7 +233,190 @@ For example,
...
+You should strive to reduce such magic in designing your own APIs, and give them constructors which
+allow JavaScript developers to create them, just like the platform does. This means adding an
+appropriate [{{Constructor}}] extended attribute to your interface, and defining the algorithm for
+creating new instances of your class.
+
+Apart from reducing the magic in the platform, adding constructors allows JavaScript developers to
+create instances of the class for their own purposes, such as testing, mocking, or interfacing with
+third-party libraries which accept instances of that class. Additionally, because of how
+JavaScript's class design works, it is only possible to create a subclass of the class if it is
+constructible.
```
marcoscaceres: Because a class can be so easily defined in IDL, and because those classes end up in all global scopes, there needs to be strong advice to really generalize classes for maximum reuse (i.e., WG should talk to the community!)... and that "data only" classes, should just use Dictionaries (i.e., objects). Payment Request has the worst of both, for example: `PaymentAddress` is non-constructible, and pure data... it's tragic :(

index.bs
```diff
@@ -318,7 +233,190 @@ For example,
 is more readable than
 <code highlight="js">window.scrollBy(50, 0)</code>.
 
-<h3 id="promises">Design asynchronous APIs using Promises</h3>
+<h2 id="api-choices">Designing idiomatic and useful APIs</h2>
+
+<h3 id="constructors">Classes should have constructors when possible</h3>
+
+By default, [[!WEBIDL]] interfaces are reified in JavaScript as "non-constructible" classes: trying
```
marcoscaceres: nit: "reified" seems like fancy jargon (specially because it conjures up "semantic web reification" 🤢). Is there a simpler term we can use here? If no, that's ok.

annevk: represented?

index.bs
annevk: You should still be able to fix that, no?

marcoscaceres: If you mean adding a Constructor on these interfaces, then yes. We should probably do that -as it would not break anything, I think.

annevk: I meant turning them into a dictionary if they're just data.

marcoscaceres: No, unfortunately. Couple of reasons:

1. It would be a significant breaking change - as, IIRC, IDL attributes can't return Dictionaries, the API would need some sort of `.getAddress()` method - right now the `PaymentAddress` is returned via an attribute. 
1. The interface on which this attribute is exposed has `.toJSON()`, which itself calls `.toJSON()` on the attribute (returning a nested JSON structure).
1. Less importantly, but something that came up, there are two implementations that have been shipping for about 1 year - so implementers were reluctant to change things in a way that was not backwards compatible.

ojanvafai: PaymentAddress does seem like a strange mix of dictionaries and classes.

However, as discussed elsewhere I don't think we should be returning dictionaries from APIs in general as it makes it impossible to add methods to them in the future. Using dictionaries as arguments to functions as a way of having optional values sounds fine though.

This might actually be a thing the TAG should have a design principle by itself. @slightlyoff @domenic

index.bs
```diff
@@ -346,7 +350,190 @@ For example,
...
+to the DOM are made.
+They also reflect changes to the DOM made in other ways
+(such as through user interaction with forms).
+
+Objects that represent a collection
+that might change over time
+(and that are not the way state is mutated)
+should generally be returned as static objects.
+This is because it is confusing to users of the API when
+a collection changes while being iterated.
```
ojanvafai: This is one of the reasons. The other is that live collections incur a cost at a distance. They make other methods that affect that collection slower (e.g. adding/removing items), which is pretty unexpected.

index.bs
```diff
@@ -346,7 +350,190 @@ For example,
...
+Because of this,
+it is generally considered a mistake that methods like
+{{Document/getElementsByTagName()}} return live objects;
+{{ParentNode/querySelectorAll()}} was made to return static objects
+as a result of this experience.
+On the other hand, even though {{URLSearchParams}} represents a collection,
+it should be live because the collection is mutated through that object.
+
+Note:  It's possible that some of this advice should be reconsidered
```
ojanvafai: As per my comment above, it's the performance impact that I think is the primary reason. I guess it's unclear if the performance impact would apply to all maplike/setlike types.

index.bs
```diff
@@ -346,7 +350,190 @@ For example,
...
+for <a>maplike</a> and <a>setlike</a> types,
+where iterators have reasonable behavior
+for mutation that happens during iteration.
+This point likely needs further discussion,
+and perhaps further experience of use of these types.
+
+It's also worth considering the implications of having
```
ojanvafai: IMO, this is the primary reason to pushback on live collections. It's not the global performance cost though, it's that the performance cost is action at a distance, e.g. holding on to NodeLists makes completely unrelated methods, e.g. appendChild, slow in a way that's non-obvious.

index.bs
domenic: It doesn't make it impossible; it just means you have to upgrade to a class at that time.

travisleithead:
Thanks @domenic and all the other contributors to this PR! (And apologies for how long this has taken to merge.)

travisleithead:
If there are remaining concerns about the text, let's get some new issues filed--due to the merge conflicts, we felt it was easier to get the text as-authored into the doc first without trying to mess around with more merging... :-)

PR #82, opened 2017-10-13
Add section on new features being detectable.

dbaron:
This was originally part of #75, but it seemed worth splitting out both
into a separate section and a separate pull request.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -111,6 +111,29 @@ and by name association.
 
 Please consult widely on names in your APIs.
 
+<h3 id="feature-detect">New features should be detectable</h3>
+
+The existence of new features should generally be detectable,
+so that web content can act appropriately if the feature is present or not.
+This applies both to features that
+are not present because they are not implemented,
+and to features that may not be present in a particular configuration
+(ranging from features that are present only if particular hardware is present,
+to features that are
+only available in secure contexts.
+
+It should generally be indistinguishable why a feature is unavailable,
+so that feature detection code written for one case of unavailability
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -111,6 +111,29 @@ and by name association.
...
+<h3 id="feature-detect">New features should be detectable</h3>
+
+The existence of new features should generally be detectable,
+so that web content can act appropriately if the feature is present or not.
+This applies both to features that
+are not present because they are not implemented,
+and to features that may not be present in a particular configuration
+(ranging from features that are present only if particular hardware is present,
+to features that are
+only available in secure contexts.
```
annevk: Lacks closing ).

index.bs
```diff
@@ -111,6 +111,29 @@ and by name association.
 
 Please consult widely on names in your APIs.
 
+<h3 id="feature-detect">New features should be detectable</h3>
+
+The existence of new features should generally be detectable,
+so that web content can act appropriately if the feature is present or not.
```
mikewest: Style nit: I'd suggest replacing "if" with "whether".

index.bs
```diff
@@ -111,6 +111,29 @@ and by name association.
...
+This applies both to features that
+are not present because they are not implemented,
+and to features that may not be present in a particular configuration
+(ranging from features that are present only if particular hardware is present,
```
mikewest: Perhaps "present only on particular platforms"? It would be strange if, for instance, the WebUSB API disappeared entirely when no USB devices were plugged in.

PR #89, opened 2018-04-06
Add section on limiting features to secure contexts

dbaron:
I'm opening a separate PR to replace #75 so that it can stay somewhat archived in its current state.

Based on the TAG's discussion yesterday, the conclusion was that we wouldn't come to consensus on strong advice on requiring secure contexts.  Given that, @slightlyoff drafted some more explanatory text, which I've now merged with some of the still-usable advice I had in #75.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -368,6 +368,67 @@ and on when to use promises and when not to use promises,
 see <strong><a href="https://www.w3.org/2001/tag/doc/promises-guide">Writing
 Promise-Using Specifications</a></strong>.
 
+<h3 id="secure-context">Consider limiting new features to secure contexts</h3>
+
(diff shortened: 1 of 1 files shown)


daurnimator:
One idea I've had is that things could still be allowed when accessing something in private ip space.
e.g. still allow fullscreen API for use by a NAS if the NAS is hosted at 192.168.1.10.

daurnimator:
> @daurnimator that may be worth raising on the secure contexts spec, since it relates to the definition of a secure context, although some of those issues may be implementation specific.

Created https://github.com/w3c/webappsec-secure-contexts/issues/60

annevk:
Disappointing.

Issue #87, opened 2018-02-28
Guideance should discourage the use of the bitmask pattern (compareDocumentPosition)


travisleithead:
Looks like this is coming up again in WebGPU effort.
https://github.com/gpuweb/gpuweb/blob/master/design/sketch.webidl

See `WebGPUBufferUsageFlags` in that proposal.

I think the reasons against a bitmask for a webapi are: ergonomically tricky to reason about; dictionaries are more logically suited for a set of configuration flags... more?

annevk:
We stopped using non-string constants a long time ago to identify things?

travisleithead:
Yep, but for folks in the wasm space, numbers are faster than strings, so this comes up again...

annevk:
Wasm could have a different binding for enums, no?

travisleithead:
I suppose so, but I'm not sure how that relates to the bit-mask pattern in question here? Please elaborate.

rektide:
What would a design-principles respecting WebGPUBufferUsage look like?

```
dictionary WebGPUBufferUsage {
    bool MAP_READ;
    bool MAP_WRITE;
    bool TRANSFER_SRC;
    bool TRANSFER_DST;
    bool INDEX;
    bool VERTEX;
    bool UNIFORM;
    bool STORAGE;
}
```

Right? I'd like to make sure I understand what cfa5526f80197ec495c4e9055cb39f3809f0438a is telling me.

cynthia:
Yes, that is the intent of the text in the document. If this turns out to be a common pattern which needs to be optimized that can probably be standardized to some extent.

annevk:
@rektide without the uppercase member names. There's naming guidelines for dictionaries that would need to be respected.

PR #92, opened 2018-04-06
Initial draft for #87.

cynthia:
Fairly simple addition, last section could definitely use some extra eyes.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -439,11 +439,14 @@ a bitmask mechanism.
 However on the Web platform, it is more common to use a string constant for
 the cases where a constant is needed. This is much more inspection friendly
 for both development and expressing the constant codes through a user facing
-interface.
+interface, and there is generally no significant performance benefit in
+JavaScript engines to using an integer.
 
-Strings, unfortunately do not cover the use case for bitmasks. For these cases,
-it is recommended to use an object dictionary which contains the state that the
-bitmask is attempting to express.
+Strings, do not directly address the use case for bitmasks. For these cases,
+you should use an object dictionary which contains the state that the
+bitmask is attempting to express, as object dictionaries can then be passed
+around from method to method as needed as easily as the state in a single
+bitmask value.
 
 <h2 id="event-design">Event Design</h2>
 
--- a/index.bs
+++ b/index.bs
@@ -439,10 +439,10 @@ a bitmask mechanism.
 However on the Web platform, it is more common to use a string constant for
 the cases where a constant is needed. This is much more inspection friendly
 for both development and expressing the constant codes through a user facing
-interface, and there is generally no significant performance benefit in
-JavaScript engines to using an integer.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -430,6 +430,21 @@ on the authentication of the origin
...
+In many other platforms and programming languages, constants and enums are
+commonly expressed using a integer constant, sometimes in conjunction with
+a bitmask mechanism.
+
+However on the Web platform, it is more common to use a string constant for
+the cases where a constant is needed. This is much more inspection friendly
+for both development and expressing the constant codes through a user facing
+interface.
+
+Strings, unfortunately do not cover the use case for bitmasks. For these cases,
```
dbaron: Maybe change "unfortunately do not cover" to something more like "do not directly address", since I don't find it particularly unfortunate.

index.bs
```diff
@@ -430,6 +430,21 @@ on the authentication of the origin
...
+<h3 id="string-constants">Constants, enums, and bitmasks</h3>
+
+In many other platforms and programming languages, constants and enums are
+commonly expressed using a integer constant, sometimes in conjunction with
+a bitmask mechanism.
+
+However on the Web platform, it is more common to use a string constant for
+the cases where a constant is needed. This is much more inspection friendly
+for both development and expressing the constant codes through a user facing
+interface.
```
travisleithead: … and there is generally no significant performance benefit in JavaScript VMs to using an integer.

index.bs
```diff
@@ -430,6 +430,21 @@ on the authentication of the origin
...
+
+Strings, unfortunately do not cover the use case for bitmasks. For these cases,
+it is recommended to use an object dictionary which contains the state that the
```
dbaron: Maybe avoid the passive voice and change "it is recommended to use" to "you can use".  (Maybe there's something better, though?)

index.bs
```diff
@@ -430,6 +430,21 @@ on the authentication of the origin
...
+bitmask is attempting to express.
```
travisleithead: … object dictionaries can then be passed around from method to method as needed as easily as the state in a single bitmask value.

index.bs
triblondon: This sounds ambiguous to me. I'd prefer "In JavaScript VMs, using integers offers no significant performance benefit over strings"

index.bs
```diff
@@ -430,6 +430,24 @@ on the authentication of the origin
...
+commonly expressed using a integer constant, sometimes in conjunction with
+a bitmask mechanism.
+
+However on the Web platform, it is more common to use a string constant for
+the cases where a constant is needed. This is much more inspection friendly
+for both development and expressing the constant codes through a user facing
+interface, and there is generally no significant performance benefit in
+JavaScript engines to using an integer.
+
+Strings, do not directly address the use case for bitmasks. For these cases,
```
dbaron: remove the comma after "Strings"

PR #94, opened 2018-04-06
Polyfills, fixes #84

### LINKED ISSUE #84: Add advice on polyfills
At TPAC, most people who came to the polyfills breakout had not heard of the polyfill finding, including most polyfill authors.  We should do a better job of marketing on a number of fronts, and one of those would be to include a polyfill point in design principles.  This could point to the finding, but should also include some very high level guidance and perhaps distinguish between script related features and elements, (see https://github.com/w3ctag/polyfills/issues/31)

triblondon:
See #84

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -728,6 +728,16 @@ Please take advantage of these other excellent resources in your design process:
 * <a href="https://w3c.github.io/apa/fast/">Web Technology Accessibility Guidelines</a>
 * <a href="https://w3c.github.io/bp-i18n-specdev/">Internationalization Best Practices for Spec Developers</a> <!-- also at https://www.w3.org/TR/international-specs/ -->
 
+<h3 id="polyfills">Polyfills</h3>
+
+Polyfills can be hugely beneficial in helping to roll out new features to the web platform. The 
+Technical Architecture Group <a href="https://www.w3.org/2001/tag/doc/polyfills/">finding on Polyfills and the Evolution of the Web</a> offers guidance
+that should be considered in the development of new features, notably:
+
+* Don't be constrained by what is 'polyfillable'
+* Make your feature easily detectable
+* Work with polyfill authors
+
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -728,6 +728,16 @@ Please take advantage of these other excellent resources in your design process:
 * <a href="https://w3c.github.io/apa/fast/">Web Technology Accessibility Guidelines</a>
 * <a href="https://w3c.github.io/bp-i18n-specdev/">Internationalization Best Practices for Spec Developers</a> <!-- also at https://www.w3.org/TR/international-specs/ -->
 
+<h3 id="polyfills">Polyfills</h3>
+
+Polyfills can be hugely beneficial in helping to roll out new features to the web platform. The 
+Technical Architecture Group <a href="https://www.w3.org/2001/tag/doc/polyfills/">finding on Polyfills and the Evolution of the Web</a> offers guidance
+that should be considered in the development of new features, notably:
+
+* Being 'polyfillable' is not essential but <strong>is</strong> beneficial
```
kenchris: , before but?

index.bs
```diff
@@ -728,6 +728,16 @@ Please take advantage of these other excellent resources in your design process:
...
+* [[#feature-detect]]
```
kenchris: I don't know if this line is explanatory enough. Could use a few extra words

dbaron: This will show up as "New features should be detectable" with a link to that section; I think that seems fine.

index.bs
dbaron: I think it's also fine without the comma.

index.bs
kenchris: ok

dbaron:
But in hindsight... I'm not sure this is actually the best point within the document for this.  At the very least, I'm concern that the stuff just above the subsection might now be missed because people don't expect the section to have anything other than subsections.

PR #96, opened 2018-04-07
Initial draft for wrapper spec recommendations.

cynthia:
As noted - this is an initial draft.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -755,17 +755,16 @@ need to be careful when exposing these to the web platform.
     (operating systems and platforms) provide this functionality, and specify based
     on common denominators rather than ship different functionality sets.
 
- : Do not propose black-box proprietary library or dedicated processor dependencies
- :: Black-box dependencies are strongly discouraged, as this prevents wide
-    implementor adoption, which can cause serious fragmentation on the final
-    implementation quality and availability.
+ : Do not propose black-box proprietary library or dedicated hardware processor dependencies
+ :: Black-box dependencies are strongly discouraged, as this not only prevents wide
+    implementor adoption, but is unhealthy to the open nature of the web.
 
  : Underlying protocols should be open
  :: APIs which require exchange with external hardware or services should not depend
     on closed protocols.
 
  : Be offline friendly
- :: Web platform APIs should ideally not discreetely depend on online services.
+ :: Web platform APIs should ideally not have a hard dependency on online services.
     The availability of these services across different regions may vary, users
     may be offline, and the online services have no guarantee of being always
     available.
--- a/index.bs
+++ b/index.bs
@@ -675,9 +675,9 @@ accelerometer),
 section contains principles for consideration when designing APIs for devices.
 
 These can be functionality provided by the underlying operating system, or provided by a native
-third-party library to interact with a device. These are a thin abstraction which "wrap" the
-native functionality without introducing significant complexity to the browser, hence are called
-wrapper APIs.
+third-party library to interact with a device. These are an abstraction which "wrap" the
+native functionality without introducing significant complexity while securing the API surface
+to the browser, hence are called wrapper APIs.
 
 <h3 id="device-ids">Use care when exposing device identifiers</h3>
 
@@ -752,9 +752,12 @@ need to be careful when exposing these to the web platform.
     be isolated to a separate process.
 
  : Don't overconstrain or overfit the API to the underlying (wrapped) OS
- :: When designing a wrapper API, you should refer to how different implementations
-    (operating systems and platforms) provide this functionality, and specify based
-    on common denominators rather than ship different functionality sets.
+ :: When designing a wrapper API, you should consider how different implementations
+    (operating systems and platforms) provide this functionality. Ideally the entire
+    specification is implementable across platforms, but in some cases it may be
+    desirable to expose distinctions made on only some platforms.
+
+    These distinctions should be clearly noted if they are known in advance.
 
  : Do not propose black-box proprietary library or dedicated hardware processor dependencies
  :: Black-box dependencies are strongly discouraged, as this not only prevents wide
@@ -770,7 +773,7 @@ need to be careful when exposing these to the web platform.
     may be offline, and the online services have no guarantee of being always
     available.
 
- : Prevent additional fingerprinting surfaces
+ : Avoid additional fingerprinting surfaces
  :: Wrapper APIs can expose unintentionally expose the user to a wider fingerprinting surface.
     Please read the TAG's finding on
     <a href="http://www.w3.org/2001/tag/doc/unsanctioned-tracking/">unsanctioned tracking</a>
--- a/index.bs
+++ b/index.bs
@@ -662,7 +662,7 @@ or {{DOMException}}
 should be used for errors,
 whether they are exceptions, promise rejection values, or properties.
 
-<h2 id="device-apis">Device APIs</h2>
+<h2 id="device-apis">Device and Wrapper APIs</h2>
 
 It is increasingly common to see new APIs developed in the web platform for interacting with
 devices. For example, authors wish to be able to use the web to connect with their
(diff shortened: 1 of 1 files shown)


annevk:
Maybe title this something like "Exposing OS APIs"?

kenchris:
> Maybe title this something like "Exposing OS APIs"?

Yes, I like that more too. Generally editors should not actively create wrappers, though that will happen in some cases, in which these guidelines can be helpful

cynthia:
Thanks for the feedback. I’ll address this in an upcoming commit.

cynthia:
> Maybe title this something like "Exposing OS APIs"?

I gave this a bit of thought, and realized that the definition of wrapper APIs are wider than just OS APIs. This could also be a native third party library of some sort. (e.g. EME's CDM is one example.)

annevk:
What I probably should have said is that I don't think "wrapper API" is very clear. Is its meaning defined somewhere?

cynthia:
Although the concept of a wrapper is commonly understood in the context of FFI, you do raise a good point - I’ll elaborate on that.

index.bs
```diff
@@ -662,7 +662,7 @@ or {{DOMException}}
 should be used for errors,
 whether they are exceptions, promise rejection values, or properties.
 
-<h2 id="device-apis">Device APIs</h2>
+<h2 id="device-apis">Device and Wrapper APIs</h2>
```
dbaron: It's not clear to me that this warrants merging.  But if it does, it seems like maybe the introduction to the section should be extended to talk about wrapper APIs.  At the very least, I think wrapper APIs deserve a definition given that I don't think it's a well-known term.

index.bs
```diff
@@ -717,6 +717,56 @@ is done in a consistent and privacy-friendly way:
...
+
+ : Underlying protocols should be open
+ :: APIs which require exchange with external hardware or services should not depend
+    on closed protocols.
+
+ : Say no to external online service dependencies
+ :: Web platform APIs should ideally not discreetely depend on online services.
+    The availability of these services across different regions may vary, users
+    may be offline, and the online services have no guarantee of being always
+    available.
```
travisleithead: Also, this makes it possible to provide a service back-end as long as the API is async as an implementation detail.

index.bs
```diff
@@ -717,6 +717,56 @@ is done in a consistent and privacy-friendly way:
...
+
+Many modern operating systems come with convenient APIs which abstract away
+normally complex technology in the form of user-friendly APIs. Exposing these
+APIs to the web platform is getting more and more common, but you need to be
+careful when these are being exposed to the web platform.
+
+ : Generalize interface when underlying API is overly specific
+ :: If the wrapped native API is extremely specific to a certain library, it is
+    recommended to design a generic API which allows different browsers to not
+    have to implement this by emulating behavior in the event a different underlying
```
dbaron: This sentence has too many negatives.  I think everything after the comma could just be "we recommend designing an API that allows implementations to depend on different underlying libraries"

index.bs
```diff
@@ -717,6 +717,56 @@ is done in a consistent and privacy-friendly way:
...
+    of the underlying native API.
+
+ : Make API promise based whenever it makes sense
+ :: Even if the underlying API is synchronous, this does not necessarily mean it
+    translates well when ported to the web platform. Exposing a synchronous native
+    API to the web platform is generally considered bad taste, and should be
+    reinterpreted into a promise based API whenever possible.
+
+    Doing this also allows putting the API behind a permission when it is deemed
+    necessary, which may be necessary for powerful features.
```
dbaron: maybe drop the "which may be necessary for powerful features", and perhaps "when it is deemed" to "if it is"

index.bs
```diff
@@ -717,6 +717,56 @@ is done in a consistent and privacy-friendly way:
...
+    have to implement this by emulating behavior in the event a different underlying
+    native API or library is used.
+
+    A common example of bad practice is exposing the exact lifecycle and structs
+    of the underlying native API.
+
+ : Make API promise based whenever it makes sense
+ :: Even if the underlying API is synchronous, this does not necessarily mean it
+    translates well when ported to the web platform. Exposing a synchronous native
+    API to the web platform is generally considered bad taste, and should be
```
dbaron: "generally considered bad taste" makes it sound like there aren't good reasons behind it... but I think there are

index.bs
```diff
@@ -717,6 +717,56 @@ is done in a consistent and privacy-friendly way:
...
+
+    Doing this also allows putting the API behind a permission when it is deemed
+    necessary, which may be necessary for powerful features.
+
+ : Prevent underlying OS fragmentation
+ :: When designing a wrapper API, you should refer to how different implementations
+    (operating systems and platforms) provide this functionality, and specify based
+    on common denominators rather than ship different functionality sets.
+
+ : Do not propose black-box proprietary library or dedicated processor dependencies
```
dbaron: It's not clear what you mean by "dedicated processor"

index.bs
```diff
@@ -717,6 +717,56 @@ is done in a consistent and privacy-friendly way:
...
+ :: Black-box dependencies are strongly discouraged, as this prevents wide
+    implementor adoption, which can cause serious fragmentation on the final
+    implementation quality and availability.
+
+ : Underlying protocols should be open
+ :: APIs which require exchange with external hardware or services should not depend
+    on closed protocols.
+
+ : Say no to external online service dependencies
+ :: Web platform APIs should ideally not discreetely depend on online services.
```
dbaron: I suspect you mean discretely rather than discreetly (they're different), but I don't actually understand either.

index.bs
```diff
@@ -717,6 +717,56 @@ is done in a consistent and privacy-friendly way:
...
+    The availability of these services across different regions may vary, users
+    may be offline, and the online services have no guarantee of being always
+    available.
+
+ : Prevent additional fingerprinting surfaces
```
dbaron: This seems like somewhat stronger advice than the TAG's general fingerprinting advice; it may be better to refer to that instead.

index.bs
```diff
@@ -717,6 +717,56 @@ is done in a consistent and privacy-friendly way:
...
+Many modern operating systems come with convenient APIs which abstract away
+normally complex technology in the form of user-friendly APIs. Exposing these
+APIs to the web platform is getting more and more common, but you need to be
+careful when these are being exposed to the web platform.
+
+ : Generalize interface when underlying API is overly specific
+ :: If the wrapped native API is extremely specific to a certain library, it is
+    recommended to design a generic API which allows different browsers to not
+    have to implement this by emulating behavior in the event a different underlying
+    native API or library is used.
```
travisleithead: I spun off #106 to add more advice on naming...

index.bs
```diff
@@ -717,6 +717,56 @@ is done in a consistent and privacy-friendly way:
...
+    API to the web platform is generally considered bad taste, and should be
+    reinterpreted into a promise based API whenever possible.
+
+    Doing this also allows putting the API behind a permission when it is deemed
+    necessary, which may be necessary for powerful features.
+
+ : Prevent underlying OS fragmentation
+ :: When designing a wrapper API, you should refer to how different implementations
+    (operating systems and platforms) provide this functionality, and specify based
+    on common denominators rather than ship different functionality sets.
```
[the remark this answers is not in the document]
cynthia: It looks like the text suggests exactly what you described.

index.bs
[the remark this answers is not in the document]
cynthia: I don't have a good example, suggestions?

index.bs
travisleithead: Maybe "OS Capability Wrapper APIs" instead? (As the overall section title? E.g., making it more generic as this advice recommends... get it?)

index.bs
```diff
@@ -717,6 +717,56 @@ is done in a consistent and privacy-friendly way:
...
+
+ : Do not propose black-box proprietary library or dedicated processor dependencies
+ :: Black-box dependencies are strongly discouraged, as this prevents wide
```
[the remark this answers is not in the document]
cynthia: In the past, stuff like this (unfortunately) has gone through - browsers do have a choice of doing this, the best the guideline document can provide is to discourage this.

index.bs
```diff
@@ -717,6 +717,56 @@ is done in a consistent and privacy-friendly way:
...
+    implementor adoption, which can cause serious fragmentation on the final
```
dbaron: This sentence doesn't seem coherent -- what's "fragmentation on the final implementation quality and availability"?  (Seems like this could perhaps say something a little stronger, too.)

index.bs
```diff
@@ -717,6 +717,56 @@ is done in a consistent and privacy-friendly way:
...
+    implementation quality and availability.
+
+ : Underlying protocols should be open
+ :: APIs which require exchange with external hardware or services should not depend
```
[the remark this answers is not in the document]
cynthia: Would something like "ideally" work? Or is that actually making it weaker?

index.bs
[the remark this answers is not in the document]
cynthia: From how I understood it in the shape detection API, it seemed like they wanted to leave the door open for cases where they could offload processing over to a cloud service. Not sure if that was a liberal interpretation from my side though.

index.bs
```diff
@@ -717,6 +717,56 @@ is done in a consistent and privacy-friendly way:
      stable id during the session for the given origin, but also be able to deterministically produce
      the same id in subsequent sessions.
 
+<h3 id="wrapper-apis">Wrapper API considerations</h3>
```
travisleithead: Slightly more wordy but descriptive title suggetion: "APIs that simply pass-through OS APIs don't typically translate well to the web."

index.bs
cynthia: I was referring to "processor" in a sense of a chip, "dedicated hardware processor" sounds like an improvement?

index.bs
cynthia: I actually intended discreetly (in a "secretly" sense), maybe secretly or silently would be better?

index.bs
cynthia: Are you referring to this document? https://www.w3.org/2001/tag/doc/unsanctioned-tracking/

index.bs
```diff
@@ -717,6 +722,58 @@ is done in a consistent and privacy-friendly way:
...
+
+ : Generalize interface when underlying API is overly specific
+ :: Don't blindly map native convenience APIs 1:1 to the web platform.
+    Native APIs do not translate well to the web platform, and doing so may result
+    in a API shape which is difficult to implement across a variety of underlying
+    platforms and native libraries.
+
+    Especially be careful about exposing the exact lifecycle and data
+    structures of the underlying native APIs. Think about making an API that
+    fits well in with existing web platform APIs and which can be expressed in
```
travisleithead: "fits well in with existing web platform APIs" you can link this to https://w3ctag.github.io/design-principles/#string-constants
in this same doc.

index.bs
```diff
@@ -717,6 +722,58 @@ is done in a consistent and privacy-friendly way:
...
+    commonly available low-level APIs.
+
+ : Prefer asynchronous function by returning promises
+ :: Even if the underlying API is synchronous, this does not necessarily mean it
+    translates well when ported to the web platform. Exposing a synchronous native
+    API to the web platform is generally discouraged, and should be
+    provided as a promise based API whenever possible.
+
+    Doing this also allows putting the API behind a permission if it is
+    necessary.
```
travisleithead: Add disk I/O (or significant I/O) and Any cross-process communication.

index.bs
```diff
@@ -717,6 +722,58 @@ is done in a consistent and privacy-friendly way:
...
+
+ : Prevent underlying OS fragmentation
```
travisleithead: I don't think this header is right... maybe: "Don't overconstrain or overfit the API to the underlying (wrapped) OS

index.bs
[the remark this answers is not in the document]
dbaron: Yeah, I don't think it all needs to be common denominators -- there are ways to expose distinctions that are available on some platforms but not others through spec text that allows the right amount of variation between implementations.

index.bs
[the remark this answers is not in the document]
dbaron: I'd probably title the section as "Avoid" rather than "Prevent" to align with the finding that you've now cited.

PR #107, opened 2018-10-30
Add info about cancellation

kenchris:
Based on feedback from Domenic on Web NFC and reusing some of his wordings.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -379,6 +379,33 @@ and on when to use promises and when not to use promises,
 see <strong><a href="https://www.w3.org/2001/tag/doc/promises-guide">Writing
 Promise-Using Specifications</a></strong>.
 
+<h3 id="aborting">Cancel asynchronous APIs/operations using AbortSignal</h3>
+
+Async functions that need cancellation should take an `AbortSignal` as part
+of an options dictionary.
+
+Example:
+
+```js
+const controller = new AbortController();
+const signal = controller.signal;
+geolocation.read({ signal });
+```
+
+Reusing the same primitive everywhere has multiplicative effects throughout
+the platform. In particular, there's a common pattern of using a single
+`AbortSignal` for a bunch of ongoing operations, and then aborting them
+(with the corresponding `AbortController`) when e.g. the user presses cancel,
+or a single-page-app navigation occurs, or similar. So the minor extra
+complexity for an individual API leads to a large reduction in complexity
+when used with multiple APIs together.
+
+There might be cases where cancellation cannot be guaranteed. That is not a
+problem because a call to `abort()` on `AbortController` is a request to abort.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -461,6 +488,24 @@ bitmask value.
 
 <h2 id="event-design">Event Design</h2>
 
+<h3 id="dont-invent-event-like">Don't invent your own event listener-like infrastructure</h3>
+
+In some cases you might not have an object to inherit from `EventTarget` but
```
travisleithead: It could be convenient to create a pair of APIs to "register"/"unregister", "addListener"/"removeListener", "subscribe"/"unsubscribe", start"/"stop", etc., that take a callback and get invoked multiple times until paired cancel API is called. Instead, use the existing event registration pattern and API controls to start/stop the underlying process (since event listeners should not have side-effects). If the callback would have been provided specific data, then this data should be added to an Event object (but see https://w3ctag.github.io/design-principles/#state-and-subclassing as this is not always necessary).

In some cases...

travisleithead: In some cases, you can transfer the state that would be surfaced in callback parameters into a more persistent object which, in turn, can inherit from EventTarget.

index.bs
```diff
@@ -379,6 +379,33 @@ and on when to use promises and when not to use promises,
...
+(with the corresponding `AbortController`) when e.g. the user presses cancel,
+or a single-page-app navigation occurs, or similar. So the minor extra
+complexity for an individual API leads to a large reduction in complexity
+when used with multiple APIs together.
+
+There might be cases where cancellation cannot be guaranteed. That is not a
+problem because a call to `abort()` on `AbortController` is a request to abort.
+How you react to it depends on your spec.
+
+We could have given it a verbose name like `requestAbort()` but we went for brevity.
```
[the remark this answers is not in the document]
kenchris: Could it be a NOTE?

index.bs
```diff
@@ -379,6 +379,33 @@ and on when to use promises and when not to use promises,
...
+
+Reusing the same primitive everywhere has multiplicative effects throughout
+the platform. In particular, there's a common pattern of using a single
+`AbortSignal` for a bunch of ongoing operations, and then aborting them
+(with the corresponding `AbortController`) when e.g. the user presses cancel,
+or a single-page-app navigation occurs, or similar. So the minor extra
+complexity for an individual API leads to a large reduction in complexity
+when used with multiple APIs together.
+
+There might be cases where cancellation cannot be guaranteed. That is not a
```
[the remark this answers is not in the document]
cynthia: I think this is a good idea.

index.bs
kenchris: Comment for wrong PR :) but added to the right one

PR #109, opened 2018-10-30
Add a section about preferring algorithmic steps and state, or other more readable formal concepts that decompose to them.

### LINKED ISSUE #10: Prefer explicit algorithmic steps and constructs
For an example, see the before-and-after on Web Audio's `OfflineAudioContext.prototype.startRendering`:

[before](https://rawgit.com/WebAudio/web-audio-api/f99795b93077b62aaee16aa4fc0e21cbd20bd427/index.html#the-offlineaudiocontext-interface)

[after](https://rawgit.com/domenic/web-audio-api/c59a3b1a16dbf967815675f3d4206d6cc316edb7/index.html#the-offlineaudiocontext-interface)

Some points of note:
- Explicit use of flags, instead of saying things like "if it's called twice, throw an error."
- Explicit use of variables.
- Sequencing the steps of an operation, to make clear which of them are async.
- Explicitly firing events, instead of saying things like "The oncomplete handler will be called once the rendering has finished."

dbaron:
Add a section about preferring algorithmic steps and state, or other more readable formal concepts that decompose to them more readable formal concepts that decompose to them.

This also expands the section on specification writing advice to include
a substantive section, rather than being only links to other resources.

Fixes #10.

-----

Note that this is a first shot at this, and I'm expecting substantive comments.  I think @domenic and @annevk and @tabatkins might be particularly interested.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -738,16 +738,90 @@ that should be considered in the development of new features, notably:
 * [[#feature-detect]]
 * Polyfill development should be encouraged
 
-<h2 id="spec-writing">Advice on specification writing</h2>
+<h2 id="spec-writing">Writing good specifications</h2>
 
-This document covers API design for the Web,
+This document mostly covers API design for the Web,
 but those who design APIs are hopefully also writing specifications
 for the APIs that they design.
+
+<h3 id="writing-resources">Other resources</h3>
+
 Some useful advice on how to write specifications is available elsewhere:
   * <a href="https://ln.hixie.ch/?start=1140242962&amp;count=1">Writing
     specifications: Kinds of statements</a> (Ian Hickson, 2006)
   * <a href="https://www.w3.org/TR/qaframe-spec/">QA Framework:
     Specification Guidelines</a> (W3C QA Working Group, 2005)
+
+<h3 id="algorithms">Specify clearly enough so that algorithms are unambiguous</h3>
+
+Specification text is often more precise
+when it is written using an explicit sequence of steps.
+Using an ordered sequence of steps makes it clear how the required behaviors interact.
+For example, if there are multiple error conditions present,
+a specification that describes the behavior using a sequence of steps
+will make it clear which error will be reported,
+since the different errors would be checked and reported in different steps.
+
+Likewise, describing state using explicit flags also makes behavior clearer.
+Using explicit flags makes it clear
+whether or not the state changes in different error conditions,
+and makes it clear when the state described by the flags is reset.
+
+Whenever these differences could be observable to web content
+and there isn't a good reason to allow variation between implementations,
+it is important that they be specified unambiguously,
+so that implementations interoperate
+and web content doesn't end up
+depending on the behavior of one implementation and failing in another
+(which can cause higher development costs to fix the content for all browsers,
+or costs to users of broken content if it wasn't detected).
+
+However, sequences of steps are not the only way of making the specification precise enough.
+When there are particular characteristics
+that specification authors want to ensure the specification always satisfies,
+it may be better to use a less algorithmic way of formal specification
+that ensures these characteristics are always true.
+Developing these other ways of specifying behavior formally
+is more important when the algorithms would be more repetitive.
+These formal specification techniques can include dependencies on
+languages such as [[WEBIDL]], on grammar productions,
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -738,17 +738,91 @@ that should be considered in the development of new features, notably:
...
+it is important that they be specified unambiguously,
+so that implementations interoperate
+and web content doesn't end up
+depending on the behavior of one implementation and failing in another
+(which can cause higher development costs to fix the content for all browsers,
+or costs to users of broken content if it wasn't detected).
+
+However, sequences of steps are not the only way of making the specification precise enough.
+When there are particular characteristics
+that specification authors want to ensure the specification always satisfies,
```
annevk: editors?

index.bs
```diff
@@ -738,17 +738,91 @@ that should be considered in the development of new features, notably:
...
+For example, it is better to specify CSS selector matching
+by moving right to left across <a>combinators</a>,
+since that is what implementations do.
+This avoids the risk of future problems caused by new features
+that would be slightly different (in unimportant ways)
+between implementations based on one algorithm and implementations based on a different one.
+Thus, it avoids small future divergence in behavior between specification and implementations.
+It also helps to keep the costs of specifying features aligned with implementing them,
+since the complexity of adding a feature to the specification is more likely to be related
+to the cost of adding it to implementations.
```
annevk: This could use a note about not going full JIT in the specification. Basically, there's a tradeoff here around optimizations, where I think you want to be pretty careful (making optimizations a problem for the implementer isn't entirely responsible either, for instance).

index.bs
```diff
@@ -738,17 +738,91 @@ that should be considered in the development of new features, notably:
...
+
+Algorithms in specifications should also reflect best practices in programming.
+For example, they should explictly describe the inputs and outputs of the algorithm,
```
annevk: I don't think we typically specify outputs upfront today, but they're clearish from return statements. Perhaps something to reconsider though.

index.bs
```diff
@@ -738,17 +738,91 @@ that should be considered in the development of new features, notably:
...
+will make it clear which error will be reported,
+since the different errors would be checked and reported in different steps.
+
+Likewise, describing state using explicit flags also makes behavior clearer.
+Using explicit flags makes it clear
+whether or not the state changes in different error conditions,
+and makes it clear when the state described by the flags is reset.
+
+Whenever these differences could be observable to web content
+and there isn't a good reason to allow variation between implementations,
```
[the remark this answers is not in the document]
dbaron: I think there sometimes are, e.g., when there's latitude for UAs to make different privacy choices or apply various interventions to what Web content wants to do.

index.bs
tabatkins: I strongly prefer specifying outputs, personally. I'm definitely not *strict* about specifying them myself, but algorithms like <https://drafts.csswg.org/css-syntax/#consume-ident-like-token> make it clear what the range of the return type is when it may not be obvious from context, so you don't have to track variables and returns.

index.bs
dbaron: writers?

annevk: I'd prefer authors over writers.

dbaron: ok, I'll change it to editors.

index.bs
dbaron: Maybe the right condition here is that it's good to  follow the algorithms implementations would use, rather than alternatives, when that doesn't significantly increase the complexity of the specification?

index.bs
dbaron: When I wrote about specifying outputs, I was thinking more about having other bits of algorithm that peek inside the algorithm and use something written as internal to it as an additional output (i.e., the opposite of the stack introspection issue).  But I think Tab's idea about being clear about result types is good as well.

I guess I'm not sure if the text I have already is fine for covering that, or if there's a need to go into detail about the distinction...

index.bs
annevk: Yeah, that sounds like the kind of wording to go for. It might be worth referencing https://infra.spec.whatwg.org/#algorithms as well as ideally all specifications with algorithms build on top of that.

index.bs
annevk: I think this is probably fine.

PR #114, opened 2018-12-04
Minor: Correct number that short mods by

littledan:
65536, not 65535


dbaron:
Thanks for the fix, and sorry it took us a while to get to it.

PR #115, opened 2018-12-04
Add guidance for use of BigInt

littledan:
Only use BigInt where it is really needed! Number is good for existing
use cases, and existing APIs shouldn't "upgrade" to BigInt unless there
is a very strong reason.

Probably not ready to merge yet, as we will need more practical use cases
to emerge in the web platform to see how this comes up. For the same reason,
WebIDL support is not yet landed, and some open questions remain about how
flexible overloading/unions should be, see
https://github.com/heycam/webidl/pull/525#issuecomment-444073662

I'm putting this out for review now just in case an instance comes up in a TAG
review where someone proposes using BigInt in a specification. Please refer to
this PR for the advice I'd give, and better yet, @ me and I'm happy to take a
look as well.

cc @cynthia

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -656,6 +656,9 @@ Instead, you will want to stick with one of:
   :: Any nonnegative JavaScript number in the integer-representable range, throwing a
       <code highlight="js">TypeError</code> outside the range and rounding inside of it
 
+  : <code>bigint</code>
+  :: When the possible integer values range more broadly than JavaScript numbers can accurately represent
+
 Additionally, you can combine any of the above with an extra line in your algorithm to validate
 that the number is within the expected domain-specific range, and throwing or performing other
 actions in response. (While it is very rarely appropriate to modify author input by taking it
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -673,6 +676,16 @@ Those coming from other languages should carefully note that despite their names
 not 64.
 </div>
 
+<code>bigint</code> is recommended to be used <em>only</em> in circumstances where a <code>[EnforceRange] long long</code>
+or <code>[EnforceRange] unsigned long long</code> would not work in terms of expressivity: that is, only when the value
+ranges greater than 2<sup>53</sup> or less than -2<sup>53</sup>. In particular, APIs should not use the <code>bigint</code>
```
kenchris: You are using "ranges" as a verb here which makes it a bit hard to understand - could you reword?

kenchris:
We decided to merge this and add a follow-up PR to add our TAG recommendation

littledan:
Do you want me to make a follow up PR for "ranges" wording?

Note that we still have an open issue being discussed in TC39 and WebIDL about whether to encourage/allow overloading parameters between BigInt and Number. However, that doesn't change any of the text in this PR.

Also note that the WebIDL PR which adds the bigint type hasn't landed yet.

kenchris:
@dbaron is creating a follow up patch for that as we type

PR #118, opened 2019-01-07
Recommend using WindowEventHandlers.

jeremyroman:
From discussing with @domenic it seems that `WindowEventHandlers` is the desired place to add the event handler attribute for events which target `window`. If so, this seems useful to point out.


dbaron:
Thanks for the PR, and sorry for not getting to it for a bit.

Issue #106, opened 2018-10-30
More naming advice: avoid "brand" names


travisleithead:
The web will likely outlive the "branded" thing your API name is based on. Better to chose a name for the capability it offers, rather than a clever brand 'de jour'.

kenchris:
We will still use names like Bluetooth because it depends on the tech. Maybe don't use brands/tech names unless its specifics are exposed in the API (which is sometimes the point, eg. USB, Bluetooth)

annevk:
(Bluetooth is somewhat controversial for the reason @travisleithead mentions. It's platform-dependent and unclear if it'll outlast the web. I don't think it's been well articulated why breaking principles in such cases is okay.)

cynthia:
There was a recent review that exposed a very specific subset/revision of Bluetooth, which I think probably would fall into the scope of this issue.

PR #121, opened 2019-02-07
Attempt at addressing #106.

cynthia:
This is by no means intended to be the final submission.


travisleithead:
Note: if we have some particularly good or bad examples of this, it migtht be nice to include them.

cynthia:
While the example doesn’t quite fit the bill, XMLHttpRequest comes to mind as a unanimously agreeable example of bad naming. How does that example sound?

PR #122, opened 2019-02-07
Some text to warn developers away from doubling up the API surface

slightlyoff:
Particularly when considering BigInt vs. Number.

Refs #115

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -689,7 +689,10 @@ As far as ergonomics, JavaScript developers are expected to keep track of whethe
 conversions here could lead to a loss of precision, defeating the purpose
 of using BigInt.
 
-Further, try to resist doubling up the surface area of an API in order to provide both `BigInt` and `Number` variants of an API. As an API designer, consider it your job to chose the correct numeric type for the feature being exposed.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -660,38 +660,42 @@ Instead, you will want to stick with one of:
...
   :: When the possible integer values range more broadly than JavaScript numbers can accurately represent
 
-Additionally, you can combine any of the above with an extra line in your algorithm to validate
-that the number is within the expected domain-specific range, and throwing or performing other
-actions in response. (While it is very rarely appropriate to modify author input by taking it
-modulo 65536, it might be appropriate to take it modulo 360, for example.)
+Additionally, you can combine any of the above with an extra line in your
+algorithm to validate that the number is within the expected domain-specific
+range, and throwing or performing other actions in response. (While it is very
+rarely appropriate to modify author input by taking it modulo 65536, it might be appropriate to take it modulo 360, for example.)
```
kenchris: as you are already adding newlines to make the lines shorter, shouldn't you do this with the last one as well?

PR #125, opened 2019-02-07
Fix #119

### LINKED ISSUE #119: Guidance on events and promises
An [issue](https://github.com/w3c/screen-orientation/issues/120) came up recently in for the Screen Orientation lock API where by we have a promise being resolved and event being dispatched due to some action (the screen locking to a new orientation, which can trigger a "change" event). 

The question that came up is: 

 * should an event be dispatched _before_ a related promise is resolved? Or should the promise be resolved first, then the event dispatched? See https://github.com/w3c/screen-orientation/pull/147

The closest the Promise Guide comes to providing guidance is in the "[one time events](https://www.w3.org/2001/tag/doc/promises-guide#one-time-events)" section. My reading of that section is that the event should fire fist, and then the promise is resolved based on that action (e.g., "the image loaded, .then(...)", "the orientation locked, .then(...)" and so on).

kenchris:


index.bs
```diff
@@ -591,6 +591,12 @@ Follow the <a href="https://www.w3.org/2001/tag/doc/promises-guide#one-time-even
 in the <strong><a href="https://www.w3.org/2001/tag/doc/promises-guide">Writing
 Promise-Using Specifications</a></strong> guideline.
 
+<h3 id="promises-and-events">Events should fire before promises resolve</h3>
+
+In the case that an event is being dispatches as part of a promise being
```
slightlyoff:
```suggestion
In the case that an event is being dispatched as part of a promise being
```

index.bs
```diff
@@ -591,6 +591,12 @@ Follow the <a href="https://www.w3.org/2001/tag/doc/promises-guide#one-time-even
...
+resolved, then the event should be dispatches before the promise resolves and
```
slightlyoff:
```suggestion
resolved, then the event should be dispatches before the promise resolves, rather 
than after.
```

index.bs
kenchris: I don't know why I always make that typo!

index.bs
```diff
@@ -591,6 +591,12 @@ Follow the <a href="https://www.w3.org/2001/tag/doc/promises-guide#one-time-even
...
+not right after.
```
slightlyoff:
```suggestion

```

travisleithead:
(When the PR comes back:)

I would phrase this as:

When an algorithm desires to both dispatch Events and use Promises, the Event(s) should be dispatched first *before* the Promise (representing the logical completion of the work) is resolved. Otherwise, ... [explain why this is bad, e.g., might *seem* like the event would fire *after* the resolved promise, but in fact it won't due to micro-task timing for Promises?]

PR #129, opened 2019-05-28
Update GC observability with WeakRefs

littledan:
Changing a "must" to "should" here, and referencing the current state
of the WeakRefs proposal (multiple implementation support). In my mind,
a typical specification review should go just as if the WeakRefs proposal
did not exist; this isn't a justification to expose GC further.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -101,9 +101,12 @@ original in order to create interoperability. This would cause current garbage c
 strategies to ossify, preventing improvement in one of the most dynamic areas of JavaScript
 virtual machine technology.
 
-In particular, this means that you should not expose any API that acts as a weak reference, e.g. with a
-property that becomes <code highlight="js">null</code> once garbage collection runs. Such freeing of memory
-should be entirely deterministic.
+For example, currently, {{Document/getElementsByTagName}} permits reusing the {{HTMLCollection}}
+which it creates, when it's called with the same receiver and tag name. In practice, engines reuse
+the output if it has not been garbage collected. This creates behavior which differs based on the
+details of garbage collection, which is strongly discouraged. If {{Document/getElementsByTagName}}
+were designed today, the advice to the specification authors would be to either reliably reuse
+the output, or to produce a new {{HTMLCollection}} each time it's invoked.
 
 <h2 id="api-surface">API Surface Concerns</h2>
 
--- a/index.bs
+++ b/index.bs
@@ -101,12 +101,21 @@ original in order to create interoperability. This would cause current garbage c
 strategies to ossify, preventing improvement in one of the most dynamic areas of JavaScript
 virtual machine technology.
 
+In particular, this means that you shouldn't expose any API that acts as a weak reference, e.g. with a
+property that becomes <code highlight="js">null</code> once garbage collection runs. Such freeing of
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -81,28 +82,28 @@ object property).
...
-
-<div class="note">
-There is some speculative discussion of exposing weak references such that their finalization is
-only observable between event loop turns in this <a
-href="https://github.com/tc39/proposal-weakrefs">stage 1 proposal under discussion at TC39</a>.
-However, this remains contentious and does not yet have consensus in TC39 or among implementers.
-</div>
+In particular, this means that you should not expose any API that acts as a weak reference, e.g. with a
+property that becomes <code highlight="js">null</code> once garbage collection runs. Such freeing of memory
+should be entirely deterministic.
```
annevk: I wonder if we should reword this to say that losing the reference should be deterministic, as we just argued that freeing being implementation-specific is nice. (To be clear, already at fault in the original, so maybe a follow-up.)

littledan: This makes sense. How about I just replace this paragraph for a concrete example, explaining why algorithms like getElementsByTagName are bad, and how you're supposed to do it, and how this ties into GC observability?

index.bs
```diff
@@ -81,28 +82,31 @@ object property).
...
+collection and whether certain objects are collected at all, which means the resulting code will be
 non-interoperable. Worse, according to the usual rules of game theory as applied to browsers, this
 kind of scenario could force other user agents to copy the garbage collection timing of the
 original in order to create interoperability. This would cause current garbage collection
 strategies to ossify, preventing improvement in one of the most dynamic areas of JavaScript
 virtual machine technology.
 
-In particular, this means that you can't expose any API that acts as a weak reference, e.g. with a
-property that becomes <code highlight="js">null</code> once garbage collection runs. Such freeing of memory must
-be entirely deterministic.
```
dbaron: I think it's worth keeping this example somehow, since it's a particularly clear and simple example of what it means for an API to depend on GC timing.

index.bs
```diff
@@ -81,28 +82,31 @@ object property).
...
-has run.
+Web APIs should not expose a way for author code to deduce when/if garbage collection of JavaScript
+objects has run.
 
-The reason for this is somewhat subtle. If garbage collection timing were observable, then authors
-could easily write code relying on specific garbage collection timing. But this timing is almost
-certainly not the same across user agents, which means the resulting code will be
+Although some APIs may expose garbage collection, such as some implementations of
+{{Document/getElementsByTagName}}, and the JavaScript <a
+href="https://github.com/tc39/proposal-weakrefs">WeakRefs proposal</a> (which has multiple
```
dbaron: It may be worth pointing out that WeakRef is less harmful because it's explicit about its dependency on GC timing, and therefore developers are more likely to be aware of it.  Whereas, if an API has something that "randomly" becomes null at some point in the future, authors are likely to be unaware that they're depending on GC timing by assuming it's non-null.

littledan:
Added some wording to get at the great points from @dbaron .

mathiasbynens:
Is this patch blocked on anything in particular? (I'd like to point to this section but it currently still has the outdated content.)

PR #141, opened 2019-10-15
Remove out-of-date WebUSB example

reillyeon:
The WebUSB specification no longer requires that a device explicitly name the origins that can connect to it. A discussion of why this requirement was removed is included in the [Attacking a Device](https://wicg.github.io/webusb/#attacking-a-device) section of the specification.


cynthia:
We should probably replace the example to something else, but that's homework for us. Thanks for the PR!

Issue #81, opened 2017-10-09
Synchronous events


annevk:
I never quite understand what this term means when I come across it. I guess it means dispatch it from something that occurs at least one task or microtask later?

But I'm also not sure the advice to always avoid them is good. E.g., we dispatch within the the abort algorithm directly to `AbortSignal`.

Another thing we should add is that you should change state in the same algorithm that the event is dispatched. That way you cannot observe state changes without the notification.

travisleithead:
We have this current guidance in the doc: https://w3ctag.github.io/design-principles/#favor-asynchronous-events.

We can probably improve that guidance... :)

Re the definition, I think "synchronous events" are dispatched from an algorithm happening in a task--they are dispatched "in line" and are expected to complete before further steps of the algorithm are executed. (e.g., like the [focus update steps](https://html.spec.whatwg.org/multipage/interaction.html#focus-update-steps) that can run as part of the `focus()` API).

Indeed, as you note, the [`abort()`](https://dom.spec.whatwg.org/#dom-abortcontroller-abort) method of the `AbortController` starts an algorithm that synchronously fires `abort`. At this point, any author code may run, including code that could then call `abort()` again, causing a reentrancy into the platform and potential stack overflow--of course, you've carefully protected against this by adding an aborted flag which is allows the algorithm to gracefully unwind. Not everyone writing algorithms are as careful. By queuing the event instead of firing it directly, you avoid the reentrant scenario at the expense of one turn of the event loop.

annevk:
You'd also introduce an observable delay in the cancelation and create a host of other issues you'd have to protect against. What if you invoke `abort()` twice in your setup?

Also, I know there's guidance in the document; this is feedback on that guidance not being great.

travisleithead:
@annevk I hope to have improved this advice with #151. OK to close?

annevk:
Yeah, thanks for tackling that and getting rid of a/synchronous!

PR #151, opened 2019-12-24
Better async event advice

travisleithead:
Taking a stab at addressing #81.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -19,7 +19,6 @@ Link Defaults: html (dfn) queue a task/in parallel/reflect
 <pre class="link-defaults">
 spec:html; type:dfn; for:/; text:browsing context
 spec:webidl; type:dfn; text:namespace
-spec:webidl; type:extended-attribute; text:Constructor
 </pre>
 <pre class="anchors">
 url: https://tc39.github.io/ecma262/#sec-error-objects; type: dfn; text: ECMAScript error objects;
@@ -387,7 +386,7 @@ of Web IDL-derived classes without ever actually calling their constructor.
 
 You should strive to reduce such magic in designing your own APIs, and give them constructors which
 allow JavaScript developers to create them, just like the platform does. This means adding an
-appropriate [{{Constructor}}] extended attribute to your interface, and defining the algorithm for
+appropriate [=constructor operation=] to your interface, and defining the algorithm for
 creating new instances of your class.
 
 Apart from reducing the magic in the platform, adding constructors allows JavaScript developers to
@@ -939,7 +938,7 @@ See [[!HIGHRES-TIME]] for more details.
 <h3 id="error-types">Use Error or DOMException for errors</h3>
 
 Errors in web APIs should be represented as
-<a>ECMAScript error objects</a> (perhaps via the WebIDL {{Error}} type)
+<a>ECMAScript error objects</a> (e.g., [=ECMAScript error objects|Error=])
 or as {{DOMException}}.
 There was at one point a trend to use {{DOMError}} when
 objects had a property representing an error.
--- a/index.bs
+++ b/index.bs
@@ -17,11 +17,13 @@ Boilerplate: feedback-header off
 Link Defaults: html (dfn) queue a task/in parallel/reflect
 </pre>
 <pre class="link-defaults">
-spec:html; type:dfn; for:/; text:browsing context
+spec:html; type:dfn; for:/; 
+    text:browsing context
+    text:focus update steps
 spec:webidl; type:dfn; text:namespace
 </pre>
 <pre class="anchors">
-url: https://tc39.github.io/ecma262/#sec-error-objects; type: dfn; text: ECMAScript error objects;
+url: https://tc39.github.io/ecma262/#sec-error-objects; type: interface; text: Error;
 url: https://www.w3.org/TR/WebCryptoAPI/#dfn-Crypto; type: interface; text: Crypto
 url: https://w3c.github.io/touch-events/#idl-def-touchevent; type: interface; text: TouchEvent
 url: https://w3c.github.io/IntersectionObserver/;
@@ -619,7 +621,7 @@ callback and get invoked multiple times until paired cancel API is called.
 
 Instead, use the existing event registration pattern and separate API controls
 to start/stop the underlying process (since event listeners should not have
-<a href="#events-are-for-notification">side-effects</a>).
+[[#events-are-for-notification|side-effects]]).
 
 If the callback would have been provided specific data, then this data should
 be added to an `Event` object (but see <a href="#state-and-subclassing">State
@@ -680,46 +682,50 @@ Try to design DOM events to deliver after-the-fact notifications of changes. It 
 
 <h3 id="guard-against-recursion">Guard against potential recursion</h3>
 
-When designing a long-running or complicated algorithm which is initiated by an 
-API call, events are appropriate for <a href="#events-are-for-notification">notifying</a>
-user code of the ongoing process, but also introduce the possibility of unexpected
+When designing a long-running or complicated algorithm that is initiated by an 
+API call, events are appropriate for [[#events-are-for-notification|notifying]]
+user code of the ongoing process. However, they also introduce the possibility of unexpected
 re-execution of the current algorithm before it has finished! Because user code gets
 to run in the middle of the algorithm, recursion can happen if that user code triggers
 the algorithm again (directly or indirectly).
 
-If the complexity of the algorithm is reasonably understood, you may add appropriate
-state to track that the algorithm is in progress and terminate early if the algorithm
+If the algorithmic complexity and call graph is reasonably understood, add appropriate
+state to track that the algorithm is in progress and terminate immediately if the algorithm
 is already concurrently running. This technique is "guarding" the algorithm:
 
 <div class="example">
-    A technique (e.g., used for [abort()](https://dom.spec.whatwg.org/#abortsignal-add)
-    or as a stack-based variation in
-    [focus()](https://html.spec.whatwg.org/multipage/interaction.html#focus-update-steps))
-    for guarding against unplanned recursion. This example is given in JavaScript
-    for terseness but would be applied in algorithmic prose in a spec.
+    A technique (e.g., as used in the [=AbortSignal/add|AbortSignal's add algorithm=]
+    or as a stack-based variation in the [=focus update steps=])
+    for guarding against unplanned recursion. 
+    
+    In this example, an object of type <dfn noexport interface>MyObject</dfn> has
+    a method <dfn method for="MyObject">doComplexOpWithEvents()</dfn>. {{MyObject}}
+    also has an internal state attribute
+    {{MyObject}}.[[<dfn noexport attribute for="MyObject">started</dfn>]] whose initial
+    value is `false`.
 
-    <pre highlight="js">
-    const isStarted = false;
-
-    function doComplexOperationWithEvents() {
-        if ( isStarted ) {
-            return; // Terminate early
-        }
-        isStarted = true;
-        // Do complex stuff
+    The {{doComplexOpWithEvents()}} method MUST act as follows:
+    
+    1. If {{MyObject}}.[[{{started}}]] is not `false` then throw an {{InvalidStateError}}
+        {{DOMException}} and terminate this algorithm.
         
-        // Notify the developer of progress by firing events
-        //   Note: any other code may run here, including calling 
-        //   doComplexOperationWithEvents() again!
+        Note: {{doComplexOpWithEvents()}} is not allowed to be run again
+        (e.g., from within a `"currentlyinprogress"` event handler) before it finishes
+        its original execution.
 
-        // Finishe doing complex stuff
-        isStarted = false;        
-    }
-    </pre>
+    2. Set {{MyObject}}.[[{{started}}]] to `true`.
+
+    3. ...Do complex stuff...
+
+    4. [=Fire an event=] named `"currentlyinprogress"` at {{MyObject}}.
+
+    5. ...Finish up complex stuff...
+
+    6. Set {{MyObject}}.[[{{started}}]] to `false`.
 </div>
 
 Sometimes "guarding" as illustrated above cannot be done because of the complexity
-of the algorithm, the number of potential algorithm entrypoints or the side-effects
+of the algorithm, the number of potential algorithm entry-points or the side-effects
 beyond the algorithm's control. In these cases another form of protecting the integrity
 of the algorithm's state is to "defer" the event to a subsequent task or microtask.
 Deferral ensures that any stack-based recursion is avoided (but does not eliminate
@@ -732,9 +738,9 @@ By way of illustration, the deprecated
 were specified to fire in the middle of node insertion and removal algorithms, which
 are triggered by a variety of common API calls like {{Node/appendChild()}}. The side-effects
 from recursivly running node mutation algorithms caused many years of security issues.
-A more modern approach embodied in
-[Mutation Observers](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver)
-applies the principle of deferral of notification--in this case to the microtask queue
+A more robust approach is embodied in
+{{MutationObserver}}
+applies the principle of deferral of notification — in this case to the microtask queue
 following the algorithm.
 
 Weigh the trade-offs when considering deferral as an alternative to guarding as noted
@@ -748,11 +754,11 @@ earlier. Guarding is generally the preferred approach. With guarding:
     that state).
 
 Deferral as an option generally means *all* API entrypoints need guards, which is
-usually more work. If this is not the case, then its likely the algorithm is already
-finished when the event fires (there's no other state changes expected), in which
-case, perhaps you should be
-[considering a Promise-based design](https://www.w3.org/2001/tag/doc/promises-guide)
-instead.
+usually more work for the spec author. If there seems to be no need for guards, then
+perhaps this is because all of the state transitions have been completed and the event
+is meant to signal completion of the algorithm. In this case, it's likely that instead
+of using an event to signal completion, the API should be designed to return and complete
+a {{Promise}} instead. See [[#one-time-events]].
 
 Note: events that expose the possibility of recursion as described in this section
 were sometimes called "synchronous events". This terminology is discouraged as it
@@ -938,7 +944,7 @@ See [[!HIGHRES-TIME]] for more details.
 <h3 id="error-types">Use Error or DOMException for errors</h3>
 
 Errors in web APIs should be represented as
-<a>ECMAScript error objects</a> (e.g., [=ECMAScript error objects|Error=])
+ECMAScript error objects (e.g., {{Error}})
 or as {{DOMException}}.
 There was at one point a trend to use {{DOMError}} when
 objects had a property representing an error.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -863,7 +938,7 @@ See [[!HIGHRES-TIME]] for more details.
 <h3 id="error-types">Use Error or DOMException for errors</h3>
 
 Errors in web APIs should be represented as
-<a>ECMAScript error objects</a> (perhaps via the WebIDL {{Error}} type)
+<a>ECMAScript error objects</a> (e.g., [=ECMAScript error objects|Error=])
```
marcoscaceres: Do you mean something like: 

```suggestion
<a>ECMAScript error objects</a> (e.g., {{TypeError}})
```

The aliasing of Error seems strange?

index.bs
```diff
@@ -679,11 +678,87 @@ are traditionally added to {{WindowEventHandlers}} rather than {{Window}}.</p>
...
 
-<h3 id="favor-asynchronous-events">Favor asynchronous events</h3>
+<h3 id="guard-against-recursion">Guard against potential recursion</h3>
 
-A few events in the platform are specified to dispatch synchronously. These events cause problems for engines and performance issues in applications due to the possibility for re-entrant behavior they open up. The deprecated [Mutation Events](https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Mutation_events), for instance, has caused many years of security issues. A more modern approach embodied in [Mutation Observers](ihttps://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) addresses most of the same use-cases in a higher-performance way which is easier to develop with and implement.
+When designing a long-running or complicated algorithm which is initiated by an 
+API call, events are appropriate for <a href="#events-are-for-notification">notifying</a>
```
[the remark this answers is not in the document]
travisleithead: Well, there's no dfn for this, it's a section header, so I guess I can do `[[#events-are-for-notification|notifying]]` which is a bit more compact.

index.bs
```diff
@@ -679,11 +678,87 @@ are traditionally added to {{WindowEventHandlers}} rather than {{Window}}.</p>
...
 
-<h3 id="favor-asynchronous-events">Favor asynchronous events</h3>
+<h3 id="guard-against-recursion">Guard against potential recursion</h3>
 
-A few events in the platform are specified to dispatch synchronously. These events cause problems for engines and performance issues in applications due to the possibility for re-entrant behavior they open up. The deprecated [Mutation Events](https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Mutation_events), for instance, has caused many years of security issues. A more modern approach embodied in [Mutation Observers](ihttps://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) addresses most of the same use-cases in a higher-performance way which is easier to develop with and implement.
+When designing a long-running or complicated algorithm which is initiated by an
```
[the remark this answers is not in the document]
travisleithead: 👍

index.bs
```diff
@@ -679,11 +678,87 @@ are traditionally added to {{WindowEventHandlers}} rather than {{Window}}.</p>
...
 
-<h3 id="favor-asynchronous-events">Favor asynchronous events</h3>
+<h3 id="guard-against-recursion">Guard against potential recursion</h3>
 
-A few events in the platform are specified to dispatch synchronously. These events cause problems for engines and performance issues in applications due to the possibility for re-entrant behavior they open up. The deprecated [Mutation Events](https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Mutation_events), for instance, has caused many years of security issues. A more modern approach embodied in [Mutation Observers](ihttps://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) addresses most of the same use-cases in a higher-performance way which is easier to develop with and implement.
+When designing a long-running or complicated algorithm which is initiated by an 
+API call, events are appropriate for <a href="#events-are-for-notification">notifying</a>
+user code of the ongoing process, but also introduce the possibility of unexpected
+re-execution of the current algorithm before it has finished! Because user code gets
```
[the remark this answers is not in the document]
travisleithead: Cool. Going with a ". However..." approach.

index.bs
```diff
@@ -679,11 +678,87 @@ are traditionally added to {{WindowEventHandlers}} rather than {{Window}}.</p>
...
+to run in the middle of the algorithm, recursion can happen if that user code triggers
+the algorithm again (directly or indirectly).
 
-If you feel you need a synchronous event in your design, please reconsider and ask the TAG for help in redesigning the API.
+If the complexity of the algorithm is reasonably understood, you may add appropriate
```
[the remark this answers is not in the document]
travisleithead: Mmm. Took another shot at less ambiguous words...

index.bs
```diff
@@ -679,11 +678,87 @@ are traditionally added to {{WindowEventHandlers}} rather than {{Window}}.</p>
...
+state to track that the algorithm is in progress and terminate early if the algorithm
```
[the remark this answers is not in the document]
travisleithead: Yep.

index.bs
```diff
@@ -679,11 +678,87 @@ are traditionally added to {{WindowEventHandlers}} rather than {{Window}}.</p>
...
+is already concurrently running. This technique is "guarding" the algorithm:
+
+<div class="example">
+    A technique (e.g., used for [abort()](https://dom.spec.whatwg.org/#abortsignal-add)
```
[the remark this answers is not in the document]
travisleithead: Actually, I want the specific algorithm inside of the API calls which showcase the examples of this being done... I'll look for how they're linked... :)

index.bs
```diff
@@ -679,11 +678,87 @@ are traditionally added to {{WindowEventHandlers}} rather than {{Window}}.</p>
...
+    or as a stack-based variation in
+    [focus()](https://html.spec.whatwg.org/multipage/interaction.html#focus-update-steps))
+    for guarding against unplanned recursion. This example is given in JavaScript
```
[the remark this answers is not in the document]
travisleithead: A very helpful reference. Thanks!

index.bs
```diff
@@ -679,11 +678,87 @@ are traditionally added to {{WindowEventHandlers}} rather than {{Window}}.</p>
...
+potentially problematic loops, as they could now occur as unending follow-up tasks).
+
+Defering an event is often specified as "[=queue a task=] to [=fire an event=]...".
+
+By way of illustration, the deprecated
+[Mutation Events](https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Mutation_events),
+were specified to fire in the middle of node insertion and removal algorithms, which
+are triggered by a variety of common API calls like {{Node/appendChild()}}. The side-effects
+from recursivly running node mutation algorithms caused many years of security issues.
+A more modern approach embodied in
```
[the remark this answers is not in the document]
travisleithead: 👍

index.bs
```diff
@@ -679,11 +678,87 @@ are traditionally added to {{WindowEventHandlers}} rather than {{Window}}.</p>
...
+[Mutation Observers](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver)
```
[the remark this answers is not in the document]
travisleithead: Like it.

index.bs
```diff
@@ -679,11 +678,87 @@ are traditionally added to {{WindowEventHandlers}} rather than {{Window}}.</p>
...
+applies the principle of deferral of notification--in this case to the microtask queue
```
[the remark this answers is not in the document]
travisleithead: Good catch.

index.bs
```diff
@@ -679,11 +678,87 @@ are traditionally added to {{WindowEventHandlers}} rather than {{Window}}.</p>
...
+following the algorithm.
+
+Weigh the trade-offs when considering deferral as an alternative to guarding as noted
+earlier. Guarding is generally the preferred approach. With guarding:
+
+* all state management can be rationalized within one "turn" of the algorithm (no
+    need to consider any state changes or API invocations between the time the algorithm
+    completes and when the event fires).
+* events can reveal internal state changes as early as possible when dispatched immediately
```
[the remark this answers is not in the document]
travisleithead: This is exactly it. Between `conditionsAreRight` and the time that `queueATask` fires, conditions may have changed (since who knows that task will have been picked up and run before your queued task). This appears to be annevk's concern with recommending queuing a task as the default suggestion back in issue #81.

index.bs
```diff
@@ -679,11 +678,87 @@ are traditionally added to {{WindowEventHandlers}} rather than {{Window}}.</p>
...
+    in the algorithm (resulting in minimal delay to user code that wants to act on
+    that state).
+
+Deferral as an option generally means *all* API entrypoints need guards, which is
+usually more work. If this is not the case, then its likely the algorithm is already
```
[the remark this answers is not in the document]
travisleithead: for the spec author :)

index.bs
```diff
@@ -679,11 +678,87 @@ are traditionally added to {{WindowEventHandlers}} rather than {{Window}}.</p>
...
+finished when the event fires (there's no other state changes expected), in which
+case, perhaps you should be
+[considering a Promise-based design](https://www.w3.org/2001/tag/doc/promises-guide)
+instead.
```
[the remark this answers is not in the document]
travisleithead: Ug. Yeah, on re-reading this the next day, it does need some help. Take 2 coming...

index.bs
travisleithead: Mmm. It moved out of webidl; I've fixed it up and dropped the alias.

index.bs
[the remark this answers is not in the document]
marcoscaceres: True, something could invalidate the task queue... like maybe the document becoming inactive. It's a hard problem. I'm not sure what we can suggest here apart from... here are a bunch of things to think hard about...

index.bs
```diff
@@ -679,11 +680,91 @@ are traditionally added to {{WindowEventHandlers}} rather than {{Window}}.</p>
...
+    
+    In this example, an object of type <dfn noexport interface>MyObject</dfn> has
+    a method <dfn method for="MyObject">doComplexOpWithEvents()</dfn>. {{MyObject}}
+    also has an internal state attribute
+    {{MyObject}}.[[<dfn noexport attribute for="MyObject">started</dfn>]] whose initial
+    value is `false`.
+
+    The {{doComplexOpWithEvents()}} method MUST act as follows:
+    
+    1. If {{MyObject}}.[[{{started}}]] is not `false` then throw an {{InvalidStateError}}
```
domenic:
```suggestion
    1. If this.[[{{started}}]] is not `false` then throw an {{InvalidStateError}}
```

MyObject.[[started]] makes it a static property shared by the constructor; I think you want the instance instead.

Same throughout the rest of the algorithm.

index.bs
```diff
@@ -679,11 +680,91 @@ are traditionally added to {{WindowEventHandlers}} rather than {{Window}}.</p>
...
+    or as a stack-based variation in the [=focus update steps=])
+    for guarding against unplanned recursion. 
+    
+    In this example, an object of type <dfn noexport interface>MyObject</dfn> has
+    a method <dfn method for="MyObject">doComplexOpWithEvents()</dfn>. {{MyObject}}
+    also has an internal state attribute
+    {{MyObject}}.[[<dfn noexport attribute for="MyObject">started</dfn>]] whose initial
+    value is `false`.
+
+    The {{doComplexOpWithEvents()}} method MUST act as follows:
```
domenic:
```suggestion
    The {{doComplexOpWithEvents()}} method must act as follows:
```

index.bs
```diff
@@ -679,11 +680,91 @@ are traditionally added to {{WindowEventHandlers}} rather than {{Window}}.</p>
...
+
+    4. [=Fire an event=] named `"currentlyinprogress"` at {{MyObject}}.
+
+    5. ...Finish up complex stuff...
+
+    6. Set {{MyObject}}.[[{{started}}]] to `false`.
+</div>
+
+Sometimes "guarding" as illustrated above cannot be done because of the complexity
+of the algorithm, the number of potential algorithm entry-points or the side-effects
```
domenic:
```suggestion
of the algorithm, the number of potential algorithm entry-points, or the side-effects
```

index.bs
```diff
@@ -679,11 +680,91 @@ are traditionally added to {{WindowEventHandlers}} rather than {{Window}}.</p>
...
+
+* all state management can be rationalized within one "turn" of the algorithm (no
+    need to consider any state changes or API invocations between the time the algorithm
+    completes and when the event fires).
+* events can reveal internal state changes as early as possible when dispatched immediately
+    in the algorithm (resulting in minimal delay to user code that wants to act on
+    that state).
+
+Deferral as an option generally means *all* API entrypoints need guards, which is
+usually more work for the spec author. If there seems to be no need for guards, then
```
domenic: I'm not sure what this sentence means. Is it getting at the fact that the state of the world may have changed between the state changing and the event firing, and thus you'll need more guards of some sort to deal with such state changes? It's not very clear...

index.bs
travisleithead: Thanks!

index.bs
travisleithead: Yep, will update.

index.bs
travisleithead: 👍

index.bs
travisleithead: I spent some time thinking about what I meant--and it took a lot more words, but I think the result is better and more clear.

index.bs
```diff
@@ -679,11 +680,117 @@ are traditionally added to {{WindowEventHandlers}} rather than {{Window}}.</p>
...
+    For exmple, the {{ProgressEvent}} adds {{ProgressEvent/loaded}}, 
+    {{ProgressEvent/total}}, etc. attributes to hold the state.
+* any coordination needed among parts of an algorithm using deferred events often
+    requires defining an explicit state machine (well-defined state transitions) to
+    ensure that when a deferred event fires, the behavior of inspecting or changing
+    state is well-defined. For example, in [[payment-request]], the {{PaymentRequest}}'s
+    \[[state]] internal slot explicitly tracks the object's state
+    through its well-defined transitions.
+    * in addition to defining state transitions, each coordinated algorithm usually 
+        applies the gaurding technique (anyway) to ensure the algorithm can only proceed
```
domenic:
```suggestion
        applies the guarding technique (anyway) to ensure the algorithm can only proceed
```

index.bs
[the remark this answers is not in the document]
travisleithead: Hopefully the list of things to think about will be helpful. Agree that it's not strictly cut and dry.

index.bs
travisleithead: Thanks.

travisleithead:
I would like help with how to reference [[state]] from payment-request, e.g., https://www.w3.org/TR/payment-request/#dfn-state

I couldn't figure how to get bikeshed to find that... Any ideas?

domenic:
Since it's not exported, you'll need to do things somewhat manually. `<a spec="payment-request">\[[state]]</a>` might work, or if not, then you'd need to add something to the anchors block at the top.

travisleithead:
Super! @torgo looking for additional TAG review and/or merge help (I don't have merge access anymore ;)

dbaron:
Is it worth mentioning explicitly that if the algorithm that triggers the events might be on a different thread or process, then queuing tasks is required?

index.bs
```diff
@@ -679,11 +681,117 @@ are traditionally added to {{WindowEventHandlers}} rather than {{Window}}.</p>
...
+Deferral ensures that any stack-based recursion is avoided (but does not eliminate
+potentially problematic loops, as they could now occur as unending follow-up tasks).
+
+Defering an event is often specified as "[=queue a task=] to [=fire an event=]...".
+
+By way of illustration, the deprecated
+[Mutation Events](https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Mutation_events),
+were specified to fire in the middle of node insertion and removal algorithms, which
+are triggered by a variety of common API calls like {{Node/appendChild()}}. The side-effects
+from recursivly running node mutation algorithms caused many years of security issues.
```
dbaron: recursively (spelling)

index.bs
```diff
@@ -679,11 +681,117 @@ are traditionally added to {{WindowEventHandlers}} rather than {{Window}}.</p>
...
+    to hold such state snapshots).
+
+Conversely, when deferring, event handlers will run after the algorithm ends (or starts 
+to run [=in parallel=]) and any number of other tasks or microtasks may run in between 
+that invalidate the object's state. Since the object's state will be unknown when
+the deffered event is dispatched:
+
+* state relevant to the event should be packaged with the deferred event, usually 
+    involving a new {{Event}}-derived type with new attributes to hold the state.
+    For exmple, the {{ProgressEvent}} adds {{ProgressEvent/loaded}},
```
dbaron: example (spelling)

travisleithead:
> Is it worth mentioning explicitly that if the algorithm that triggers the events might be on a different thread or process, then queuing tasks is required?

Yes, that would be a good addition.

index.bs
```diff
@@ -679,11 +681,117 @@ are traditionally added to {{WindowEventHandlers}} rather than {{Window}}.</p>
...
+were specified to fire in the middle of node insertion and removal algorithms, which
+are triggered by a variety of common API calls like {{Node/appendChild()}}. The side-effects
+from recursivly running node mutation algorithms caused many years of security issues.
+A more robust approach is embodied in
+{{MutationObserver}}
+applies the principle of deferral of notification — in this case to the microtask queue
+following the algorithm.
+
+Weigh the trade-offs when considering deferral as an alternative to guarding as noted
+earlier. Guarding is generally the preferred approach. With guarding:
```
dbaron: I'm a little nervous about giving advice this strong about preferring guarding.  In particular, I think having an algorithm unexpectedly throw a "you can't do this now" exception can be problematic (both in terms of handling the exception, and in terms of the algorithm that needed to be run not actually being run).  It seems like some of the cases where such exceptions would be fired would be caught quickly in testing... but on the other hand I suspect some of them wouldn't be.  Perhaps it's at least worth mentioning this downside a little bit more?

travisleithead: Can you clarify a bit more how this concern is different from any algorithm that wants to unexpectedly throw an exception? I agree that running afoul of the guard can be unexpected, but (at least in this section) the guard is there to prevent unwanted recursion into the algorithm, which usually means you'll still be executing the algorithm, just not the recursive part. (Also IIRC, in some cases the guard will simply terminate an algorithm without throwing an exception.)

dbaron: The concern about bailing out generally (whether an exception or not):  If the algorithm is one that's needed to ensure some aspect of the application's state is up-to-date, failing to run the algorithm could lead to that state being out-of-date for an extended period of time, leading to user-visible bugs (possibly only in edge cases).  (Then again, there are similar sequences of evets that lead to user-visible bugs only in edge cases with the deferral approach.)

Specifically for exceptions: there are many calls where developers don't need to consider the possibility of exceptions.  There are some that throw exceptions rarely where one would reasonably expect errors to occur some of the time (e.g., network failures), but it seems less then ideal for things to fail a low portion of the time in ways that people wouldn't expect.

travisleithead: I think this is a fair caution to express, lest guarding be seen as almost always the right solution. I'll try to condense it down.

travisleithead: I added two cautionary notes below the example algorithm that I think touch on these concerns. I also softened the language about recommending guarding to a more neutral "there are tradeoffs to both approaches" which leaves the reader with a little less prescriptive guidance and more opportunity to think about which approach works given their situation.

PR #155, opened 2020-02-07
Add sections on the Web being safe to browse, trusted user interface, and asking users for consent.

### LINKED ISSUE #146: Don't Erode Users' Trust in the Web
@dbaron wrote, in [a comment](https://github.com/w3ctag/security-questionnaire/issues/48#issuecomment-495566652) on w3ctag/security-questionnaire#48:

> I also think there's a third piece that needs to be considered:  not just the specific security risks added from one specific feature, but the effect of adding those security risks to the overall picture that users have about the idea that it is safe to visit websites.  A big piece of the web's value is that it is safe for users to visit websites.  In order for this value to exist, users must *understand* that it is safer to visit websites than to install apps or software, and act based on that understanding.  Doing things that complicate that understanding or complicate what users need to understand about the safety of the web (e.g., adding features that are less safe) reduces the ability of users to act based on that understanding of safety, or to act in ways that correctly reflect the safety that exists.

[I think](https://twitter.com/hober/status/1134509819937906688) this is a sufficiently important insight that we should capture it somewhere, either here in our Design Principles (this issue), in our Ethical Web Principles finding (w3ctag/ethical-web-principles#20), or in a resurrected HTML Design Principles finding (w3ctag/design-reviews#426).

dbaron:
This is a very rough draft of changes to add sections on the Web being safe to browse, trusted user interface, and asking users for consent.  It is intended to fix #146, though it's possible I've gone a little bit beyond my remit in that issue.

This contains a bunch of ideas that I think are widely accepted, but also some that are probably less so, and some that could certainly use more research on what users *actually* understand about the web's safety.  (I thought about whether to explicitly call out the desire for research on that, but so far haven't done so.)

I think this will likely need a considerable amount of review and refinement.  I suspect if I kept it on my computer until next week and then looked at it again, I'd want to rewrite most or all of it.  But nevertheless I think it's useful to get some feedback on it, particularly on the high-level aspects:  are these the right topics to bring up here, the right way to organize them, and the right points to make?

I've picked a few reviewers who I think are likely to have comments, but I'd welcome comments from others as well.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -50,23 +50,19 @@ url: https://dom.spec.whatwg.org/#ref-for-concept-getelementsbytagname; type: in
 
 Hyperlinks, links from one page to another, are one of the foundations of the Web.
 
-Following a link, or visiting a web page, should be safe.
-Users who are following a link to a web page they haven't seen before
+Following a link, or visiting a web page, should be safe:
+users doing this
 should not have to fear for the security of their computer
 or for essential aspects of their privacy.
-(One caveat here is that users of the Web
-should not assume that they are free from consequences
+(But it's not completely safe,
+in the sense that users may face consequences
 if their use of the Web is harming others.)
+Furthermore, users should understand that it is safe (and how it isn't)
+so they can make informed decisions between use of the Web versus other technologies.
 
 Saying “essential aspects” here skips over quite a bit of detail.
-The Web isn't perfectly private.
-But we can probably make it better than it is today.
-In doing so, we may want to focus on the cases where
-user expectations of what information is revealed when using the Web
-don't match the technical details.
-We might be able to use the design of user interfaces
-to make it clear what users of the Web should expect,
-since one aspect of privacy problems is when reality doesn't meet expectations.
+The Web today is far from being perfectly private.
+One aspect of privacy problems is when reality doesn't meet expectations.
 For example,
 a person walking down the street generally expects to be recognized by their friends,
 but depending on the country,
@@ -74,6 +70,16 @@ may not expect that they were walking down that street to be recorded in a perma
 Online, people might have less understanding of what is or isn't possible,
 and their expectations might differ more widely.
 <!-- Can we cite research on this??? -->
+
+We can probably make privacy on the Web better than it is today:
+  * We can improve the user interfaces through which the Web is used
+    to make it clearer what users of the Web should expect.
+  * We can change the technical foundations of the Web
+    so that they match user expectations of privacy.
+  * We can consider the cases where users would be better off
+    if expectations were higher,
+    and in those cases
+    try to change both technical foundations and expectations
 
 When we add new features to the Web
 that might weaken the security or privacy characteristics that the Web currently has,
@@ -88,7 +94,8 @@ and this erodes one of the core values of the Web:  that it is safe to visit web
 We should seek to give users a true picture of the safety of the Web
 so that they can act based on that understanding of safety
 (in comparison to the safety of alternatives),
-and at the same time we should seek to keep the Web safe for its users.
+and at the same time we should seek to keep the Web safe for its users
+and work to make it safer.
 
 <h3 id="trusted-ui">Trusted user interface should be trustworthy</h3>
 
@@ -99,29 +106,39 @@ such as by displaying part or all of the URL or other information about the site
 or whether the connection is secure.
 Since users rely on this information to learn what site they're on
 and to make judgments about whether it is trustworthy,
-it is important that sites not be able to spoof or override this user interface.
-
-When we add new features to the web,
+this software is designed so that
+sites are not able to override this user interface
+or spoof it in a way that would confuse users.
+Therefore, when we add new features to the web,
 we should consider whether they require new complexity in this user interface,
 and whether that complexity would reduce users' ability
 to make correct judgments about who they're interacting with
-and whether the user interface is trustworthy or spoofed.
+or whether the user interface is trustworthy or spoofed.
 
 <h3 id="informed-consent">Ask users for informed consent when appropriate</h3>
 
 In some cases,
 we might add features to the Web that are not appropriate to allow for any website a user visits,
-but that are reasonable to have if the user gives informed consent.
-An example of such a feature might be support for location:
+but that are reasonable to have if the user agrees to use the feature
+based on a reasonable understanding of what will happen as a result.
+One example of such a feature might be support for location:
 many users seem likely to understand what it means to share their current location with a website
 and be able to consent to doing so
 (even though they might not fully understand the privacy implications of doing so).
+Another example of such a feature would be camera access,
+where users understand that they're sharing the image
+in the rectangle visible to the camera
+(even if they might not understand everything someone might do with the data).
 
 We should not depend on asking the user for consent
 (via a permission prompt or other mechanism)
 if we don't have a way to express that request in a way
 that users will understand what is being requested
 and the main implications of giving their consent.
+Understanding whether this is the case may require
+studying users understanding of potential user interfaces,
+although in many cases experienced user experience designers
+may be able to extrapolate from previous experience.
 
 We should also not ask users to consent to something
 that can also happen if the user does not consent.
@@ -138,7 +155,7 @@ which they learned from trusted user interface,
 did not match reality.
 
 Asking users for consent via permission prompts
-can both reinforce the idea that the web is safe
+can both reinforce the idea that <a href="#safe-to-browse">the web is safe</a>
 by showing the user that certain things won't happen without their permission,
 and can also show how scary a place the web is
 by showing how many sites are willing to ask for intrusive and unnecessary permissions.
--- a/index.bs
+++ b/index.bs
@@ -116,7 +116,7 @@ and whether that complexity would reduce users' ability
 to make correct judgments about who they're interacting with
 or whether the user interface is trustworthy or spoofed.
 
-<h3 id="informed-consent">Ask users for informed consent when appropriate</h3>
+<h3 id="consent">Ask users for meaningful consent when appropriate</h3>
 
 In some cases,
 we might add features to the Web that are not appropriate to allow for any website a user visits,
@@ -156,9 +156,9 @@ which they learned from trusted user interface,
 did not match reality.
 
 Asking users for consent via permission prompts
-can both reinforce the idea that <a href="#safe-to-browse">the web is safe</a>
-by showing the user that certain things won't happen without their permission,
-and can also show how scary a place the web is
+can reinforce the idea that <a href="#safe-to-browse">the web is safe</a>
+by showing the user that certain things won't happen without their permission.
+But frequently asking users for consent can also show how scary a place the web is
 by showing how many sites are willing to ask for intrusive and unnecessary permissions.
 
 <h2 id="js">JavaScript Language</h2>
--- a/index.bs
+++ b/index.bs
@@ -65,8 +65,9 @@ The Web today is far from being perfectly private.
 One aspect of privacy problems is when reality doesn't meet expectations.
 For example,
 a person walking down the street generally expects to be recognized by their friends,
-but depending on the country,
-may not expect that they were walking down that street to be recorded in a permanent government database.
+but (depending on the country)
+may not expect that they walked down that street at that time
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -44,6 +44,105 @@ url: https://dom.spec.whatwg.org/#ref-for-concept-getelementsbytagname; type: in
...
+
+<h3 id="safe-to-browse">It should be safe to visit a web page</h3>
+
+Hyperlinks, links from one page to another, are one of the foundations of the Web.
+
+Following a link, or visiting a web page, should be safe.
+Users who are following a link to a web page they haven't seen before
+should not have to fear for the security of their computer
+or for essential aspects of their privacy.
+(One caveat here is that users of the Web
```
alice: I'm not sure I understand this caveat. I agree with the general principle, of course, but I'm not sure what it adds here - to me it reads as though there can/should be escape hatches from users' privacy protections in cases where the user is harming others, which I'm sure is not the intended reading.

index.bs
```diff
@@ -44,6 +44,105 @@ url: https://dom.spec.whatwg.org/#ref-for-concept-getelementsbytagname; type: in
...
+should not assume that they are free from consequences
+if their use of the Web is harming others.)
+
+Saying “essential aspects” here skips over quite a bit of detail.
```
alice: I'm also a little unclear on what this paragraph is saying. What user interfaces are you referring to? Can we rephrase this to focus on the point about user expectations and privacy, and either omit the aspirational aspects or give them their own paragraph?

index.bs
```diff
@@ -44,6 +44,105 @@ url: https://dom.spec.whatwg.org/#ref-for-concept-getelementsbytagname; type: in
...
+a person walking down the street generally expects to be recognized by their friends,
+but depending on the country,
+may not expect that they were walking down that street to be recorded in a permanent government database.
+Online, people might have less understanding of what is or isn't possible,
+and their expectations might differ more widely.
+<!-- Can we cite research on this??? -->
+
+When we add new features to the Web
+that might weaken the security or privacy characteristics that the Web currently has,
+we should consider the tradeoffs involved in that particular feature.
```
alice: Perhaps this is worthy of its own sub-heading?

index.bs
```diff
@@ -44,6 +44,105 @@ url: https://dom.spec.whatwg.org/#ref-for-concept-getelementsbytagname; type: in
...
+<h3 id="trusted-ui">Trusted user interface should be trustworthy</h3>
+
+Since users of the Web can follow links to sites they might not know already,
+software that lets them access the Web generally provides user interface
+to show the user who they're interacting with and how,
+such as by displaying part or all of the URL or other information about the site,
+or whether the connection is secure.
+Since users rely on this information to learn what site they're on
+and to make judgments about whether it is trustworthy,
+it is important that sites not be able to spoof or override this user interface.
```
alice: This seems out of scope for API design, albeit a critical point for user agent UI design.

It seems like the point below is the more critical one: if a feature relies on adding extra decisions for a user to make, care should be taken that the feature is providing enough value (either through providing a clear user benefit, or through simplifying some other decision process for the user) to make it worthwhile.

index.bs
```diff
@@ -44,6 +44,105 @@ url: https://dom.spec.whatwg.org/#ref-for-concept-getelementsbytagname; type: in
...
+and be able to consent to doing so
+(even though they might not fully understand the privacy implications of doing so).
+
+We should not depend on asking the user for consent
+(via a permission prompt or other mechanism)
+if we don't have a way to express that request in a way
+that users will understand what is being requested
+and the main implications of giving their consent.
+
+We should also not ask users to consent to something
```
alice: Is this worth a separate sub-heading as well?

index.bs
```diff
@@ -44,6 +44,105 @@ url: https://dom.spec.whatwg.org/#ref-for-concept-getelementsbytagname; type: in
...
+
+<h3 id="informed-consent">Ask users for informed consent when appropriate</h3>
+
+In some cases,
+we might add features to the Web that are not appropriate to allow for any website a user visits,
+but that are reasonable to have if the user gives informed consent.
+An example of such a feature might be support for location:
+many users seem likely to understand what it means to share their current location with a website
+and be able to consent to doing so
+(even though they might not fully understand the privacy implications of doing so).
```
alice: Is this an editorial on the geolocation API?

index.bs
```diff
@@ -44,6 +44,105 @@ url: https://dom.spec.whatwg.org/#ref-for-concept-getelementsbytagname; type: in
...
+
+We should not depend on asking the user for consent
+(via a permission prompt or other mechanism)
+if we don't have a way to express that request in a way
+that users will understand what is being requested
```
alice: Which users? How do we determine what the "target audience" users will or won't be able to understand?

index.bs
```diff
@@ -44,6 +44,105 @@ url: https://dom.spec.whatwg.org/#ref-for-concept-getelementsbytagname; type: in
...
+for something that they can't effectively prevent.
+Doing so would send the wrong message about the safety characteristics of the Web
+and would lead users to be disappointed when the expectations of privacy,
+which they learned from trusted user interface,
+did not match reality.
+
+Asking users for consent via permission prompts
+can both reinforce the idea that the web is safe
+by showing the user that certain things won't happen without their permission,
+and can also show how scary a place the web is
```
alice: Is this a risk we're warning about?

index.bs
hober: I see your point, @alice, but I think it's in scope to the extent that different API shapes make different UI designs easier / more likely / etc.

alice: Perhaps that point should be made more explicitly, in that case?

hober: Agreed!

index.bs
dbaron: So the caveat is really to the sentence before the quote -- the idea that following a link should always be safe.  Taking that position absolutely would imply that we're taking the position that governments shouldn't punish people for, say, viewing child pornography.  And I don't think that's a position that I'd like to take.  So I want some way to say that we want it to be safe from a technical perspective but not necessarily from a legal one.

(At the same time, there's an entirely separate discussion about mechanisms that governments want to use to enforce some legal rules, and what the downsides of those mechanisms are, and while some of that discussion is relevant to the TAG, I'd like to avoid getting into it in this section.)

I should think more about a better way to word this.

index.bs
dbaron: I was trying to say that I think the geolocation API is ok since it's a question that users can understand, even if they don't necessarily follow all the implications of how location tracking is used in today's ecosystem of tracking/sharing/reselling user data.

index.bs
```diff
@@ -44,6 +44,105 @@ url: https://dom.spec.whatwg.org/#ref-for-concept-getelementsbytagname; type: in
...
+In doing so, we may want to focus on the cases where
+user expectations of what information is revealed when using the Web
+don't match the technical details.
+We might be able to use the design of user interfaces
+to make it clear what users of the Web should expect,
+since one aspect of privacy problems is when reality doesn't meet expectations.
+For example,
+a person walking down the street generally expects to be recognized by their friends,
+but depending on the country,
+may not expect that they were walking down that street to be recorded in a permanent government database.
```
hober: this sentence reads strangely to me. I think I know what you're trying to say, but this is a confusing way to say it.

Does this read better to anyone else:

> For example, a person walking down the street generally expects to be recognized by their friends but,
> depending on the country,
> they may not expect the fact that they walked down that particular street at that particular time to be recorded in a permanent government database.

index.bs
```diff
@@ -44,6 +44,105 @@ url: https://dom.spec.whatwg.org/#ref-for-concept-getelementsbytagname; type: in
...
+and to make judgments about whether it is trustworthy,
+it is important that sites not be able to spoof or override this user interface.
+
+When we add new features to the web,
+we should consider whether they require new complexity in this user interface,
+and whether that complexity would reduce users' ability
+to make correct judgments about who they're interacting with
+and whether the user interface is trustworthy or spoofed.
+
+<h3 id="informed-consent">Ask users for informed consent when appropriate</h3>
```
hober: Internally on the WebKit team, we tend to talk about "meaningfully informed consent", by which we mean that not only was the user informed of the risks, they were informed of the risks in a way we believe they're likely to be able to understand and make sound judgments based upon. I don't know if this is a distinction worth making here.

index.bs
hober: Does this paragraph argue against the existence of private browsing mode? Users sometimes think that there will be no *server-side* record of what they did while visiting a site with private browsing mode on.

index.bs
alice: https://github.com/w3ctag/design-reviews/issues/461#issuecomment-587618596 this seems like a good first pass.

index.bs
dbaron: Sorry for the delay getting back to these -- I'm going to post revisions soon (I hope) that will address many of the comments.

But while I'm working on that, I wanted to get back to this one:  I think this paragraph is still pretty central to the idea that it should be safe to visit a web page, and I think I'd like to keep them together in the same subsection.  (But perhaps some of the other edits I'm currently making will make this a little clearer.)

index.bs
dbaron: I'm a little confused by @alice's last comment in this thread -- was it meant as a comment on the following section (on informed consent), rather than this section on trusted UI?

(Also, I think the "point below" in the initial comment was referring to the second paragraph of this section, and not the following section, but please correct me if I'm wrong.)

In any case, what I've tried to do here (and I realize I should comment before pushing the changes since pushing the changes will likely hide the review comments) in convert these two paragraphs into one and connect them more closely.  Though perhaps you were looking for something different?

index.bs
dbaron: To @alice's question of separate subheading starting with the third paragraph of this section -- I think both the second and third paragraphs of this section are trying to define the "when appropriate" in the section's title, so I prefer not.

Regarding @hober's question on private browsing mode:  I don't think it argues against the existence of such a mode, but I would argue (although this is more tied to the first of the three subsections) that implementations offering such a mode ought to try to adjust its user interface so that users have more accurate expectations of what it does and doesn't do.

index.bs
dbaron: I'll expand on this based on https://github.com/w3ctag/design-reviews/issues/461#issuecomment-587852956 and https://github.com/w3ctag/design-reviews/issues/461#issuecomment-587895409 .

index.bs
dbaron: Maybe.  I fully admit it's a little bit off-topic for the section, but I think it's a risk worth mentioning.  But I'm open to taking it out if you think it doesn't fit.

index.bs
dbaron: I've tried to do this; I think I pulled the aspirational bits out into their own paragraph, which I actually ended up structuring as a list.

index.bs
dbaron: I agree it's better.  I took most of your suggestions but turned "depending on the country" into a parenthetical and omitted a few words ("the fact", "particular", "particular").

index.bs
dbaron: Maybe, though I'm actually worried, based on its definition in other contexts (particularly medical ones), that the term "informed consent" might even be too strong, since it seems to imply being properly informed of all risks.  Does "meaningful consent" make sense?  I've left it alone for now, but I'm definitely open to changing it.

dbaron:
Also, I've made [a live version of the changes on this branch](https://dbaron.github.io/design-principles/#api-principles) visible.

index.bs
alice: I like "meaningful consent"

index.bs
alice: I think the parenthetical might be best as its own clarifying sentence, in that case.

index.bs
alice: Perhaps it should be rewritten as two sentences illustrating two sides of a trade-off.

index.bs
dbaron: Perhaps, but I'm going to leave this one as is for now.

index.bs
dbaron: I changed it to 2 sentences.

index.bs
dbaron: I changed it to "meaningful consent" (and the `id` to "consent").

Issue #108, opened 2018-10-30
Add naming advice?


kenchris:
Like

- Use US English
- Use ASCII (No emoji etc)

atanassov:
This feedback has been addressed.

PR #159, opened 2020-03-03
API naming advise about language and charset feedback in #108

atanassov:
Following up on issue https://github.com/w3ctag/design-principles/issues/108. We are adding prose for the language and character set to be used.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -138,7 +138,7 @@ and by name association.
 Please consult widely on names in your APIs.
 
 API naming *must* be done in easily readable US-English language and fall into the ASCII range. 
-For example `horizongalAlignment` is more English-readible name than `alignmentHorizontal`. 
+For example `horizontalAlignment` is a more English-readable name than `alignmentHorizontal`. 
 Whenever possible consider readability over brevity.
 
 Boolean properties, options, or API parameters which are asking a question about
@@ -1234,4 +1234,3 @@ so that readers can quickly decide whether they need to read the steps in detail
 url: https://w3c.github.io/hr-time/#dom-domhighrestimestamp; spec: HIGHRES-TIME; type: typedef
     text: DOMHighResTimeStamp
 </pre>
-
--- a/index.bs
+++ b/index.bs
@@ -137,6 +137,10 @@ and by name association.
 
 Please consult widely on names in your APIs.
 
+API naming *must* be done in easily readable US-English language and fall into the ASCII range. 
+For example `horizongalAlignment` is more English-readible name than `alignmentHorizontal`. 
+Whenever possible consider readability over brevity.
+
 Boolean properties, options, or API parameters which are asking a question about
 their argument *should not* be prefixed with <code>is</code>, while methods
 that serve the same purpose, given that it has no side effects, *should* be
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -137,6 +137,10 @@ and by name association.
 
 Please consult widely on names in your APIs.
 
+API naming *must* be done in easily readable US-English language and fall into the ASCII range. 
+For example `horizongalAlignment` is more English-readible name than `alignmentHorizontal`.
```
annevk:
```suggestion
For example `horizontalAlignment` is a more English-readable name than `alignmentHorizontal`. 
```

annevk:
I think the rebase failed unfortunately and now it's changing more than it should.

index.bs
```diff
@@ -425,7 +432,61 @@ the API visibly relates to garbage collection details. By contrast, special APIs
 purpose such as `WeakRef` and `FinalizationGroup` have their GC interaction as a specific part of
 their contract with the developer.
 
-<h2 id="api-surface">JavaScript API Surface Concerns</h2>
+<h2 id="api-surface">API Surface Concerns</h2>
```
hober: this heading rename shouldn't happen

index.bs
```diff
@@ -425,7 +432,61 @@ the API visibly relates to garbage collection details. By contrast, special APIs
...
+
+<h3 id="naming-is-hard">Naming things</h3>
```
hober: please don't duplicate this whole section

index.bs
atanassov: Addressed with the rebase.

index.bs
atanassov: Addressed with the rebase.

index.bs
```diff
@@ -4,9 +4,9 @@ Group: W3C TAG
 Shortname: design-principles
 Repository: w3ctag/design-principles
 Status: DREAM
-Editor: Travis Leithead, Microsoft, travil@microsoft.com
```
hober: Why is this change in here? Did something go wrong with the rebase?

atanassov: Yeah, moved to Former Editor: - already part of master

hober: If it's already in master, why is it showing up in this diff?

annevk:
You also removed some of the earlier fixes I suggested.

index.bs
atanassov: Added the dropped 'a' from @annevk change and got rid of the editor's list change which is in master and not sure why it was flagged as diff here...

dbaron:
With this kind of PR, it's better (next time) to squash rather than merge, so you don't have all of that messy history permanently in the repository.

dbaron:
@atanassov force-pushed to revert from 68b84b366f4b1bf3dcbf1c746693f0d31ed8ec57 back to 82dea718c47a6f4a9e3dc9541a0c88a4cf55186e, so this needs to get re-landed.  Not sure how to get github to recognize that, though.

atanassov:
I did manually re-land the squashed commit to master as b66b7f45c047adf2ad63f5902a875d1f1bb65f82 
Next, we should change this repo policy to have squash-merge as the default option :)

PR #163, opened 2020-03-04
Naming things: expand language around the use of common words, brevity, wide consultation, and future-proofing.

hober:


index.bs
```diff
@@ -165,37 +165,61 @@ by showing how many sites are willing to ask for intrusive and unnecessary permi
...
+in the specification where the API is defined;
+this principle applies only to the API names themselves.
 
-API naming *must* be done in easily readable US English. For example `horizontalAlignment` 
-is a more English-readable name than `alignmentHorizontal`. Whenever possible consider 
-readability over brevity.
+<h4 id="naming-charset" class="no-num no-toc">Use ASCII names</h4>
 
-Names should adhere to the local language restrictions, for example CSS ident rules etc. 
+Names must adhere to the local language restrictions, for example CSS ident rules etc.
```
hober: @atanassov, is this s/should/must/ okay?

atanassov: When we originally discussed it we decided to use 'should' here. Looking at it further I don't see a reason we shouldn't make it stronger, thus I'm good with the 'must'

hober:
Okay, this has been rebased onto @atanassov's squashed merge. Please review.

index.bs
```diff
@@ -165,37 +165,61 @@ by showing how many sites are willing to ask for intrusive and unnecessary permi
 
 <h3 id="naming-is-hard">Naming things</h3>
 
-Naming is hard! We would all like a silver-bullet for naming APIs...
-
 Names take meaning from:
 
 * signposting (the name itself)
 * use (how people come to understand the name over time)
 * context (the object on the left-hand side, for example)
```
atanassov: Should we add more descriptive suggestions about naming here - describe _what_ an API does and not _how_ etc.

atanassov: ... or maybe make add that in the Future-proofing section?

index.bs
```diff
@@ -165,37 +165,61 @@ by showing how many sites are willing to ask for intrusive and unnecessary permi
...
-Consistency is a good principle that helps to create a platform that users can navigate intuitively
-and by name association.
+<h4 id="naming-common-words" class="no-num no-toc">Use common words</h4>
 
-Please consult widely on names in your APIs.
+API naming *must* be done in easily readable US English.
+Keep in mind that most web developers are not native English speakers.
+Whenever possible, names should be chosen that use common vocabulary
+non-native English speakers are likely to understand when first encountering the name.
+Always value readability over brevity.
```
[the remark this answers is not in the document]
hober: We used to have a different example, which I dropped in favor of yours. I'm happy to re-add it.

index.bs
```diff
@@ -165,37 +165,61 @@ by showing how many sites are willing to ask for intrusive and unnecessary permi
...
 * signposting (the name itself)
 * use (how people come to understand the name over time)
 * context (the object on the left-hand side, for example)
 
-Consistency is a good principle that helps to create a platform that users can navigate intuitively
-and by name association.
+<h4 id="naming-common-words" class="no-num no-toc">Use common words</h4>
 
-Please consult widely on names in your APIs.
+API naming *must* be done in easily readable US English.
```
annevk: I'd prefer if we used something more specific than "easy".

index.bs
```diff
@@ -165,37 +165,61 @@ by showing how many sites are willing to ask for intrusive and unnecessary permi
...
-
 Names take meaning from:
 
 * signposting (the name itself)
 * use (how people come to understand the name over time)
 * context (the object on the left-hand side, for example)
 
-Consistency is a good principle that helps to create a platform that users can navigate intuitively
-and by name association.
+<h4 id="naming-common-words" class="no-num no-toc">Use common words</h4>
```
annevk: Why are these all no-num/no-toc?

index.bs
annevk: I'm not sure I entirely agree with this. E.g., how would this judge `fetch()` or `response.json()`?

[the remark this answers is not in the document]
atanassov: Tess, you're right... What threw me off was the paragraph below that doesn't appear to be properly formatted. 
see ln 182 "<p class="example"For example...."

atanassov: Applying any or all of these rules retrospectively may not necessary work. In your example I'd argue that `fetch()` is a common vocabulary for getting things on one's behalf. On the other hand, `json` is a well known reference similar to `style`, `css` etc. that shouldn't need a long form... if the format isn't known to the user it should be trivial to find of course.

index.bs
atanassov: What would you recommend?

index.bs
annevk: Well, it might not be clear that it `json()` consumes the body of the response though. That's more the thing I was getting it. But also, I think that there's value in brevity as it's never possible to encapsulate the full complexity in the name.

index.bs
dbaron: Given the structure of the document, I think it makes sense to have them `no-num no-toc`

index.bs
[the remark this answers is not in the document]
hober: Fixed in 65012b8.

index.bs
hober: Yeah, this is @atanassov's text from #159. Let's fix in a follow-up.

index.bs
```diff
@@ -165,37 +165,61 @@ by showing how many sites are willing to ask for intrusive and unnecessary permi
...
-readability over brevity.
+<h4 id="naming-charset" class="no-num no-toc">Use ASCII names</h4>
 
-Names should adhere to the local language restrictions, for example CSS ident rules etc. 
+Names must adhere to the local language restrictions, for example CSS ident rules etc.
 and *should* be in the [ASCII range](https://infra.spec.whatwg.org/#ascii-code-point).
 
+<h4 id="naming-consultation" class="no-num no-toc">Consultation</h4>
+
+Please consult widely on names in your APIs.
```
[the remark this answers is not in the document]
dbaron: We had some discussion on this in today's teleconference.  One of the issues with this is that it can tend to lead to a lot of bikeshedding.  It may be useful to describe what advice is useful to listen to, e.g., perhaps advice based on underlying principles (perhaps the ones in this document, perhaps more domain-specific ones).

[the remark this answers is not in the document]
hober: I've tried to capture this in 2f9ba01, @atanassov & @dbaron.

index.bs
hober: I've tried to capture this in d59b3a7, @atanassov & @annevk.

index.bs
hober: I think that's a good idea, but I'd rather not hold up this PR on it. Can you file an followup, @atanassov?

index.bs
```diff
@@ -182,37 +182,100 @@ by showing how many sites are willing to ask for intrusive and unnecessary permi
...
+have relied on this research
+when editing that document. [[URL]]
+
+</div>
+
+<h4 id="naming-future-proofing" class="no-num no-toc">Future-proofing</h4>
+
+Naming should be generic and future-proof whenever possible.
+
+The name should not be directly associated with a brand or specific revision of
```
alice, line 258: Do we have an example for this one? A name which might have been technology-specific, but isn't?

hober: I've added two examples in da48077.

hober:
@annevk, could you take another look? Thanks!

torgo:
Approved on [call 6 May 2020](https://github.com/w3ctag/meetings/blob/gh-pages/2020/telcons/05-04-agenda.md)

PR #164, opened 2020-03-17
Add the Priority of constituencies principle.

hober:
Part of https://github.com/w3ctag/design-principles/projects/1

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -53,7 +53,7 @@ consider users over authors over implementors over specifiers over theoretical p
 In other words costs or difficulties to the user
 should be given more weight than costs to authors;
 which in turn should be given more weight than costs to implementors;
-which should be given more weight than costs to authors of the spec itself,
+which should be given more weight than costs to authors of the specification itself,
 which should be given more weight than those proposing changes for theoretical reasons alone.
 Of course, it is preferred to make things better for multiple constituencies at once.
 
--- a/index.bs
+++ b/index.bs
@@ -46,6 +46,22 @@ url: https://dom.spec.whatwg.org/#ref-for-concept-getelementsbytagname; type: in
 
 <h2 id="basic-principles">Principles behind design of Web APIs</h2>
 
+<h3 id="priority-of-constituencies">Priority of constituencies</h3>
+
+In case of conflict,
+consider users over authors over implementors over specifiers over theoretical purity.
+In other words costs or difficulties to the user
+should be given more weight than costs to authors;
+which in turn should be given more weight than costs to implementors;
+which should be given more weight than costs to authors of the spec itself,
+which should be given more weight than those proposing changes for theoretical reasons alone.
+Of course, it is preferred to make things better for multiple constituencies at once.
+
+See also
+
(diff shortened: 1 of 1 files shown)


annevk:
Someone with admin privileges should really protect this repository...

index.bs
```diff
@@ -46,6 +46,22 @@ url: https://dom.spec.whatwg.org/#ref-for-concept-getelementsbytagname; type: in
...
 <h2 id="basic-principles">Principles behind design of Web APIs</h2>
 
+<h3 id="priority-of-constituencies">Priority of constituencies</h3>
+
+In case of conflict,
+consider users over authors over implementors over specifiers over theoretical purity.
+In other words costs or difficulties to the user
+should be given more weight than costs to authors;
+which in turn should be given more weight than costs to implementors;
+which should be given more weight than costs to authors of the spec itself,
```
annevk:
```suggestion
which should be given more weight than costs to authors of the specification itself,
```
It seems it was also spec in the original, but I think it's better to not use abbreviations.

hober:
> This looks good apart from the weird second change that is from a different PR...

Needed to rebase. Fixed.

dbaron:
So, five years and one day ago I wrote [a blog post](https://dbaron.org/log/20150318-priority-of-constituencies) about the Priority of Constituencies principle which has some comments on it.  I guess I'm curious what you think of those comments and whether they're worth incorporating.

(I think the fundamental question in the first part of that post is whether the priority of constituencies exists simply because there are more users than there are authors, more authors than there are browser implementers, etc., or for some more fundamental reason like the idea that end users should get priority because we're running things on their machines and in ways that could affect their lives.)

annevk:
I was surprised when reading cost to be fair and went to check the original principle to see if it had that too. I like replacing that with "value".

And I like the idea of enshrining the fundamental importance of end users, though perhaps that ought to be a separate principle? Or is deferring to mnot's work is sufficient?

hober:
@dbaron wrote:

> So, five years and one day ago I wrote [a blog post](https://dbaron.org/log/20150318-priority-of-constituencies) about the Priority of Constituencies principle which has some comments on it. I guess I'm curious what you think of those comments and whether they're worth incorporating.

They certainly are!

> (I think the fundamental question in the first part of that post is whether the priority of constituencies exists simply because there are more users than there are authors, more authors than there are browser implementers, etc., or for some more fundamental reason like the idea that end users should get priority because we're running things on their machines and in ways that could affect their lives.)

I think it exists for both of those reasons.

@annevk wrote:

> And I like the idea of enshrining the fundamental importance of end users, though perhaps that ought to be a separate principle? Or is deferring to mnot's work is sufficient?

I think it can be the same principle (users come first because of their fundamental importance). I think we should point to @mnot's work, for sure, but I think our stating of this principle needs to stand on its own.

hober:
> Someone with admin privileges should really protect this repository...

@dbaron [did so](https://github.com/organizations/w3ctag/settings/audit-log).

hober:
We decided to land this in our plenary this week. @dbaron, please file a follow-up issue or PR with suggested additional text.

mnot:
FWIW - for-the-users should be an RFC soon.

PR #166, opened 2020-04-02
Add a section on CSS principles. Fixes #126

### LINKED ISSUE #126: add a section on CSS to the design principles document
Currently this document doesn't say anything about CSS; I think it should.

I think there are some sources of principles that we could pull from on the CSS working group's wiki, but I might try to frame some of them at a slightly higher level.

dbaron:
This adds a somewhat substantive section on CSS to the design principles document.  It's multiple commits that are **meant to be squashed** when landed.

You should be able to [preview this here](https://dbaron.github.io/design-principles/#css) ~at some point soon, though github seems sluggish right now and it's not up yet~.

Fixes #126.

I'd like to solicit review from @tabatkins and @fantasai as well, but I can't actually mark them in reviewers.

I'm sure there's a lot that can be changed/added here -- I'd like to strike a reasonable balance between trying to polish this, and also trying to defer most "also add X" requests to later PRs.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -352,34 +352,6 @@ secure contexts reduce the risks to user privacy.
 
 <h2 id="css">Cascading Style Sheets (CSS)</h2>
 
-<h3 id="css-versus-markup">Presentation should be in CSS, content should be in HTML</h3>
-
-The separation between CSS (style) and HTML (markup) was originally designed around the idea
-that the HTML is for the *content* and
-the CSS is for the *presentation* of that content.
-The CSS was intended to be optional,
-and the HTML intended so that it could be fully understood without the CSS.
-While the idea that the CSS can be dropped and the markup can still be usable
-may not make sense for *applications* built using web technologies,
-it still seems like a separation worth maintaining for *documents*,
-which remain a major use of web technology.
-
-If non-visual consumers of a page (such as search engines or screenreaders)
-seem likely to need the information (at least for well-designed pages),
-then it probably belongs in the markup rather than in the CSS.
-
-<p class="example">For example, information about the human language (English, Chinese, Arabic, etc.)
-that the text is in belongs in the markup
-(in particular, the <code><a href="https://html.spec.whatwg.org/multipage/dom.html#the-lang-and-xml:lang-attributes">lang</a></code> attribute),
-even though a number of the uses of that information (such as font selection or hyphenation)
-are related to the presentation.
-On the other hand, information about the font size makes sense to be in the style,
-since in pages that use HTML correctly
-the information shown by font sizes should also be obtainable from the markup
-(e.g., <code highlight="html">&lt;h1&gt;</code>, <code highlight="html">&lt;strong&gt;</code>)
-even though on some poorly designed pages the notion of what is a heading might not be visible to search engines or screenreaders except through the style
-(e.g., <code highlight="html">&lt;div style="font-size: 1.8em"&gt;</code>).
-
 <h3 id="css-property-separation">Separate CSS properties based on what should cascade separately</h3>
 
 Decisions about what should be in the same CSS property or in different CSS properties
--- a/index.bs
+++ b/index.bs
@@ -353,16 +353,50 @@ secure contexts reduce the risks to user privacy.
 
 <h3 id="css-versus-markup">Presentation should be in CSS, content should be in HTML</h3>
 
-CSS originally intended as optional... maybe we've moved past that?
+The separation between CSS and HTML was originally designed around the idea
+that the HTML is for the *content* and
+the CSS is for the *presentation* of that content.
+The CSS was intended to be optional,
+and the HTML intended so that it could be fully understood without the CSS.
+While the idea that the CSS can be dropped and the markup can still be usable
+may not make sense for *applications* built using web technologies,
+it still seems like a separation worth maintaining for *documents*,
+which remain a major use of web technology.
 
-But still:
-- a11y tools
-- copy/paste
-
-Things that are essential to understanding the content should be in the markup.
-For example, the content's language belongs in the <{html/lang}> attribute, not in CSS.
+If non-visual consumers of a page (such as search engines or screenreaders)
+seem likely to need the information (at least for well-designed pages),
+then it probably belongs in the markup rather than in the CSS.
+For example, information about the human language (English, Chinese, Arabic, etc.)
+that the text is in belongs in the markup
+(in particular, the <code><a href="https://html.spec.whatwg.org/multipage/dom.html#the-lang-and-xml:lang-attributes">lang</a></code> attribute),
+even though a number of the uses of that information (such as font selection or hyphenation)
+are related to the presentation.
+On the other hand, information about the font size makes sense to be in the style,
+since in pages that use HTML correctly
+the information in font sizes should also be obtainable from the markup
+(e.g., <code highlight="html">&lt;h1&gt;</code>, <code highlight="html">&lt;strong&gt;</code>)
+even though on some poorly designed pages the notion of what is a heading might not be visible to search engines or screenreaders except through the style
+(e.g., <code highlight="html">&lt;div style="font-size: 1.8em"&gt;</code>).
 
 <h3 id="css-property-separation">Separate CSS properties based on what should cascade separately</h3>
+
+Decisions about what should be in the same CSS property or in different CSS properties
+should be based on what makes sense to set independently.
+CSS cascading allows declarations (which are property-value pairs)
+from different rules or different style sheets
+to override one another in a process called the cascade.
+Thus things that should always be set together belong in the same properties,
+but things that make sense to set in different places should be in separate properties.
+
+For example, the "size" and "sink" aspects of the 'initial-letters' property
+belong in a single property
+because they are part of a single initial letter effect
+(e.g., a drop cap, sunken cap, or raised cap)
+and should nearly always be specified together.
+However, the 'initial-letters-align' property should be separate because it
+sets an alignment policy for all of these effects across the document
+which is a general stylistic choice
+and a function of the script (e.g., Latin, Cyrillic, Arabic) used in the document.
 
 <h3 id="css-inherited-or-not">Make appropriate choices for whether CSS properties are inherited</h3>
 
--- a/index.bs
+++ b/index.bs
@@ -416,6 +416,14 @@ If setting the property on a descendant element is a separate effect
 that adds to setting it on an ancestor,
 then the property should probably not be inherited.
 
+One specific example that is common is that if a property has an effect on text,
+then it is almost always in the class where
+a descendant element needs to override (rather than add to) the effect of setting it on an ancestor,
+and the property should be inherited.
+This is also needed to maintain the design principle that
+inserting an unstyled inline element around a piece of text
+does not change the appearance of that text.
+
 A specification of an non-inherited property requiring that the handling of an element
 look at the value of that property on its ancestors (which may also be slow)
 is a "code smell" that suggests that the property likely should have been inherited.
--- a/index.bs
+++ b/index.bs
@@ -503,7 +503,7 @@ of walking up the ancestor chain looking for text decorations on ancestors.
 Another important choice in the design of every CSS property
 is what its <a>computed value</a> is.
 This is an abstract description of the value that is inherited when the property inherits.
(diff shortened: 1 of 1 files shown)


dbaron:
I realize the section on computed values should probably say something about animation as well.

index.bs
```diff
@@ -349,6 +349,218 @@ One example of a feature that should be limited to secure contexts
 is geolocation, since the authentication and confidentiality provided by
 secure contexts reduce the risks to user privacy.
 
+<h2 id="css">Cascading Style Sheets (CSS)</h2>
+
+<h3 id="css-versus-markup">Presentation should be in CSS, content should be in HTML</h3>
```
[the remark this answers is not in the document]
hober: This principle strikes me as essentially the same thing as the [Separation of Concerns](https://www.w3.org/TR/html-design-principles/#separation-of-concerns) [HTML Design Principle](https://www.w3.org/TR/html-design-principles/).

If you agree, I should track this PR as part of https://github.com/w3ctag/design-principles/projects/1, and maybe the principle should be moved under [§ 2. API Design Across Languages](https://w3ctag.github.io/design-principles/#api-across-languages).

index.bs
```diff
@@ -349,6 +349,218 @@ One example of a feature that should be limited to secure contexts
...
+
+<h3 id="css-inherited-or-not">Make appropriate choices for whether CSS properties are inherited</h3>
+
+An important choice in the design of every CSS property is whether the property is
+<a href="https://drafts.csswg.org/css-cascade-3/#inherited-property">inherited</a>.
+This affects what happens when no value is specified for the property on an element.
+If the property is not inherited, then that element gets the property's initial value.
+If the property is inherited, then that element gets the parent element's computed value.
+
+The decision about whether a property should be inherited should be based largely
```
tabatkins, line 410: I think it should be noted that, to a first approximation, inherited or not is about whether the property affects text or boxes. The number of exceptions to that rule is fairly small, and you can go into more detail about those exceptions afterwards.

index.bs
```diff
@@ -349,6 +349,218 @@ One example of a feature that should be limited to secure contexts
...
+
+Words in properties and values are separated by hyphens.
+Abbreviations are generally avoided.
+
+The list of values of a property should generally be chosen
+so that new values can be added.
+Avoid values like ''yes'', ''no'', ''true'', ''false'',
+or things with more complex names that are basically equivalent to them.
+
+See [[#naming-is-hard]] for general (cross-language) advice on naming.
```
tabatkins, line 562: I haven't followed this link yet, but maybe note also that generic words like "mode" or "state" should be avoided in property names due to being semantically meaningless, since it's more or less true by default that a property is selecting a mode or setting a state.

index.bs
```diff
@@ -349,6 +349,218 @@ One example of a feature that should be limited to secure contexts
...
+</div>
+
+<h3 id="css-computed-value">Choose the computed value type based on how the property should inherit</h3>
+
+Another important choice in the design of every CSS property
+is what its <a>computed value</a> is.
+This is an abstract description of the value that is inherited when the property inherits.
+This is most important for inherited properties, but it does matter for all properties.
+
+Due to a somewhat unfortunate history,
```
tabatkins, line 499: I'd make this paragraph a `Note:`, since it's a diversion from the point.

index.bs
```diff
@@ -349,6 +349,218 @@ One example of a feature that should be limited to secure contexts
...
+this is *not* the value that is returned by {{getComputedStyle()}},
+which is actually the <a>resolved value</a>,
+which can be the result of further computation starting from the computed value.
+
+The definition of a CSS property often has interactions with other CSS properties.
+Whether those interactions affect the property's computed value is a design choice,
+and it affects the behavior of inheritance.
```
tabatkins, line 506:
```suggestion
The definition of a CSS property often has interactions with other CSS properties.
Inheritance passes down the [=computed value=], as well,
so whether or not the interaction should affect the value that descendants inherit
is an important factor in whether the interactions are handled in the [=computed value=]
or in a later value stage such as the [=used value=].
```

index.bs
```diff
@@ -349,6 +349,218 @@ One example of a feature that should be limited to secure contexts
...
+
+<a href="https://wiki.csswg.org/spec/property-dependencies">The dependencies
+that do affect computed values</a>
+are required not be circular.
```
tabatkins, line 510:
```suggestion
are required to not be circular.
```

index.bs
```diff
@@ -349,6 +349,218 @@ One example of a feature that should be limited to secure contexts
...
+Complexity in implementations might affect how fast things run for end users.)
+
+<div class="example">
+For example, the 'line-height' property accepts values that are integers,
+such as ''line-height: 1.4''.
+This value is a multiple of the 'font-size',
+so if the 'font-size' is ''20px'' then this would mean the line height is ''28px''.
+
+However, the computed value in this case is the number ''1.4'', not the length ''28px''.
+This is important because the value can inherit to elements that have a different font size:
```
tabatkins, line 528:
```suggestion
This is important because the value can inherit to elements that have a different font size,
and preserving the "multiple of the current element's font" behavior would be broken
if it turned into a length at computed-value time:
```

index.bs
```diff
@@ -349,6 +349,218 @@ One example of a feature that should be limited to secure contexts
...
+are required not be circular.
+
+Making decisions about whether dependencies should affect computed values
+often involves tradeoffs of complexity.
+Sometimes it is less complex
+for the behavior to happen as part of the computation of the computed value;
+sometimes it is less complex 
+for the behavior to happen in whatever handles the computed value.
+(This complexity might be in the specification, in implementations, or both.
+Complexity in implementations might affect how fast things run for end users.)
```
tabatkins, line 519: Link to <https://wiki.csswg.org/spec/computed-values> for some additional details about what can and can't affect computed values?

index.bs
```diff
@@ -349,6 +349,220 @@ One example of a feature that should be limited to secure contexts
...
+The names of CSS properties are usually nouns,
+and the names of their values are usually adjectives (although sometimes nouns).
+
+Words in properties and values are separated by hyphens.
+Abbreviations are generally avoided.
+
+The list of values of a property should generally be chosen
+so that new values can be added.
+Avoid values like ''yes'', ''no'', ''true'', ''false'',
+or things with more complex names that are basically equivalent to them.
```
plinss, line 562: Wondering if this should be expanded to generic advice of avoiding booleans in other apis, or possibly advice as to when to use a boolean vs a string or enum.

atanassov:
It would be good to hear from @astearns as well

index.bs
```diff
@@ -349,6 +349,220 @@ One example of a feature that should be limited to secure contexts
...
+because they inherit better than length values.
+</div>
+
+<h3 id="css-naming">Naming of CSS properties and values</h3>
+
+The names of CSS properties are usually nouns,
+and the names of their values are usually adjectives (although sometimes nouns).
+
+Words in properties and values are separated by hyphens.
+Abbreviations are generally avoided.
```
fantasai, line 557: We also prefer to use the base semantic word form, without additional grammatical morphemes. (This isn't a strict rule, but one that's usually followed.)

index.bs
```diff
@@ -349,6 +349,220 @@ One example of a feature that should be limited to secure contexts
...
+The separation between CSS and HTML was originally designed around the idea
+that the HTML is for the *content* and
+the CSS is for the *presentation* of that content.
+The CSS was intended to be optional,
+and the HTML intended so that it could be fully understood without the CSS.
+While the idea that the CSS can be dropped and the markup can still be usable
+may not make sense for *applications* built using web technologies,
+it still seems like a separation worth maintaining for *documents*,
+which remain a major use of web technology.
+
```
fantasai, line 365: Separation of content and presentation matters for applications also. It might be much harder to understand an app without any styling, because it requires a richer semantic vocabulary than base-level HTML does, but it does need to be understandable without CSS for non-visual UAs, and alternate styling should be possible on top of the same base markup. I don't think features designed for apps should be given a free pass on this point.

index.bs
```diff
@@ -349,6 +349,220 @@ One example of a feature that should be limited to secure contexts
...
+then it probably belongs in the markup rather than in the CSS.
+
+<p class="example">For example, information about the human language (English, Chinese, Arabic, etc.)
+that the text is in belongs in the markup
+(in particular, the <code><a href="https://html.spec.whatwg.org/multipage/dom.html#the-lang-and-xml:lang-attributes">lang</a></code> attribute),
+even though a number of the uses of that information (such as font selection or hyphenation)
+are related to the presentation.
+On the other hand, information about the font size makes sense to be in the style,
+since in pages that use HTML correctly
+the information in font sizes should also be obtainable from the markup
```
astearns, line 377: 'the information conveyed by font sizes'

index.bs
```diff
@@ -349,6 +349,220 @@ One example of a feature that should be limited to secure contexts
 is geolocation, since the authentication and confidentiality provided by
 secure contexts reduce the risks to user privacy.
 
+<h2 id="css">Cascading Style Sheets (CSS)</h2>
+
+<h3 id="css-versus-markup">Presentation should be in CSS, content should be in HTML</h3>
+
+The separation between CSS and HTML was originally designed around the idea
```
astearns, line 356: Perhaps do 'CSS (style) and HTML (markup)' here, since you use the terms interchangeably after this? Or be consistent about using CSS and HTML instead of style and markup.

index.bs
astearns: If that's added I think it should go at the end of the section, as I'm not sure there's a way of stating that rule that provides as much explanation as what's currently here.

index.bs
```diff
@@ -349,6 +349,220 @@ One example of a feature that should be limited to secure contexts
...
+often involves tradeoffs of complexity.
+Sometimes it is less complex
+for the behavior to happen as part of the computation of the computed value;
+sometimes it is less complex 
+for the behavior to happen in whatever handles the computed value.
+(This complexity might be in the specification, in implementations, or both.
+Complexity in implementations might affect how fast things run for end users.)
+
+<div class="example">
+For example, the 'line-height' property accepts values that are integers,
```
astearns, line 522: s/integers/numbers/

index.bs
```diff
@@ -349,6 +349,220 @@ One example of a feature that should be limited to secure contexts
...
+  &lt;h2 style="font-size: 200%"&gt;
+    And this heading also has a line-height that's
+    1.4 times the h2's larger font size, and *not*
+    1.4 times the body's font size, which would lead
+    to text overlap
+  &lt;/h2&gt;
+&lt;/body&gt;
+  </pre>
+
+These integer values are generally the preferred values to use for 'line-height'
```
astearns, line 547: s/integer/number/

index.bs
dbaron: I rewrote this paragraph a different way from what you suggested, but incorporating a bunch of your suggestions.

index.bs
dbaron: I've added a paragraph on this as the third paragraph of the section.

index.bs
dbaron: Added.

index.bs
dbaron: Added.

index.bs
dbaron: I went with "shown by".

index.bs
dbaron: fixed

dbaron:
I've addressed *most* of the feedback but not all (and marked the addressed stuff as resolved, so you can see what's left).  I also updated [the online preview](https://dbaron.github.io/design-principles/#css) and it actually worked this time because github was not broken.

index.bs
dbaron: As discussed in today's teleconference, I'm going to drop this section and merge the rest.  Landing a replacement for this text is covered in #169.

index.bs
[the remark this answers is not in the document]
dbaron: As discussed in today's teleconference, I'm going to drop this section and merge the rest.  Landing a replacement for this text is covered in #169.

index.bs
dbaron: I've moved this to #170.

dbaron:
As discussed in today's teleconference, I've broken out #169 and #170 for followup, and I'm going to merge this with the first subsection removed.

dbaron:
And definitely feel free to file followup issues if you find further issues with what has landed here.

PR #180, opened 2020-05-06
Add principles related to device enumeration & tweak the existing device identifier text.

### LINKED ISSUE #69: Update section on device IDs based on thinking recorded in design reviews issue
See issue https://github.com/w3ctag/design-reviews/issues/64

### LINKED ISSUE #152: Discourage device enumeration, prefer less powerful alternatives
Consider an API which exposes a certain kind of attached device to websites. For instance, an API which can vend one or more monitor objects that represents one or more of the monitors attached to the computer.

## Common approaches

1. a **device enumeration API**: exposes / returns a **list of all devices**
2. a **sorted device enumeration API**: exposes / returns an **ordered list of all devices**, ordered by some set of site-provided preferences
3. a **filtered device enumeration API**: exposes / returns a **filtered list of devices** which meet some set of constraints
4. a **picker-style API**: exposes / returns **one device** which meets some set of constraints
5. an **availability-style API**: exposes **that at least one device is available to use** and a way to **use an available device** without exposing any devices directly

## Examples

1. a **device enumeration API**: ["give me a list of all the screens attached to this computer"](https://github.com/webscreens/screen-enumeration/blob/master/EXPLAINER.md) (see also w3ctag/design-reviews#413, https://github.com/w3ctag/design-reviews/issues/413#issuecomment-543580676, and webscreens/screen-enumeration#23), ["give me a list of all the USB devices"](https://wicg.github.io/webusb/#dom-usb-getdevices), ["give me a list of all the cameras and microphones"](https://w3c.github.io/mediacapture-main/#dom-mediadevices-enumeratedevices) (see also w3c/mediacapture-main#612, w3c/mediacapture-main#640, and w3c/mediacapture-main#646)
2. a **sorted device enumeration API** ["give me a sorted list of all the GPUs on the system"](https://github.com/gpuweb/gpuweb/pull/524)
3. a **filtered device enumeration API**: offhand I can't think of one, but I'm sure there's one out there.
4. a **picker-style API**: ["give me a high-performance  or low-power GPU"](https://gpuweb.github.io/gpuweb/#requestadapter) ["give me a USB device that meets some criteria"](https://wicg.github.io/webusb/#dom-usb-requestdevice), `<input type=file>`
5. an **availability-style API**: ["switch to remote playback if the user wants"](https://www.w3.org/TR/remote-playback/)

## Observations

Only some of these approaches center the user's device preferences.

Merely exposing the presence of devices is itself a fingerprinting concern.

Exposing device data grows the fingerprinting surface per exposed device, which can get very bad very quickly if the API exposes a large amount of per-device information.

[The Rule of Least Power](https://www.w3.org/2001/tag/doc/leastPower.html) suggests that we should go with the least powerful API which meets its use cases.

## Conclusion

Therefore, we should recommend that, when API designers need to expose devices in some manner, that they consider these in reverse order.

An **availability-style API** is best: it allows the site to use the device the user wants to use without exposing device information. It is the least powerful option.

A **picker-style API** also respects user preferences. It exposes device information, but only of one device. This minimizes the fingerprinting data exposed.

A **filtered device enumeration API** exposes a subset of devices, so is potentially better re: fingerprinting than the next possibilities

A **sorted device enumeration API** exposes too much information, but at least it does so in a way that makes the site's job easier

A **device enumeration API** is to be resorted to only when the other options are infeasible for some reason (compat, perhaps).

hober:
Fixes #69, #152, and #181.

The diff is a bit bigger than strictly necessary because I added clause line breaks to the `[[#device-ids]]` section.

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/180.html" title="Last updated on May 28, 2020, 5:21 PM UTC (2ec5b8e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/180/12e38e5...2ec5b8e.html" title="Last updated on May 28, 2020, 5:21 PM UTC (2ec5b8e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1371,16 +1371,19 @@ This section contains principles for consideration when designing APIs for devic
 
 <h3 id="device-ids">Use care when exposing identifying information about devices</h3>
 
-Exposing device identifiers
-(or information about a device
-or its capabilities
-that may be used to synthesize such an identifier)
-increases the fingerprinting surface of a user agent,
-thus harming user privacy.
 Merely exposing the presence of a device
-is itself a fingerprinting concern.
-Please read the TAG's finding
-on [unsanctioned tracking](http://www.w3.org/2001/tag/doc/unsanctioned-tracking/)
+may harm user privacy.
+Exposing
+additional information about a device
+or its capabilities
+increases this risk further.
+Exposing device identifiers
+increases this risk even more.
+Any such information
+increases the set of fingerprinting data available to sites.
+But there are possible user privacy harms here
+that go beyond fingerprinting.
+Please read [[FINGERPRINTING-GUIDANCE]] and [[UNSANCTIONED-TRACKING]]
 for additional details.
 
 Think carefully about
--- a/index.bs
+++ b/index.bs
@@ -1699,7 +1699,6 @@ so that readers can quickly decide whether they need to read the steps in detail
      https://github.com/w3c/hr-time/issues/90
      https://github.com/whatwg/html/issues/5515
      https://github.com/w3c/remote-playback/issues/137
-     https://github.com/w3c/mediacapture-main/issues/695
      -->
 <pre class="anchors">
 urlPrefix: https://w3c.github.io/hr-time/; spec: HIGHRES-TIME
@@ -1709,6 +1708,4 @@ urlPrefix: https://html.spec.whatwg.org/multipage/; spec: HTML
 urlPrefix: https://w3c.github.io/remote-playback/; spec: REMOTE-PLAYBACK
     text: RemotePlayback; type:interface; url: #remoteplayback-interface
     text: remote playback device; type:dfn; url: #dfn-remote-playback-devices
-urlPrefix: https://w3c.github.io/mediacapture-main/getusermedia.html; spec: MEDIACAPTURE-STREAMS
-    text: constrainable object; type:dfn; url: #constrainable-interface
 </pre>
--- a/index.bs
+++ b/index.bs
@@ -1433,17 +1433,23 @@ that this is done in a consistent and privacy-friendly way:
     despite the physical device being the same.
 :: Within an origin,
     ids may have a stable representation to web content.
-    This ensures a consistent developer experience if requesting the same device twice.
+    This ensures a consistent developer experience
+    if requesting the same device twice.
 : Persistable when necessary
-:: Device identifiers
-    obtained after a complex or time-consuming device selection process
-    may prefer that author code
-    be allowed to persist the id
+:: Some device identifiers
+    can only be obtained
+    via a complex or time-consuming device selection process.
+    In such cases,
+    it may make sense for author code
+    to be able to persist such identifiers
     for use in a later session
     in order to avoid the selection process a second time.
-    In this case,
-    the API should not only provide a stable id during the session for the given origin,
-    but also be able to deterministically produce the same id in subsequent sessions.
+    When this is the case,
+    the API should not only
+    provide a stable id during the session for the given origin,
+    but it should also be capable of
+    deterministically producing the same id
+    in subsequent sessions.
 
 <h3 id="device-enumeration">Use care when exposing APIs for selecting or enumerating devices</h3> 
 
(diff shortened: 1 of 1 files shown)


jyasskin:
2¢: For future reviews, it'd help to isolate the introduction of clause line breaks into one commit, and then edit from there.

index.bs
```diff
@@ -1360,56 +1360,176 @@ devices. For example, authors wish to be able to use the web to connect with the
...
+    such as branding, make and model numbers, etc.
+    In many cases
+    using a randomized number or unique id
+    is preferable to
+    a generic string identifier such as "device1".
+
+    Device ids expressed as numbers
+    should contain sufficient entropy
+    so as to avoid re-use or potential sharing among other devices,
+    and should not be easily guessable.
```
jyasskin, line 1412: This question has come up in some PING reviews: @pes10k has argued that device IDs should contain minimal entropy so that they can't be used as user identifiers even within a single [storage partition](https://github.com/privacycg/storage-partitioning). I think it's fine for them to contain a lot of entropy as long as they are partitioned the same way as storage. If y'all are taking a position on that question, you should do it intentionally. :)

index.bs
```diff
@@ -1360,56 +1360,176 @@ devices. For example, authors wish to be able to use the web to connect with the
...
+by a less powerful API. [[LEAST-POWER]]
+
+Despite these concerns,
+you may find that it is necessary
+to expose identifying information about a device to the web platform.
+
+The following guidelines will help ensure
+that this is done in a consistent and privacy-friendly way:
+
+: Limit identifiable information in the id
```
jyasskin, line 1399: I looks like the gist of this section is
1. Don't expose identifying information about a device "such as branding, make and model numbers, etc."
2. Here's how to mint an id number/string for each device-within-a-[storage-partition](https://github.com/privacycg/storage-partitioning) safely.

I think you could say that more clearly.

index.bs
```diff
@@ -1360,56 +1360,176 @@ devices. For example, authors wish to be able to use the web to connect with the
...
+
+As in the previous section,
+please consider if your use cases could be satisfied
+by a less powerful API. [[LEAST-POWER]]
+
+If the purpose of the API
+is to enable the user **to select a device**
+from the set of available devices of a particular kind,
+you may not need to expose a list to script at all.
+An API which invokes a User Agent-provided device picker
```
jyasskin, line 1469:
```suggestion
An API which invokes a User-Agent-provided device picker
```

index.bs
```diff
@@ -1360,56 +1360,176 @@ devices. For example, authors wish to be able to use the web to connect with the
...
-     the same id in subsequent sessions.
+This section contains principles for consideration when designing APIs for devices.
+
+<h3 id="device-ids">Use care when exposing identifying information about devices</h3>
+
+Exposing device identifiers
+(or information about a device
+or its capabilities
+that may be used to synthesize such an identifier)
+increases the fingerprinting surface of a user agent,
```
jyasskin, line 1378: This seems to imply that only information that uniquely identifies one device or device model increases the fingerprinting surface, but of course even revealing, for example, that a camera has a particular resolution will reduce the user's anonymity set.

An identifier for a particular model of device also has privacy harms beyond fingerprinting: knowing that someone has a particular speaker could help speaker-sellers target their marketing more precisely.

In the long run, we should go into more detail about this in https://w3cping.github.io/privacy-threat-model/#hl-sensitive-information...

index.bs
hober: This PR doesn't take a position on this question; the text in question isn't changed (besides the line breaking). I think it'd be useful to file a new issue to capture this concern.

index.bs
hober: I agree. This PR didn't change this text (besides the line breaking). Maybe file a new issue for this?

hober:
This is now blocked on https://github.com/tobie/specref/pull/601.

index.bs
hober: > This seems to imply that only information that uniquely identifies one device or device model increases the fingerprinting surface, but of course even revealing, for example, that a camera has a particular resolution will reduce the user's anonymity set.
> 
> An identifier for a particular model of device also has privacy harms beyond fingerprinting: knowing that someone has a particular speaker could help speaker-sellers target their marketing more precisely.

This is great feedback. I've tried to capture it in 2ec5b8e.

> In the long run, we should go into more detail about this in https://w3cping.github.io/privacy-threat-model/#hl-sensitive-information...

Please let us know when you think that's ready for citing!

Issue #95, opened 2018-04-07
Guide spec authors to use one manifest (web app manifest)


travisleithead:
We should provide guidelines on what to do if you want to create some data that should be in a manifest. We would like to avoid a proliferation of manifest files.

E.g., Assume you should integrate into web-app-manifest. If you can't for [good reason here] then [do this alternat thing].

See also https://github.com/w3ctag/design-reviews/issues/231#issuecomment-379432055 which motivated the opening of this issue (among other previous discussions).

hober:
See also https://github.com/w3ctag/design-reviews/issues/344

BigBlueHat:
The Web App Manifest doesn't actually (on it's own) "manifest" anything. It points to icons and metadata to be used in launcher icon creation and application discovery. It then points to a ServiceWorker for actually "manifestation" of the Web App and to a "scope" for limiting navigation within the constrained navigation environment (i.e. "standalone" browser environment).

Other manifest formats (like Audiobooks and Publication Manifest) encode a (deliberately) static, non-procedural list of resources--as one would do in a physical manifest document (i.e. the list of things to include + metadata about that list).

The Publication WG did look into Web App Manifest as a foundation, but it's dependency on ServiceWorkers (and therefore JavaScript) to do the actually manifestation brought in a new significant requirement that actually made a publication (or audiobook's) construction less dependable and more fragile and subjective (vs. a consistently structured, static list of resources).

Hopefully that clears up some of the confusion caused by the varied use of the word "manifest" among current W3C specifications.

dbaron:
One other question about web app manifests that might be relevant here:  is it OK for web app manifest to be the only way to get to a feature, or should everything in web app manifest be an alternative way of getting to a feature that's available some other way (but that might, say, make it available before the user has interacted with the app)?  (I think I've seen that distinction discussed elsewhere at some point recently.)

torgo:
Will add to breakout C this week.

kenchris:
- point to 2.2 for naming (underscore)
- if the fetching of the metadata/manifest is at a different time at than existing manifest, a new manifest can be considered (example payment handler manifest) but if at all avoidable, please extend an existing manifest
- if a new key is added, make sure it is really needed (define well thought out use-cases) and check if a similar key exists. If an existing key does more or less what is needed, work with the existing spec to extend it to your use-case if possible.

torgo:
@tantek's text from issue #148 
[...]

Issue #148, opened 2019-09-19
Proliferation of manifests at W3C


tantek:
Here at #w3cTPAC 2019 I’m learning that there are quite a few different manifest file format efforts by different groups for different specifications with different but often similar use-cases. Filing this design issue per @hober’s request to track the growing number of manifest formats and approaches to hopefully collapsing and reducing them.

Here is a few I know of along with their individual TAG review issues.

  * [Web App Manifest](https://www.w3.org/TR/appmanifest/) ([W3CTAG w3ctag/design-reviews#30](https://github.com/w3ctag/design-reviews/issues/30))
  * [Payment Method Manifest](https://w3c.github.io/payment-method-manifest/) ([W3CTAG w3ctag/design-reviews#162](https://github.com/w3ctag/design-reviews/issues/162), [Payment Method w3ctag/design-reviews#27](https://github.com/w3c/payment-method-manifest/issues/27), [W3CTAG w3ctag/design-reviews#231 Payment Handler](https://github.com/w3ctag/design-reviews/issues/231))
  * [Audio Book Manifest](https://w3c.github.io/wpub/#audiobook) ([W3CTAG w3ctag/design-reviews#345](https://github.com/w3ctag/design-reviews/issues/345))
  * [Publication Manifest](https://w3c.github.io/wpub/#webidl) ([W3CTAG w3ctag/design-reviews#344](https://github.com/w3ctag/design-reviews/issues/344))

And one apparent need for a manifest: [W3CTAG w3ctag/design-reviews#218](https://github.com/w3ctag/design-reviews/issues/218). There are likely others, please add more below.

From a web author, developer, publisher perspective, it would be great if we could simplify the number of manifests, ideally one, perhaps with core properties, optional properties for specific use-cases, and a method of expanding each of those by separate working groups for maximum re-use and avoiding collisions / duplication of slightly different properties.

This issue is intended as a meta-issue just to track the growing number of manifests.

For specific proposals to provide guidelines to both spec and web authors about how to create some data and provide it in a manifest file, see: [W3CTAG w3ctag/design-reviews#95](https://github.com/w3ctag/design-principles/issues/95) (please do not mark as duplicate, I think these serve two different purposes). Thanks for your consideration!

(Originally published at: https://tantek.com/2019/261/b1/proliferation-manifests-w3c)

torgo:
We've suggested adding a section 6.3 to the document "Manifests" with some text to the effect that specs should make use of existing manifests:

[...]

Issue #128, opened 2019-05-21
guidance on use of JSON-LD?


dbaron:
I think we should consider giving guidance on the use of JSON-LD.  In particular, it seems to me that JSON-LD documents can be interpreted in two ways, either (1) as a JSON document or (2) based on the data model defined in the JSON-LD spec.  It's not clear to me if there are clearly-defined terms for these two things, though.

Specifications that use JSON-LD ought to be clear when users of the JSON-LD involved must interpret it in one of the two ways, or when they're free to use either.  (I could imagine the answer might vary between specifications, or might vary between conformance classes in a specification.)

I'd note that I prefer that browsers always do (1).

hadleybeeman:
@dbaron, we've discussed this on today's TAG breakout. While it doesn't look like many (if any?) browsers currently parse JSON-LD with its data model - I worry about explicitly say that they won't or shouldn't. @torgo wondered if we could pull together some language about how a spec authory may not be able to expect the browser to parse it... But we then brainstormed a number of different things that a spec author may not expect the browser to do, and were struggling to work out why parsing JSON-LD with its data model might be worth calling out. What are your thoughts here? What would be the thing you'd hope we could solve by adding something about this to the Design Principles doc?

marcoscaceres:
TAG should work with the editors of JSON-LD to define the actual data model of JSON-LD in terms of WHATWG Infra. I think the current spec is potentially harmful because it tries to naively piggyback on WebIDL without consideration that WebIDL is very browser specific.  

This is potentially harmful to JSON-LD (and any organization adopting it) in the long term, because JSON-LD wants to be language/implementation agnostic, which WebIDL is not: WebIDL is very much bound to how browsers work. There is no WebIDL implementation in Ruby, or even completely in JavaScript, for instance - that should be of concern. 

An alternative might be to suggest that JSON-LD be expressed in terms of WHATWG Infra types + use Infra's algorithms for converting JSON into infra types. That would give a good language/implementation agnostic data model, plus allow JSON-LD to be treated truly as JSON. The browsers don't need to care because they support JSON (addressing @dbaron's concern (1)) and it addresses (2), in that Infra types map 1-to-1 to JavaScript types - as well as C++ types, and Rust, and basically any other language that supports primitive abstract types (strings, lists, numbers, booleans, bytes, etc.).

torgo:
[...]

PR #189, opened 2020-05-26
added section on manifests

torgo:
new section to address https://github.com/w3ctag/design-principles/issues/148

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/189.html" title="Last updated on May 27, 2020, 3:25 PM UTC (a005f40)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/189/c5ee760...a005f40.html" title="Last updated on May 27, 2020, 3:25 PM UTC (a005f40)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1726,8 +1726,8 @@ which creates developer confusion. The use of different manifests on the web sho
 
 Metadata can reside inside a new manifest if that is related to a different function or the fetch timing 
 of that metadata is different. Application metadata should be added to the Web App Manifest or be an extension of it. 
-Parameters for a specific function (like Payment Method Manifest) are a completely different type. Publication Manifest,
-using JSON-LD is another type as well.
+Parameters for a specific function may require different approaches.
+Payment Method Manifest and Publication Manifest are examples of these cases.
 
 Note that in all cases, the naming conventions should be harmonized (see section 2.2)
  
--- a/index.bs
+++ b/index.bs
@@ -1728,7 +1728,7 @@ files are required but the use of different manifests on the web should be kept 
 You may need to make a new manifest file if the domain of the manifest file is different from the existing manifest files,
 if the fetch timing is different, or if the complexity of the manifest warrents it.  Application metadata should be added 
 to the Web App Manifest or be an extension of it. 
-Parameters for a specific function may require different approaches.
+Manifests designated to be used for specific applications or which require interoperability with non-browsers may need to take a different approach.
 Payment Method Manifest, Publication Manifest, and Origin Policy are examples of these cases.
 
 For example, if you have a single piece of metadata, even if the fetech timing is different than an exisiting manfiest, it is probabluy best to use an existing manifest (or ideally design the feature in such a way that a manifest is not required). However, if your feature requires a complex set of metadata that is specific to a functional domain, then creation of a new manifest may be warented.
--- a/index.bs
+++ b/index.bs
@@ -1721,9 +1721,9 @@ that should be considered in the development of new features, notably:
 
 <h3>Extend existing manifest files rather than creating new ones</h3>
 
-New web features should be self-contained and self-descriving and ideally should not require an additional manifest file
-which can add complexity to the development and deployment process for web technologies.  However, sometimes manifest 
-files are required. However, they can also add complexity to the web platform, which creates developer confusion.
+New web features should be self-contained and self-describing and ideally should not require an additional manifest file.
+Manifest files can add complexity to the development and deployment process for web technologies.  However, sometimes manifest 
+files are required.
 The use of different manifests on the web should therefore be kept to a minimum.
 
 Metadata can reside inside a new manifest if that is related to a different function or the fetch timing 
@@ -1747,7 +1747,7 @@ Some of the existing manifest files include
 * <a href="https://www.w3.org/TR/appmanifest/">Web App Manifest</a> which contains features related to web applications.
 * <a href="https://www.w3.org/TR/payment-method-manifest/">Payment Method Manifest</a> which is used for payment methods in the context of the web payment API
 * <a href="https://www.w3.org/TR/pub-manifest/">Publication Manifest</a> which is used by some web publications working group standards
-* <a href="https://wicg.github.io/origin-policy/">Origin Policy</a> which is used t set security policies
+* <a href="https://wicg.github.io/origin-policy/">Origin Policy</a> which is used to set security policies.
 
 We encourage people to think of manifest files as extensible. Always try to get the changes into the original spec, 
 or at least have a discussion around the extension with the spec editors, as that will usually lead to a better design
--- a/index.bs
+++ b/index.bs
@@ -1731,12 +1731,12 @@ using JSON-LD is another type as well.
 
 Note that in all cases, the naming conventions should be harmonized (see section 2.2)
  
-[Note: Due to historical reasons, existing manifests use lowercase, underscore-delimited names. There have been times 
+Note: For historical reasons, existing manifests use lowercase, underscore-delimited names. There have been times 
 where it has been beneficial to re-use dictionaries from a manifest in DOM APIs as well, which are camel-cased. 
 One such example is the <a href="https://w3c.github.io/image-resource/">image resource</a>. For this reason, if a 
-key can clearly be expressed as a single word, that is recommended.]
+key can clearly be expressed as a single word, that is recommended.
  
-When designing new keys and values for a manifest, make sure they are really needed (thus, define well thought out use-cases).
+When designing new keys and values for a manifest, make sure they are really needed (that is, they enable well-thought-out use-cases).
 Also please check if a similar key exists. If an existing key/value pair does more or less what is needed, 
 work with the existing spec to extend it to your use-case if possible.
 
@@ -1749,11 +1749,13 @@ Some of the existing manifest files include
 We encourage people to think of manifest files as extensible. Always try to get the changes into the original spec, 
 or at least have a discussion around the extension with the spec editors, as that will usually lead to a better design
 and to something that can be integrated into the spec. There are certain times the spec authors might not want to 
-integrate immediately due to process (like going to CR) or due to it having a different scope, like extension to Web App
+integrate immediately due to process (like going to CR) or due to it having a different scope, like extensions to Web App
 Manifest only affecting store or payment use-cases. In that case it is fine to monkey-patch as long as that is agreed 
 with the original spec editors.
 
-An example of something that was done as a monkey patch but scheduled to be integrated into the web app manifest in 
+ISSUE(184): when we write up a principle on monkey patching, be sure to take this nuance into account.
+
+An example of something that was done as a monkey patch that is scheduled to be integrated into the web app manifest in 
 a future level (post CR):
 
 * https://wicg.github.io/web-share-target/#extension-to-the-web-app-manifest 
--- a/index.bs
+++ b/index.bs
@@ -1727,7 +1727,7 @@ which creates developer confusion. The use of different manifests on the web sho
 Metadata can reside inside a new manifest if that is related to a different function or the fetch timing 
 of that metadata is different. Application metadata should be added to the Web App Manifest or be an extension of it. 
 Parameters for a specific function may require different approaches.
-Payment Method Manifest and Publication Manifest are examples of these cases.
+Payment Method Manifest, Publication Manifest, and Origin Policy are examples of these cases.
 
 Note that in all cases, the naming conventions should be harmonized (see section 2.2)
  
@@ -1745,6 +1745,7 @@ Some of the existing manifest files include
 * <a href="https://www.w3.org/TR/appmanifest/">Web App Manifest</a>
 * <a href="https://www.w3.org/TR/payment-method-manifest/">Payment Method Manifest</a>
 * <a href="https://www.w3.org/TR/pub-manifest/">Publication Manifest</a>
+* <a href="https://wicg.github.io/origin-policy/">Origin Policy</a>
 
 We encourage people to think of manifest files as extensible. Always try to get the changes into the original spec, 
 or at least have a discussion around the extension with the spec editors, as that will usually lead to a better design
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1719,6 +1719,45 @@ that should be considered in the development of new features, notably:
...
 * Polyfill development should be encouraged
 
+<h3>Extend existing manifest files rather than creating new ones</h3>
+
+Manifest files are sometimes required. However, they can also add complexity to the web platform,
+which creates developer confusion. The use of different manifests on the web should therefore be kept to a minimum.
+
+Metadata can reside inside a new manifest if that is related to a different function or the fetch timing 
+of that metadata is different. Application metadata should be added to the Web App Manifest or be an extension of it. 
+Parameters for a specific function (like Payment Handler Manifest) are a completely different type. Publication Manifest,
```
marcoscaceres, line 1729:
```suggestion
Parameters for a specific function (like Payment Method Manifest) are a completely different type. Publication Manifest,
```

index.bs
```diff
@@ -1719,6 +1719,45 @@ that should be considered in the development of new features, notably:
...
+Also please check if a similar key exists. If an existing key/value pair does more or less what is needed, 
+work with the existing spec to extend it to your use-case if possible.
+
+Some of the existing manifest files include
+
+* <a href="https://www.w3.org/TR/appmanifest/">Web App Manifest</a>
+* <a href="https://www.w3.org/TR/payment-method-manifest/">Payment Method Manifest</a>
+* <a href="https://www.w3.org/TR/pub-manifest/">Publication Manifest</a>
+
+We encourage people to think of manifest files as extensible. So attempt to get the changes into the original spec,
```
marcoscaceres, line 1749:
```suggestion
We encourage people to think of manifest files as extensible. Always try to get the changes into the original spec, 
```

index.bs
```diff
@@ -1719,6 +1719,45 @@ that should be considered in the development of new features, notably:
...
+or at least have a discussion around the extension with the spec editors, as that will usually lead to a better design
+and to something that can be integrated into the spec. There are certain times the spec authors might not want to 
+integrate immediately due to process (like going to CR) or due to it having a different scope, like extension to Web App
+Manifest only affecting store or payment use-cases. In that case it is fine to monkey-patch as long as that is agreed 
+with the original spec editors.
+
+An example of something that was done as a monkey patch but scheduled to be integrated into the web app manifest in 
+a future level (post CR):
+
+* https://wicg.github.io/web-share-target/#extension-to-the-web-app-manifest
```
marcoscaceres, line 1759: Kind risky linking to this because it will break once we move it over... What would be idea is, "new property is defined in existing manifest, but processing of the new property can be defined in another spec".

index.bs
```diff
@@ -1719,6 +1719,45 @@ that should be considered in the development of new features, notably:
...
 
+<h3>Extend existing manifest files rather than creating new ones</h3>
+
+Manifest files are sometimes required. However, they can also add complexity to the web platform,
+which creates developer confusion. The use of different manifests on the web should therefore be kept to a minimum.
+
+Metadata can reside inside a new manifest if that is related to a different function or the fetch timing 
+of that metadata is different. Application metadata should be added to the Web App Manifest or be an extension of it. 
+Parameters for a specific function (like Payment Handler Manifest) are a completely different type. Publication Manifest,
+using JSON-LD is another type as well.
```
annevk, line 1730: I don't think we should mention JSON-LD, see https://github.com/w3ctag/design-principles/issues/128.

index.bs
```diff
@@ -1719,6 +1719,45 @@ that should be considered in the development of new features, notably:
...
+
+Note that in all cases, the naming conventions should be harmonised (see section 2.2)
```
annevk, line 1732:
```suggestion
Note that in all cases, the naming conventions should be harmonized (see section 2.2)
```

index.bs
cynthia: Since process wise it has already moved out of incubation, could you give us the new URL? (if it exists, that is.)

Otherwise you could just send a PR whenever the new repo is ready.

marcoscaceres: Still in incubation... but yes, can try to remember to send a PR.

torgo:
Also intended to close https://github.com/w3ctag/design-principles/issues/95

index.bs
cynthia: Okay, then remind us later.

index.bs
cynthia: This doesn't really seem to suggest JSON-LD is the way to go; it just mentions that as one of many examples in a neutral way. (Not that I'm a huge fan of JSON-LD, just to be clear)

I'm more concerned about both of the examples listed after the application manifest recommendation - it lists two examples with no analysis or guidance for the reader to work with - so I'm not sure listing those two examples really help in any way at all.

annevk: I'm opposed to mentioning JSON-LD in the design principles. I don't think including it is neutral at all.

index.bs
```diff
@@ -1719,6 +1719,45 @@ that should be considered in the development of new features, notably:
...
+of that metadata is different. Application metadata should be added to the Web App Manifest or be an extension of it. 
+Parameters for a specific function (like Payment Method Manifest) are a completely different type. Publication Manifest,
+using JSON-LD is another type as well.
+
+Note that in all cases, the naming conventions should be harmonized (see section 2.2)
+ 
+[Note: Due to historical reasons, existing manifests use lowercase, underscore-delimited names. There have been times 
+where it has been beneficial to re-use dictionaries from a manifest in DOM APIs as well, which are camel-cased. 
+One such example is the <a href="https://w3c.github.io/image-resource/">image resource</a>. For this reason, if a 
+key can clearly be expressed as a single word, that is recommended.]
```
domenic, line 1737: I don't think image resource is a good example of a manifest. It's IDL, and used for APIs only.

In general I think this parenthetical adds more confusion than it helps. For example "Due to historical reasons, existing manifests use lowercase, underscore-delimited names" doesn't seem right. It's due to following the design principles in section 2.2.

index.bs
```diff
@@ -1719,6 +1719,45 @@ that should be considered in the development of new features, notably:
...
+ 
+When designing new keys and values for a manifest, make sure they are really needed (thus, define well thought out use-cases).
+Also please check if a similar key exists. If an existing key/value pair does more or less what is needed, 
+work with the existing spec to extend it to your use-case if possible.
+
+Some of the existing manifest files include
+
+* <a href="https://www.w3.org/TR/appmanifest/">Web App Manifest</a>
+* <a href="https://www.w3.org/TR/payment-method-manifest/">Payment Method Manifest</a>
+* <a href="https://www.w3.org/TR/pub-manifest/">Publication Manifest</a>
```
domenic, line 1747: You have plenty of examples already, but I'll offer origin policy as another: https://wicg.github.io/origin-policy/. In particular that one exemplifies the different fetch timing and scope (see https://github.com/WICG/origin-policy/blob/master/policy-format.md#appendix-relation-to-web-app-manifest).

index.bs
```diff
@@ -1719,6 +1719,45 @@ that should be considered in the development of new features, notably:
...
 * Polyfill development should be encouraged
 
+<h3>Extend existing manifest files rather than creating new ones</h3>
+
+Manifest files are sometimes required. However, they can also add complexity to the web platform,
+which creates developer confusion. The use of different manifests on the web should therefore be kept to a minimum.
+
+Metadata can reside inside a new manifest if that is related to a different function or the fetch timing 
+of that metadata is different. Application metadata should be added to the Web App Manifest or be an extension of it. 
+Parameters for a specific function (like Payment Method Manifest) are a completely different type. Publication Manifest,
```
hober, line 1729: This sentence (starting "Parameters for a specific function" is extremely hard to understand.

index.bs
```diff
@@ -1719,6 +1719,45 @@ that should be considered in the development of new features, notably:
...
+using JSON-LD is another type as well.
+
+Note that in all cases, the naming conventions should be harmonized (see section 2.2)
+ 
+[Note: Due to historical reasons, existing manifests use lowercase, underscore-delimited names. There have been times
```
hober, line 1734: This should be "Note: ", not "[Note: ", so that you get Bikeshed's note styling.

hober: also s/due to/for/

index.bs
```diff
@@ -1719,6 +1719,45 @@ that should be considered in the development of new features, notably:
...
+using JSON-LD is another type as well.
+
+Note that in all cases, the naming conventions should be harmonized (see section 2.2)
+ 
+[Note: Due to historical reasons, existing manifests use lowercase, underscore-delimited names. There have been times 
+where it has been beneficial to re-use dictionaries from a manifest in DOM APIs as well, which are camel-cased. 
+One such example is the <a href="https://w3c.github.io/image-resource/">image resource</a>. For this reason, if a 
+key can clearly be expressed as a single word, that is recommended.]
+ 
+When designing new keys and values for a manifest, make sure they are really needed (thus, define well thought out use-cases).
```
hober, line 1739: s/thus, define/that is, they enable/

s/well thought out/well-thought-out/

index.bs
```diff
@@ -1719,6 +1719,45 @@ that should be considered in the development of new features, notably:
...
+Also please check if a similar key exists. If an existing key/value pair does more or less what is needed, 
+work with the existing spec to extend it to your use-case if possible.
+
+Some of the existing manifest files include
```
hober, line 1743: Are these good examples? Would we add these manifests now, in a world where this advice exists? It would be good to have examples that (retroactively) follow these guidelines and also examples that do not, but that we're stuck with for compat reasons.

index.bs
```diff
@@ -1719,6 +1719,45 @@ that should be considered in the development of new features, notably:
...
+
+* <a href="https://www.w3.org/TR/appmanifest/">Web App Manifest</a>
+* <a href="https://www.w3.org/TR/payment-method-manifest/">Payment Method Manifest</a>
+* <a href="https://www.w3.org/TR/pub-manifest/">Publication Manifest</a>
+
+We encourage people to think of manifest files as extensible. Always try to get the changes into the original spec, 
+or at least have a discussion around the extension with the spec editors, as that will usually lead to a better design
+and to something that can be integrated into the spec. There are certain times the spec authors might not want to 
+integrate immediately due to process (like going to CR) or due to it having a different scope, like extension to Web App
```
hober, line 1752: s/like extension to/like extensions to/

index.bs
```diff
@@ -1719,6 +1719,45 @@ that should be considered in the development of new features, notably:
...
+Manifest only affecting store or payment use-cases. In that case it is fine to monkey-patch as long as that is agreed 
+with the original spec editors.
```
hober, line 1754: Please add an issue marker here:

ISSUE(184): when we write up a principle on monkey patching, be sure to take this nuance into account.

index.bs
```diff
@@ -1719,6 +1719,45 @@ that should be considered in the development of new features, notably:
...
+
+An example of something that was done as a monkey patch but scheduled to be integrated into the web app manifest in
```
hober, line 1756: s/but scheduled/that is scheduled/

index.bs
kenchris: @annevk is it fine to leave the Publication Manifest and just not mention JSON-LD ?

index.bs
kenchris: I think this came from @dbaron but might be wrong. I am unsure about what else to write or how to rephrase it

index.bs
kenchris: True, but it used to be in manifest and it is still used as JSON in the Web App Manifest @marcoscaceres

index.bs
cynthia: Removed.

annevk: @kenchris it seems better to not include that either then.

index.bs
cynthia: Added this example.

index.bs
cynthia: This has been reworded from "due to historical reasons" to "by principle", let me know if that addresses your concerns.

domenic: I still think referencing image resource here is a bad idea.

index.bs
```diff
@@ -1719,6 +1719,56 @@ that should be considered in the development of new features, notably:
...
+Manifests designated to be used for specific applications or which require interoperability with non-browsers may need to take a different approach.
+Payment Method Manifest, Publication Manifest, and Origin Policy are examples of these cases.
+
+For example, if you have a single piece of metadata, even if the fetch timing is different than an existing manifest,
+it is probably best to use an existing manifest (or ideally design the feature in such a way that a manifest is not required).
+However, if your feature requires a complex set of metadata specific to a functional domain, the creation of a new manifest may be justified.
+
+Note that in all cases, the naming conventions should be harmonized (see section 2.2)
+ 
+Note: By principle, existing manifests use lowercase, underscore-delimited names.
```
kenchris, line 1740: So let's just make it this then @domenic ?

Note: By principle, existing manifest keys use lowercase, underscore-delimited names. If a 
key can clearly be expressed as a single word, that is recommended.

domenic: Seems reasonable, perhaps with a link to https://w3ctag.github.io/design-principles/#casing-rules

index.bs
```diff
@@ -1719,6 +1719,56 @@ that should be considered in the development of new features, notably:
...
+
+New web features should be self-contained and self-describing and ideally should not require an additional manifest file.
+Manifest files can add complexity to the development and deployment process for web technologies.  Sometimes manifest 
+files are required, but the use of different manifests on the web should be kept to a minimum.
+
+You may need to make a new manifest file if the domain of the manifest file is different from the existing manifest files.
+For example, if the fetch timing is different, or if the complexity of the manifest warrants it.  Application metadata should be added 
+to the Web App Manifest or be an extension of it. 
+Manifests designated to be used for specific applications or which require interoperability with non-browsers may need to take a different approach.
+Payment Method Manifest, Publication Manifest, and Origin Policy are examples of these cases.
```
annevk, line 1732: I don't think we should mention Publication Manifest, both here and below, as it uses JSON-LD, which we're at best undecided on.

index.bs
marcoscaceres: Agree with @domenic.

PR #194, opened 2020-05-28
Add section about optional parameters. Resolves #73.

### LINKED ISSUE #73: Optional arguments
Unless there are legacy reasons (e.g., XMLHttpRequest), omitting an argument and passing undefined must do the same thing.

This is more or less enforced by IDL, but if folks start using overloading (we should recommend against that too if we aren't already) it requires due diligence.

When it comes to optional arguments that can be dictionaries, passing `{}` must be no different from passing undefined or omitting the argument altogether.

cynthia:
This is a first draft written in 10 minutes during a breakout - quality writing is not something one should be expecting.

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/194.html" title="Last updated on May 28, 2020, 9:44 AM UTC (4b28ee7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/194/65aeebf...4b28ee7.html" title="Last updated on May 28, 2020, 9:44 AM UTC (4b28ee7)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1020,7 +1020,7 @@ is more readable than
 <h3 id="constructors">Make non-critical function parameters optional</h3>
 
 Parameters that are not strictly necessary should be made optional.
-Making arguments optional with reasonable defaults reduce overhead when the function is used,
+Making arguments optional with reasonable defaults reduces overhead when the function is used,
 and increases user-friendliness.
 
 Optional parameters should treat empty or undefined as equal to the omission of the parameter.
--- a/index.bs
+++ b/index.bs
@@ -1029,7 +1029,7 @@ foo(undefined) and foo() should be equal,
 along with foo({}) and foo() if the optional parameter accepts a dictionary.
 
 Exceptions have been made for legacy interoperability reasons (such as XMLHttpRequest),
-but this should be considered an exception rather than recommended practice.
+but this should be considered a design mistake rather than recommended practice.
 
 <h3 id="constructors">Classes should have constructors when possible</h3>
 
--- a/index.bs
+++ b/index.bs
@@ -1017,6 +1017,20 @@ For example,
 is more readable than
 <code highlight="js">window.scrollBy(50, 0)</code>.
 
+<h3 id="constructors">Make non-critical function parameters optional</h3>
+
+Parameters that are not strictly necessary should be made optional.
+Making arguments optional with reasonable defaults reduce overhead when the function is used,
+and increases user-friendliness.
+
+Optional parameters should treat empty or undefined as equal to the omission of the parameter.
+For example, given a function foo() which takes a single optional parameter,
+foo(undefined) and foo() should be equal,
+along with foo({}) and foo() if the optional parameter accepts a dictionary.
+
+Exceptions have been made for legacy interoperability reasons (such as XMLHttpRequest),
+but this should be considered an exception rather than recommended practice.
+
 <h3 id="constructors">Classes should have constructors when possible</h3>
 
 By default, [[!WEBIDL]] interfaces are reified in JavaScript as "non-constructible" classes: trying
--- a/index.bs
+++ b/index.bs
@@ -1020,15 +1020,25 @@ is more readable than
 <h3 id="constructors">Make non-critical function parameters optional</h3>
 
 Parameters that are not strictly necessary should be made optional.
-Making arguments optional with reasonable defaults reduces overhead when the function is used,
+Making arguments optional with reasonable defaults
+reduces overhead when the function is used,
 and increases user-friendliness.
 
-Optional parameters should treat empty or undefined as equal to the omission of the parameter.
-For example, given a function foo() which takes a single optional parameter,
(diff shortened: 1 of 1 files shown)


dbaron:
Is it worth suggesting that *all* of the optional arguments go in a dictionary?  I think that often seems to be a best practice.

index.bs
```diff
@@ -1017,6 +1017,20 @@ For example,
...
 
+<h3 id="constructors">Make non-critical function parameters optional</h3>
+
+Parameters that are not strictly necessary should be made optional.
+Making arguments optional with reasonable defaults reduce overhead when the function is used,
+and increases user-friendliness.
+
+Optional parameters should treat empty or undefined as equal to the omission of the parameter.
+For example, given a function foo() which takes a single optional parameter,
+foo(undefined) and foo() should be equal,
```
dbaron, line 1028: Maybe worth mentioning that WebIDL does this for you unless overloading is involved?  (At least I think that's what @annevk said in #73.)

cynthia: That's a good point, I'll make an edit to reflect this.

index.bs
```diff
@@ -1017,6 +1017,20 @@ For example,
 is more readable than
 <code highlight="js">window.scrollBy(50, 0)</code>.
 
+<h3 id="constructors">Make non-critical function parameters optional</h3>
+
+Parameters that are not strictly necessary should be made optional.
+Making arguments optional with reasonable defaults reduce overhead when the function is used,
```
annevk, line 1023:
```suggestion
Making arguments optional with reasonable defaults reduces overhead when the function is used,
```

index.bs
```diff
@@ -1017,6 +1017,20 @@ For example,
 is more readable than
 <code highlight="js">window.scrollBy(50, 0)</code>.
 
+<h3 id="constructors">Make non-critical function parameters optional</h3>
```
annevk, line 1020: Should we call talk about method and arguments instead? That would match IDL and might be clearer.

index.bs
```diff
@@ -1017,6 +1017,20 @@ For example,
...
+Making arguments optional with reasonable defaults reduce overhead when the function is used,
+and increases user-friendliness.
+
+Optional parameters should treat empty or undefined as equal to the omission of the parameter.
+For example, given a function foo() which takes a single optional parameter,
+foo(undefined) and foo() should be equal,
+along with foo({}) and foo() if the optional parameter accepts a dictionary.
+
+Exceptions have been made for legacy interoperability reasons (such as XMLHttpRequest),
+but this should be considered an exception rather than recommended practice.
```
annevk, line 1032:
```suggestion
but this should be considered a design mistake rather than recommended practice.
```

index.bs
```diff
@@ -1017,6 +1017,20 @@ For example,
...
 <code highlight="js">window.scrollBy(50, 0)</code>.
 
+<h3 id="constructors">Make non-critical function parameters optional</h3>
+
+Parameters that are not strictly necessary should be made optional.
+Making arguments optional with reasonable defaults reduce overhead when the function is used,
+and increases user-friendliness.
+
+Optional parameters should treat empty or undefined as equal to the omission of the parameter.
+For example, given a function foo() which takes a single optional parameter,
```
alice, line 1027: `<code highlight="js">` around code examples.

Also, can we make it a specific, real-world example, e.g.

```html
<p class="example">For example, 
<code highlight="js">element.scrollIntoView(false, {});</code> 
is equivalent to 
<code highlight="js">element.scrollIntoView(false);</code>,
and
<code highlight="js">element.addEventListener("click", onClick, undefined);</code>
is equivalent to 
<code highlight="js">element.addEventListener("click", onClick)</code>.
</p>
```

index.bs
cynthia: The naming, while not the best - is for consistency with the previous section. Whole document level terminology aligning is probably better suited in another PR since we jump back and forth quite a bit. I didn't want two consecutive sections to use different terms.

annevk: Filed #195.

index.bs
cynthia: Addressed in 0f0ae

index.bs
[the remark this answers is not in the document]
cynthia: Addressed in 0f0ae

index.bs
```diff
@@ -1017,6 +1017,20 @@ For example,
...
+foo(undefined) and foo() should be equal,
+along with foo({}) and foo() if the optional parameter accepts a dictionary.
+
+Exceptions have been made for legacy interoperability reasons (such as XMLHttpRequest),
```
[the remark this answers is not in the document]
cynthia: Addressed in 0f0ae

index.bs
cynthia: Thanks!

index.bs
cynthia: Addressed in 0f0ae

PR #204, opened 2020-05-29
Reword the 'priority of constituencies' principle

alice:
Rename to "Put user needs first", and reword in simpler language.

***
<a href="https://pr-preview.s3.amazonaws.com/alice/design-principles/pull/204.html" title="Last updated on May 29, 2020, 12:36 AM UTC (11f85c9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/204/850b40c...alice:11f85c9.html" title="Last updated on May 29, 2020, 12:36 AM UTC (11f85c9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -49,16 +49,25 @@ url: https://drafts.csswg.org/css-transitions-1/#transitions; type: dfn; text: C
 
 <h2 id="basic-principles">Principles behind design of Web APIs</h2>
 
-<h3 id="priority-of-constituencies">Priority of constituencies</h3>
+<h3 id="priority-of-constituencies">Put user needs first (Priority of Constituencies)</h3>
 
-In case of conflict,
-consider users over authors over implementors over specifiers over theoretical purity.
-In other words costs or difficulties to the user
-should be given more weight than costs to authors;
-which in turn should be given more weight than costs to implementors;
-which should be given more weight than costs to authors of the specification itself,
-which should be given more weight than those proposing changes for theoretical reasons alone.
-Of course, it is preferred to make things better for multiple constituencies at once.
+If a trade-off needs to be made, 
+always put user needs above all.
+
+User needs come before the needs of web page authors,
+which come before than the needs of user agent implementors,
+which come before than the needs of spec writers,
+which come before theoretical purity.
+
+Similarly, when beginning to design an API, 
+be sure to understand and document the user need that the API aims to address.
+
+Like all principles, 
+this is not absolute.
+Ease of authoring impacts how content reaches users.
+User agent implementation priorities affect how features reach authors.
+Spec writers also have finite resources,
+and theoretical concerns exist for a reason.
 
 See also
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -49,16 +49,25 @@ url: https://drafts.csswg.org/css-transitions-1/#transitions; type: dfn; text: C
...
+which come before than the needs of spec writers,
+which come before theoretical purity.
+
+Similarly, when beginning to design an API, 
+be sure to understand and document the user need that the API aims to address.
+
+Like all principles, 
+this is not absolute.
+Ease of authoring impacts how content reaches users.
+User agent implementation priorities affect how features reach authors.
```
dbaron, line 68: I think it might be worth mentioning finite resources here (like in the next line) rather than priorities, since they're the reason we need to have priorities.

index.bs
```diff
@@ -49,16 +49,25 @@ url: https://drafts.csswg.org/css-transitions-1/#transitions; type: dfn; text: C
...
+which come before than the needs of user agent implementors,
+which come before than the needs of spec writers,
+which come before theoretical purity.
+
+Similarly, when beginning to design an API, 
+be sure to understand and document the user need that the API aims to address.
+
+Like all principles, 
+this is not absolute.
+Ease of authoring impacts how content reaches users.
```
dbaron, line 67: Maybe "impacts" -> "affects" to match the next line.  (I'm a fan of parallel construction, and not a fan of "impact".)

index.bs
```diff
@@ -49,16 +49,25 @@ url: https://drafts.csswg.org/css-transitions-1/#transitions; type: dfn; text: C
...
+be sure to understand and document the user need that the API aims to address.
+
+Like all principles,
+this is not absolute.
+Ease of authoring impacts how content reaches users.
+User agent implementation priorities affect how features reach authors.
+Spec writers also have finite resources,
+and theoretical concerns exist for a reason.
 
 See also
```
dbaron, line 72: Given that you're editing this section, maybe add a ":" to the end of this line?

cynthia:
LGTM.

index.bs
```diff
@@ -49,16 +49,25 @@ url: https://drafts.csswg.org/css-transitions-1/#transitions; type: dfn; text: C
...
-In other words costs or difficulties to the user
-should be given more weight than costs to authors;
-which in turn should be given more weight than costs to implementors;
-which should be given more weight than costs to authors of the specification itself,
-which should be given more weight than those proposing changes for theoretical reasons alone.
-Of course, it is preferred to make things better for multiple constituencies at once.
+<h3 id="priority-of-constituencies">Put user needs first (Priority of Constituencies)</h3>
+
+If a trade-off needs to be made,
+always put user needs above all.
```
dbaron, line 55: I think it might be worth trying to explain *why* user needs are most important.  From the initial comment, part of it is that they're the largest number, but I think that's not all of it, and we should try to explain why if we can figure out how to.

index.bs

index.bs

index.bs

index.bs
alice: Made an attempt at this.

PR #208, opened 2020-05-29
Die on warnings when Travis tries to deploy.

hober:
Should have been part of #200.


hober:
Please take a look, @plinss, since you run the server this points at and I don't know if I got the syntax right.

PR #211, opened 2020-06-04
Make curl return error status when bikeshed server gives HTTP error status

dbaron:


dbaron:
At some point we should probably test that this causes the entire pipeline to fail correctly, but I have verified that this at least causes the `curl` command to return an failure exit status in the right cases.

Issue #145, opened 2019-12-04
clarify advice on synchronous versus asynchronous APIs


dbaron:
The discussion in w3ctag/design-reviews#439 (and really the discussion at TPAC in Fukuoka that led to filing it) point to the issue that the [advice in the design-principles document on synchronous versus asynchronous APIs](https://w3ctag.github.io/design-principles/#promises) is perhaps too strong, or at least can be interpreted in too strong a way.  It points out the reasons to prefer asynchronous APIs, but doesn't address possible tradeoffs against them.

(Two examples from that particular issue are (a) tradeoffs against user convenience and (b) tradeoffs against racy behavior that results from an API answering a question whose result might vary depending on whether the document is currently handling a user gesture.)

There's more detail after following the link above (and further links from that).

It should probably clearer how strong the advice to prefer asynchronous APIs is (which may differ across the various reasons) and what factors should be considered in the other direction.

alice:
Earlier: https://github.com/w3ctag/design-principles/issues/81

hober:
@cynthia to draft a PR.

torgo:
We discussed today and came up with the idea of a subsection providing guidance on "when to use synchronous".

cynthia:
I've removed the "should not fail" and replaced it to "not expected to be gated behind a permission".

cynthia:
Discussed in today's call; we've noticed that the principles document tends to advocate async, and we should probably tweak the tone throughout the document so it doesn't feel like we are advocating async.

alice:
Adding a note here to update the "Design asynchronous APIs using Promises" section of design principles as part of this work.

torgo:
Discussed on 05-01 and agreed to close as this is done.

PR #218, opened 2020-07-07
When to use synchronous - #145.

cynthia:
Needs more polish.

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/218.html" title="Last updated on Nov 3, 2020, 8:47 AM UTC (44502b7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/218/d0ad23a...44502b7.html" title="Last updated on Nov 3, 2020, 8:47 AM UTC (44502b7)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1227,7 +1227,7 @@ and need less infrastructure set-up (such as making functions `async`).
 An API should generally be synchronous if the following rules of thumb apply:
 
   * The API is not expected to ever be gated behind a permission prompt, or another dialog such as a device selector.
-  * The API implementation will not be blocked by filesystem or network access.
+  * The API implementation will not be blocked by filesystem or network access, for example, inter-process communication.
   * The execution time is short and deterministic.
 
 
--- a/index.bs
+++ b/index.bs
@@ -1220,7 +1220,7 @@ discuss.</p>
 
 <h3 id="synchronous">Use synchronous when appropriate</h3>
 
-Synchronous APIs should be considered when creating a new API.
+Where possible, prefer synchronous APIs when designing a new API.
 Synchronous APIs are simpler to use,
 and need less infrastructure set-up (such as making functions `async`).
 
--- a/index.bs
+++ b/index.bs
@@ -1227,7 +1227,7 @@ and need less infrastructure set-up (such as making functions `async`).
 An API should generally be synchronous if the following rules of thumb apply:
 
   * The API is not expected to ever be gated behind a permission prompt, or another dialog such as a device selector.
-  * The API implementation will not be blocked by filesystem or network access, for example, inter-process communication.
+  * The API implementation will not be blocked by a lock, filesystem or network access, for example, inter-process communication.
   * The execution time is short and deterministic.
 
 
--- a/index.bs
+++ b/index.bs
@@ -1223,7 +1223,6 @@ discuss.</p>
 Synchronous APIs should be considered when minting a new API.
 Synchronous APIs are simpler to use,
 and need less infrastructure set-up (such as making functions `async`).
-Moreover, it reduces the amount of infrastructural setup overhead needed for the function call.
 
 The general rule of thumb for an API that should be synchronous is the following:
 
--- a/index.bs
+++ b/index.bs
@@ -1226,7 +1226,7 @@ Moreover, it reduces the amount of infrastructural setup overhead needed for the
 
 The general rule of thumb for an API that should be synchronous is the following:
 
-  * The API at not point is expected to be gated behind a permission or a dialog, such as a device selector.
+  * The API at no point is expected to be gated behind a permission or a dialog, such as a device selector.
   * The execution time is short and deterministic.
 
 
--- a/index.bs
+++ b/index.bs
@@ -1227,6 +1227,7 @@ and need less infrastructure set-up (such as making functions `async`).
 An API should generally be synchronous if the following rules of thumb apply:
 
   * The API is not expected to ever be gated behind a permission prompt, or another dialog such as a device selector.
+  * The API implementation will not be blocked by filesystem or network access.
   * The execution time is short and deterministic.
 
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1218,6 +1218,20 @@ discuss.</p>
...
 
+<h3 id="synchronous">Use synchronous when appropriate</h3>
+
+Synchronous APIs should be considered when minting a new API.
+Providing an API in synchronous form simplifies the usage from the caller's perspective.
+Moreover, it reduces the amount of infrastructural setup overhead needed for the function call.
+
+The general rule of thumb for an API that should be synchronous is the following:
+
+  * The API call is not expected to fail.
```
annevk, line 1229: Why does this matter?

cynthia: The synchronous part, or the not-failing part?

annevk: Not failing. A lot of non-promise returning methods have some kind of failure scenario.

torgo: Just revisiting this today. Is there a compromise text we could introduce here? @cynthia

plinss:
Wondering if this could be better stated as reasons when to make methods async, and then say everything else should be sync.

index.bs
cynthia: This was a suggestion from @kenchris - I think we can probably remove that clause and add it back when we have more context.

index.bs
```diff
@@ -1218,6 +1218,20 @@ discuss.</p>
...
 
+<h3 id="synchronous">Use synchronous when appropriate</h3>
+
+Synchronous APIs should be considered when minting a new API.
+Providing an API in synchronous form simplifies the usage from the caller's perspective.
+Moreover, it reduces the amount of infrastructural setup overhead needed for the function call.
+
+The general rule of thumb for an API that should be synchronous is the following:
+
+  * The API call is not expected to be gated behind a permission.
```
kenchris, line 1229: , now or in the future.

index.bs
```diff
@@ -1218,6 +1218,20 @@ discuss.</p>
 </div>
 
 
+<h3 id="synchronous">Use synchronous when appropriate</h3>
+
+Synchronous APIs should be considered when minting a new API.
```
kenchris, line 1223: adding/creating instead of minting

index.bs
```diff
@@ -1218,6 +1218,20 @@ discuss.</p>
...
+Providing an API in synchronous form simplifies the usage from the caller's perspective.
+Moreover, it reduces the amount of infrastructural setup overhead needed for the function call.
+
+The general rule of thumb for an API that should be synchronous is the following:
+
+  * The API call is not expected to be gated behind a permission.
+  * The execution time is short and deterministic.
+
+No API call is completely fail-proof; the expectation above is for the general case.
```
kenchris, line 1232: Remove

index.bs
```diff
@@ -1218,6 +1218,18 @@ discuss.</p>
 </div>
 
 
+<h3 id="synchronous">Use synchronous when appropriate</h3>
+
+Synchronous APIs should be considered when minting a new API.
+Providing an API in synchronous form simplifies the usage from the caller's perspective.
```
alice, line 1224:
```suggestion
Synchronous APIs are simpler to use,
and need less infrastructure set-up (such as making functions `async`).
```

index.bs
```diff
@@ -1218,6 +1218,18 @@ discuss.</p>
...
+Moreover, it reduces the amount of infrastructural setup overhead needed for the function call.
```
alice, line 1225:
```suggestion
```

index.bs
```diff
@@ -1218,6 +1218,18 @@ discuss.</p>
 </div>
 
 
+<h3 id="synchronous">Use synchronous when appropriate</h3>
+
+Synchronous APIs should be considered when minting a new API.
```
alice, line 1223:
```suggestion
Synchronous APIs should be considered when creating a new API.
```

index.bs
```diff
@@ -1218,6 +1218,18 @@ discuss.</p>
 </div>
 
 
+<h3 id="synchronous">Use synchronous when appropriate</h3>
+
+Synchronous APIs should be considered when creating a new API.
```
alice, line 1223:
```suggestion
Where possible, prefer synchronous APIs when designing a new API.
```

Switch to active voice.

index.bs
```diff
@@ -1218,6 +1218,18 @@ discuss.</p>
...
+Synchronous APIs are simpler to use,
+and need less infrastructure set-up (such as making functions `async`).
+
+The general rule of thumb for an API that should be synchronous is the following:
```
alice, line 1227:
```suggestion
An API should generally be synchronous if the following rules of thumb apply:
```

index.bs
```diff
@@ -1218,6 +1218,18 @@ discuss.</p>
...
+
+  * The API at no point is expected to be gated behind a permission or a dialog, such as a device selector.
```
alice, line 1229:
```suggestion
  * The API is not expected to ever be gated behind a permission prompt, or another dialog such as a device selector.
```

alice (approved):
Advice looks good, just some editorial tweaks.

annevk:
I would add "any kind of I/O" to the list of exceptions and maybe also "potential for the answer to be in a different process" given modern browser architectures.

cynthia:
> I would add "any kind of I/O" to the list of exceptions and maybe also "potential for the answer to be in a different process" given modern browser architectures.

Reflected in latest commit. I dropped the "in a different process", because "accessing another process" will be done either through filesystem or network. (although the intent of that sentence including IPC might not be obvious to all spec authors, now that I think of it...)

annevk:
Maybe add "including inter-process communication" or some such?

cynthia:
> Maybe add "including inter-process communication" or some such?

annevk:
"Including" acknowledges that while also reminding folks that might not realize, right? (You could make it a parenthetical.) However, I think the more important thing is to somehow point out that even if it's not the case today, it might be tomorrow. And perhaps not even cross-process, but simply running in a different thread.

cynthia:
I threw in IPC as an example, for the thread example that would be a lock - so also added that.

torgo (approved):

alice (approved):

PR #219, opened 2020-07-07
Readability edits: Basic principles section

alice:
***
<a href="https://pr-preview.s3.amazonaws.com/alice/design-principles/pull/219.html" title="Last updated on Jul 19, 2020, 11:50 PM UTC (96cf532)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/219/d0ad23a...alice:96cf532.html" title="Last updated on Jul 19, 2020, 11:50 PM UTC (96cf532)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -163,48 +163,44 @@ If this feature was proposed today, it would probably not proceed.
 
 <h3 id="consent">Ask users for meaningful consent when appropriate</h3>
 
-In some cases,
-we might add features to the Web that are not appropriate to allow for any website a user visits,
-but that are reasonable to have if the user agrees to use the feature
-based on a reasonable understanding of what will happen as a result.
-One example of such a feature might be support for location:
-many users seem likely to understand what it means to share their current location with a website
-and be able to consent to doing so
-(even though they might not fully understand the privacy implications of doing so).
-Another example of such a feature would be camera access,
-where users understand that they're sharing the image
-in the rectangle visible to the camera
-(even if they might not understand everything someone might do with the data).
+If a useful feature has the potential to cause harm to users,
+make sure that the user can give *meaningful consent* for that feature to be used,
+and that they can refuse consent effectively.
 
-We should not depend on asking the user for consent
-(via a permission prompt or other mechanism)
-if we don't have a way to express that request in a way
-that users will understand what is being requested
-and the main implications of giving their consent.
-Understanding whether this is the case may require
-studying users understanding of potential user interfaces,
-although in many cases experienced user experience designers
-may be able to extrapolate from previous experience.
+In order to give *meaningful consent*, the user must:
+- **understand** what permission they may choose whether to grant the web page
+- be able to choose to give or refuse that permission **effectively**.
 
-We should also not ask users to consent to something
-that can also happen if the user does not consent.
-The existence of prompts
-that request a user's consent for something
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -96,144 +96,149 @@ See also:
...
-Features on the web should work across different input and output devices,
-different platforms, and different media,
-as much as this is reasonably possible.
-For example, they should not be specific to a particular screen size,
-a particular operating system or browser engine,
-or assume the user is using a mouse.
+As much as possible,
+ensure that features on the web work across different input and output devices,
+different screen sizes, different interaction modes,
+different platforms, and different media.
```
cynthia, line 211: Feels like different could be simplified to one instance here.

index.bs
```diff
@@ -96,144 +96,149 @@ See also:
...
+<div class="example">
+The <a href="https://www.w3.org/TR/CSS2/visuren.html#block-formatting">`display: block`</a>,
+<a href="https://drafts.csswg.org/css-flexbox-1/#box-model">`display: flex`</a>
+and <a href="https://www.w3.org/TR/css-grid-1/#grid-containers">`display: grid`</a> layout models in CSS
 all default to placing content within the available space and without overlap,
-so that it works across devices of various sizes,
-across browsers and operating systems that have different fonts available,
-and across different user settings for default text size.
+so that it works across screen sizes,
+and allows users to choose their own font and font size without causing text to overflow.
```
torgo, line 240: I think the "across browsers" thing shouldn't get lost here.

index.bs

index.bs
alice: What does it mean in this context? "Across browsers" seems more about interoperability of implementation than about API design.

dbaron: The existing text there was about availability of fonts (across browsers and operating systems).  One way pages in some layout systems (e.g., absolute positioning) are brittle is when they depend on a particular font.  For example, depending on having "Times New Roman" installed for a particular piece of text to fit in `173px` of width without breaking into multiple lines, in a case where in many of the fallback fonts it would break into two lines and overlap other text that's positioned there.

alice: I see. 

Is that a critical detail to convey in this example (or a separate example)?

index.bs
dbaron: It could, although I sometimes intentionally repeat the last common word before a branching structure in a sentence.  Doing this makes it more obvious where the sentence branches and thus avoids ambiguity (where it's not clear where the part after a comma is picking up from).

index.bs
```diff
@@ -96,144 +96,148 @@ See also:
...
-We should seek to give users a true picture of the safety of the Web
-so that they can act based on that understanding of safety
-(in comparison to the safety of alternatives),
-and at the same time we should seek to keep the Web safe for its users
-and work to make it safer.
+    try to change both technical foundations and expectations.
+
+A new feature which introduces safety risks may still improve user safety overall,
+if it allows users to perform a task more safely on a web page
+than it would be for them to install an app to do the same thing.
```
torgo, line 134: Probably a good idea to specifically call out "native app" here rather than just "app" because app can sometimes mean web app.

alice:
Will wait for @dbaron to approve, especially on the Media Independence principle.

PR #220, opened 2020-07-07
Readability edits: API Design Across Languages section

alice:
***
<a href="https://pr-preview.s3.amazonaws.com/alice/design-principles/pull/220.html" title="Last updated on Sep 9, 2020, 5:56 AM UTC (b0cb8ee)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/220/d0ad23a...alice:b0cb8ee.html" title="Last updated on Sep 9, 2020, 5:56 AM UTC (b0cb8ee)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -346,35 +346,48 @@ limiting the feature to secure contexts might cause problems
 for code (such as libraries) that may be used in either secure or non-secure contexts.
 </div>
 
-<h3 id="do-not-expose-use-of-private-browsing-mode">Do not reveal that private browsing mode is engaged</h3>
+<h3 id="do-not-expose-use-of-private-browsing-mode">Don't reveal that private browsing mode is engaged</h3>
+
+Make sure that your feature doesn't give authors a way to detect private browsing mode.
 
 Some people use [private browsing mode](https://www.w3.org/2001/tag/doc/private-browsing-modes/)
-for their own <a href="#safe-to-browse">safety</a>;
-just the fact that they have or are using private browsing mode
+for their own <a href="#safe-to-browse">safety</a>.
+Because of this, the fact that someone is using private browsing mode
 may be [sensitive information](https://www.w3.org/TR/security-privacy-questionnaire/#sensitive-data) about them.
-[Consider the harm](https://www.w3.org/2001/tag/doc/ethical-web-principles/#noharm) that could be visited upon people—especially vulnerable people—were
-their use of private browsing mode revealed
-to others who wield power over them
-(such as employers,
-parents,
-partners,
+This information may [harm](https://www.w3.org/2001/tag/doc/ethical-web-principles/#noharm) people
+if it is revealed to a web site contolled by others who have power over them
+(such as employers, parents, partners,
 or [state actors](https://www.w3.org/2001/tag/doc/ethical-web-principles/#expression)).
 
 Given such dangers,
 websites should not be able to detect that private browsing mode is engaged.
 
 <div class="example">
-
 User Agents which support [IndexedDB](https://www.w3.org/TR/IndexedDB/)
 should not disable it in private browsing mode,
 because that would reveal that private browsing mode is engaged
-
 </div>
 
-See also
+<div class="example">
+The [Payment Request API](https://www.w3.org/TR/payment-request/)'s
+[show() method](https://www.w3.org/TR/payment-request/#show-method),
+when called,
+allows User Agents <q cite="https://www.w3.org/TR/payment-request/#show-method">to act
+as if the user had immediately
+[aborted the payment request](https://www.w3.org/TR/payment-request/#dfn-user-aborts-the-payment-request)</q>.
+
+This enables User Agents to automatically abort payment requests
+in private browsing mode
+(thus protecting [sensitive information](https://www.w3.org/TR/security-privacy-questionnaire/#sensitive-data)
+such as the user's shipping or billing address)
+without revealing that private browsing mode is engaged.
+</div>
+
+See also:
 
 * [Security and privacy are essential](https://www.w3.org/2001/tag/doc/ethical-web-principles/#privacy)
 * [What data does this specification expose to an origin?](https://www.w3.org/TR/security-privacy-questionnaire/#underlying-platform-data)
+* [[#private-browsing-mode]]
 
 <h3 id="private-browsing-mode">Consider how your API should behave in private browsing mode</h3>
 
--- a/index.bs
+++ b/index.bs
@@ -240,11 +240,22 @@ and across different user settings for default text size.
 <h3 id="simplicity">Prefer simple solutions</h3>
 <!-- was "Avoid Needless Complexity" in the HTML Design Principles -->
 
+Look hard for simple solutions to the
+<a href="#priority-of-constituencies">user needs</a>
+you intend to address.
+
 Simple solutions are generally better than complex solutions,
 although they may be harder to find.
 Simpler features are easier for user agents to implement and test,
 more likely to be interoperable,
-and easier for authors to understand. See related [[LEAST-POWER]]
+and easier for authors to understand.
+
+Make sure that your <a href="#priority-of-constituencies">user needs</a>
+are well-defined, so that you can avoid scope creep.
+
+See also:
+
+* [[LEAST-POWER]]
 
 <h3 id="naming-is-hard">Naming things</h3>
 
--- a/index.bs
+++ b/index.bs
@@ -277,7 +277,6 @@ because it hasn't been implemented yet, or
 because it's only available in <a href="#secure-context">secure contexts</a>.
 
 Authors shouldn't need to write different code to handle each scenario.
-and shouldn't be able to know which cases applied.
 
 <p class="note">
 When a feature is available but isn't feasible to use
--- a/index.bs
+++ b/index.bs
@@ -377,7 +377,7 @@ for code (such as libraries) that may be used in either secure or non-secure con
 Make sure that your feature doesn't give authors a way to detect private browsing mode.
 
 Some people use [private browsing mode](https://www.w3.org/2001/tag/doc/private-browsing-modes/)
-for their own <a href="#safe-to-browse">safety</a>.
+to protect their own personal safety.
 Because of this, the fact that someone is using private browsing mode
 may be [sensitive information](https://www.w3.org/TR/security-privacy-questionnaire/#sensitive-data) about them.
 This information may [harm](https://www.w3.org/2001/tag/doc/ethical-web-principles/#noharm) people
--- a/index.bs
+++ b/index.bs
@@ -303,6 +303,13 @@ if there is a user need or design principle which would fail
 if feature detection were available for the feature,
 then you should not support feature detection.
 
+Also, if a feature is generally not exposed to developers,
+it is not appropriate to support feature detection.
+For example, private browsing mode is a concept
+which is recognised in web specifications,
+but not exposed to authors.
+For private browsing mode to support the user's needs,
+it must not be feature detected.
 See also:
 
 * [[#do-not-expose-use-of-private-browsing-mode]]
--- a/index.bs
+++ b/index.bs
@@ -277,7 +277,10 @@ An existing feature may not be available on a page for a number of reasons.
 Two of the more common reasons are
 because it hasn't been implemented yet, or
 because it's only available in <a href="#secure-context">secure contexts</a>.
+
 Authors shouldn't need to write different code to handle each scenario.
+That way, even if an author only knows or cares about one scenario,
+the code will handle all of them.
 
 <p class="note">
 When a feature is available but isn't feasible to use
--- a/index.bs
+++ b/index.bs
@@ -257,216 +257,13 @@ See also:
 
 * [[LEAST-POWER]]
 
-<h3 id="naming-is-hard">Naming things</h3>
+<h3 id="name-thoughtfully">Name things thoughtfully</h3>
 
-Names take meaning from:
+Name APIs with care.
+Naming APIs well makes it much easier for authors to use them correctly.
 
-* signposting (the name itself)
-* use (how people come to understand the name over time)
-* context (the object on the left-hand side, for example)
-
-<h4 id="naming-common-words" class="no-num no-toc">Use common words</h4>
-
-API naming *must* be done in easily readable US English.
-Keep in mind that most web developers are not native English speakers.
-Whenever possible, names should be chosen that use common vocabulary
-non-native English speakers are likely to understand when first encountering the name.
-
-<div class="example">
-
-For example `horizontalAlignment` is
-a more English-readable name
-than `alignmentHorizontal`.
-
-</div>
-
-Value readability over brevity,
-though always keep in mind that sometimes
-the shorter name is the clearer one.
-For instance,
-it may be appropriate to use technical language or well-known terms of art
-in the specification where the API is defined.
-
-<div class="example">
-
-For example,
-the Fetch API's {{Body}} mixin's {{Body/json()}} method
-is named for the kind of object it returns.
-JSON is a well-known term of art
-among web developers likely to use the Fetch API.
-It would harm comprehension
-to name this API less directly connected to its return type. [[FETCH]]
-
-</div>
-
-<h4 id="naming-charset" class="no-num no-toc">Use ASCII names</h4>
-
-Names must adhere to the local language restrictions, for example CSS ident rules etc.
-and *should* be in the [=ascii code point|ASCII range=].
-
-<h4 id="naming-consultation" class="no-num no-toc">Consultation</h4>
-
-Please consult widely on names in your APIs.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -267,43 +267,44 @@ for specific guidance on naming.
...
+should be based on the <a href="#priority-of-constituencies">user need</a>
+for the feature:
+if there is a user need or design principle which would fail
+if feature detection were available for the feature,
+then you should not support feature detection.
+
+See also:
+
+* [[#do-not-expose-use-of-private-browsing-mode]]
+* [[#do-not-expose-use-of-assistive-tech]]
```
dbaron, line 306: I would prefer to retain the general point that feature detection is for features that developers use and not for features that users use.  (In other words, it's for things that are part of the API surface of the Web.)

index.bs
```diff
@@ -267,43 +267,44 @@ for specific guidance on naming.
...
-but in some cases the feature should also be detectable
-in the language where it is used
+If appropriate, provide a way for authors to programmatically detect
+whether your feature is available,
+so that web content may gracefully handle the feature not being present.
+
+An existing feature may not be available on a page for a number of reasons, including
+- because it has not been implemented yet, or
+- because it's only available in <a href="#secure-context">secure contexts</a>, or
+- because the user denied the page permission to use the feature.
```
dbaron, line 277: I don't think permission denial is a good example here, because at that point the developer has used part of the feature already to cause the user to be asked for permission.

index.bs
```diff
@@ -267,43 +267,44 @@ for specific guidance on naming.
...
+
+Authors should need to write different code to handle each scenario,
```
dbaron, line 279: I think this is saying the opposite of what is intended.  (The same code should work for all scenarios.)

index.bs
```diff
@@ -267,43 +267,44 @@ for specific guidance on naming.
...
+and should not be able to deduce which cases applied.
```
dbaron, line 280: perhaps change "deduce" to either "determine" or "figure out" or "find out"

index.bs
```diff
@@ -267,43 +267,44 @@ for specific guidance on naming.
...
+
```
dbaron, line 281: I think it would be good to add back the justification, e.g., by adding a second sentence to the paragraph like "This means that an author who writes code to handle one of these scenarios will likely have handled all of them."

index.bs
```diff
@@ -267,43 +267,44 @@ for specific guidance on naming.
...
+<p class="note">
+When a feature is available but isn't feasible to use
+because a required device isn't present,
+it may be better to expose that the feature is available
```
dbaron, line 285: I think "may be" weakens the advice too much here.  Perhaps "is usually" or "is generally"?

index.bs
```diff
@@ -267,43 +267,44 @@ for specific guidance on naming.
...
-(e.g., the feature not being implemented yet in some browsers)
-also works in other cases
-(e.g., a library being used in a non-secure context when
-the feature is limited to secure contexts).
-
-However, some features may be for
-interaction with devices that may or may not be available.
-Detection of device availability should generally be separate from feature detection.
-This helps developers who may want to handle device unavailability
-differently from lack of support for the feature.
```
dbaron, line 482: I'd somewhat prefer to retain this justification.

index.bs
```diff
@@ -308,57 +308,43 @@ See also:
...
 
+<div class="example">
+One example of a feature that should be limited to secure contexts
+is [Geolocation](https://www.w3.org/TR/geolocation-API/),
+since it would be a risk to users' privacy to transmit their location in an insecure way.
+</div>
+
+For other features, carefully weigh up whether
+<a href="priority-of-constituencies">user needs</a> are best served
+by making the feature available in an insecure context.
```
dbaron: So the previous text here reflected a lack of TAG consensus on whether we should be asking for:
* all new features to be limited to secure contexts (in order to move the web more quickly to https, which does serve user needs given that many of the existing features of the web are often privacy-sensitive), or
* only limiting features to secure contexts when there's a specific reason that they're *more* sensitive

This text much more concretely takes a position on that question where we didn't previously have consensus (see #32, #75, and #89 for the history).

If I were to try to summarize that disagreement clearly, I'd say (replacing the last sentence here):

> For other features, the TAG does not have consensus.  Some members of the TAG believe that all clearly-distinct new features should be limited to secure contexts.  Others believe that those designing new features should make decisions for each feature based on whether user needs are best served by making the feature available in an insecure context.

index.bs
```diff
@@ -308,57 +308,43 @@ See also:
...
+without the authentication, integrity, or confidentiality
+that's present only in secure contexts.
 
+<div class="example">
+One example of a feature that should be limited to secure contexts
+is [Geolocation](https://www.w3.org/TR/geolocation-API/),
+since it would be a risk to users' privacy to transmit their location in an insecure way.
+</div>
+
+For other features, carefully weigh up whether
```
dbaron, line 322: "weigh up" seems odd to my ears.  Maybe just "weigh", or "consider"?

index.bs
```diff
@@ -308,57 +308,43 @@ See also:
...
+<div class="note">
 Specification authors can limit most features defined in
 <a href="https://heycam.github.io/webidl/">Web IDL</a>,
 to secure contexts
 by using the
 <code>[<a href="https://w3c.github.io/webappsec-secure-contexts/#integration-idl">SecureContext</a>]</code> extended attribute
 on interfaces, namespaces, or their members (such as methods and attributes).
-Similar ways of marking features as limited to secure contexts should be added
-to other major points where the Web platform is extended over time
-(for example, the definition of a new CSS property).
```
dbaron, line 528: I think we should still be asking for more mechanisms like `[SecureContext]` to enable more things that aren't defined in WebIDL to be defined as restricted to secure contexts.

index.bs
```diff
@@ -346,35 +346,48 @@ limiting the feature to secure contexts might cause problems
...
 because that would reveal that private browsing mode is engaged
+</div>
+
+<div class="example">
+The [Payment Request API](https://www.w3.org/TR/payment-request/)'s
+[show() method](https://www.w3.org/TR/payment-request/#show-method),
+when called,
+allows User Agents <q cite="https://www.w3.org/TR/payment-request/#show-method">to act
+as if the user had immediately
+[aborted the payment request](https://www.w3.org/TR/payment-request/#dfn-user-aborts-the-payment-request)</q>.
```
dbaron: It's not clear to me that this is a good example.  Are there other cases where the UA will act as though the user had immediately aborted the payment request, or is that a clear sign the user is in private mode?  I think if you want to explain why this is a good example, you should explain what other things would appear the same to the UA.

index.bs
```diff
@@ -391,58 +391,30 @@ See also:
 
 <h3 id="private-browsing-mode">Consider how your API should behave in private browsing mode</h3>
 
-In general,
-APIs should behave the same in and out of [private browsing mode](https://www.w3.org/2001/tag/doc/private-browsing-modes/).
-This is because
-the use of private browsing mode should not be revealed to sites.
-See [[#do-not-expose-use-of-private-browsing-mode]].
```
dbaron, line 560: I think it would be good to retain some reference to the other section, since somebody might be linked here directly.

index.bs
```diff
@@ -391,58 +391,30 @@ See also:
...
-as discussed in our [privacy & security questionnaire](https://www.w3.org/TR/security-privacy-questionnaire/#private-browsing).
+If necessary, specify how your API should behave differently
+in private browsing mode.
 
-For instance,
-if your API would reveal information
-that would allow for the correlation
-of a single user’s activity
+For example, if your API would reveal information
+that would allow someone to correlation
```
dbaron, line 398: correlation -> correlate

dbaron (dismissed):

dbaron:
I've now reviewed all the commits in this PR.

index.bs
hober: @dbaron, I suppose I'm one of those others, and I don't think your last sentence represents my position. I believe that bifurcating the web's API surface is harmful, and therefore, only features with privacy or security impact should be limited to secure contexts.

alice: I had a go at representing both of these positions.

index.bs
alice: Ok, I tried to capture that.

index.bs
```diff
@@ -346,35 +346,48 @@ limiting the feature to secure contexts might cause problems
...
+Because of this, the fact that someone is using private browsing mode
 may be [sensitive information](https://www.w3.org/TR/security-privacy-questionnaire/#sensitive-data) about them.
-[Consider the harm](https://www.w3.org/2001/tag/doc/ethical-web-principles/#noharm) that could be visited upon people—especially vulnerable people—were
-their use of private browsing mode revealed
-to others who wield power over them
-(such as employers,
-parents,
-partners,
+This information may [harm](https://www.w3.org/2001/tag/doc/ethical-web-principles/#noharm) people
+if it is revealed to a web site contolled by others who have power over them
```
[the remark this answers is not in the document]

index.bs
alice: Hm, I just moved this up from the other private browsing section, since it seemed to fit better here.

I think @hober might have written the original example?

index.bs
alice: Added to "See also" links at the bottom.

index.bs

index.bs
dbaron: OK, maybe the right distinction is just features that are exposed to developers in some way versus those that aren't.

Other examples of features that aren't exposed to developers:  find in page (although we're thinking about it), reader mode, view source, developer tools, perhaps browser UI generally (e.g., shortcut keys, ability to look at back/forward history before navigating it, etc.).

index.bs
dbaron: I think it points out the value to users that, while the author might not care about all of the scenarios, the user gets the value across all of them of the author caring about just one of them.

index.bs
dbaron: looks good

index.bs
dbaron: I think that works.

index.bs
dbaron: Hmmm.  On second thought I'm not sure either of them is right.  An author with knowledge of browsers could certainly figure this out.  It's just that we want a single codepath for the different failure cases. because then it's a lot more likely to be tested than separate codepaths.

Maybe drop the whole bit after the comma?  And possibly replace it with something about testing, like "Then the single codepath to handle the feature being unavailable is more likely to be tested and work correctly."

index.bs
dbaron: I think it fits better in the other section (where it was originally) because I think it might actually be a bad example for this one... although I'm not sure, and it's possible it could be a good example here with more detail added to demonstrate that.

index.bs
alice: @hober since David is out, could you take a look at the wording here and see if it seems fair?

index.bs
alice: @hober Could use your opinion here as well

index.bs
```diff
@@ -240,344 +240,153 @@ and across different user settings for default text size.
...
-Look at other web platform specifications,
-and seek advice from others working in related areas of the platform.
+An existing feature may not be available on a page for a number of reasons.
+Two of the more common reasons are
+because it hasn't been implemented yet, or
+because it's only available in <a href="#secure-context">secure contexts</a>.
 
-Pay particular attention to advice you receive
-with clearly-stated rationale based on underlying principles.
+Authors shouldn't need to write different code to handle each scenario,
```
alice, line 279:
```suggestion
Authors shouldn't need to write different code to handle each scenario.
```

index.bs
```diff
@@ -240,344 +240,153 @@ and across different user settings for default text size.
...
+and shouldn't be able to know which cases applied.
```
alice, line 280:
```suggestion
```

index.bs
```diff
@@ -240,344 +240,152 @@ and across different user settings for default text size.
...
+(such as ''@supports'' in CSS).
 
-Naming should be generic and future-proof whenever possible.
+In some cases, it may **not** be appropriate to allow feature detection.
+Whether the feature should be detectable or not
+should be based on the <a href="#priority-of-constituencies">user need</a>
+for the feature:
+if there is a user need or design principle which would fail
+if feature detection were available for the feature,
+then you should not support feature detection.
```
alice, line 305:
```suggestion

Also, if a feature is generally not exposed to developers,
it is not appropriate to support feature detection.
For example, private browsing mode is a concept
which is recognised in web specifications,
but not exposed to authors.
For private browsing mode to support the user's needs,
it must not be feature detected.
```

index.bs
```diff
@@ -240,344 +240,159 @@ and across different user settings for default text size.
...
+
+Also, if a feature is generally not exposed to developers,
+it is not appropriate to support feature detection.
+For example, private browsing mode is a concept
+which is recognised in web specifications,
+but not exposed to authors.
+For private browsing mode to support the user's needs,
+it must not be feature detected.
+See also:
```
alice, line 313:
```suggestion

See also:
```

index.bs
alice: I added an extra paragraph on this.

index.bs
```diff
@@ -240,344 +240,160 @@ and across different user settings for default text size.
...
-
-<h4 id="naming-common-words" class="no-num no-toc">Use common words</h4>
+and easier for authors to understand.
 
-API naming *must* be done in easily readable US English.
-Keep in mind that most web developers are not native English speakers.
-Whenever possible, names should be chosen that use common vocabulary
-non-native English speakers are likely to understand when first encountering the name.
+Make sure that your <a href="#priority-of-constituencies">user needs</a>
+are well-defined, so that you can avoid scope creep.
```
cynthia, line 254: Not sure what would be the best wording, but assuming scope creep is overengineering, there is also the problem of underengineering. (e.g. not taking non-latin languages into account for an API that deals with strings)

index.bs
hober: Yes, it seems fair to me.

index.bs
hober: I think this is a pre-existing problem with this example that is made a bit more obvious by the move. We should improve the example in a follow-on PR. I think @alice is filing an issue for this.

index.bs
```diff
@@ -240,344 +240,160 @@ and across different user settings for default text size.
...
-What are similar APIs named on other platforms,
-or in popular libraries in various programming languages?
-Ask end users and developers what they call
-things that your API works with or manipulates.
-Look at other web platform specifications,
-and seek advice from others working in related areas of the platform.
+An existing feature may not be available on a page for a number of reasons.
+Two of the more common reasons are
+because it hasn't been implemented yet, or
+because it's only available in <a href="#secure-context">secure contexts</a>.
```
alice, line 278:
```suggestion
```

hober (approved):
This looks great! Thanks for all your hard work on this, @alice.

cynthia (approved):
LGTM.

index.bs
```diff
@@ -240,344 +240,159 @@ and across different user settings for default text size.
...
-See [[#do-not-expose-use-of-private-browsing-mode]].
+Make sure that your feature doesn't give authors a way to detect private browsing mode.
 
-However,
-there are some situations in which
-APIs should consider behaving differently
-when private browsing mode is engaged,
-as discussed in our [privacy & security questionnaire](https://www.w3.org/TR/security-privacy-questionnaire/#private-browsing).
+Some people use [private browsing mode](https://www.w3.org/2001/tag/doc/private-browsing-modes/)
+for their own <a href="#safe-to-browse">safety</a>.
```
torgo, line 378:
```suggestion
to protect their own personal safety.
```

index.bs
```diff
@@ -240,344 +240,159 @@ and across different user settings for default text size.
...
-among web developers likely to use the Fetch API.
-It would harm comprehension
-to name this API less directly connected to its return type. [[FETCH]]
-
-</div>
+and easier for authors to understand.
 
-<h4 id="naming-charset" class="no-num no-toc">Use ASCII names</h4>
+Make sure that your <a href="#priority-of-constituencies">user needs</a>
+are well-defined, so that you can avoid scope creep.
```
alice, line 254:
```suggestion
are well-defined.
This allows you to avoid scope creep,
and make sure that your API does actually meet the needs of all users.
```

index.bs
```diff
@@ -240,344 +240,161 @@ and across different user settings for default text size.
...
-</div>
+Provide a way for authors to programmatically detect
+whether your feature is available,
+so that web content may gracefully handle the feature not being present.
+
+An existing feature may not be available on a page for a number of reasons.
+Two of the more common reasons are
+because it hasn't been implemented yet, or
+because it's only available in <a href="#secure-context">secure contexts</a>.
+Authors shouldn't need to write different code to handle each scenario.
```
alice, line 280:
```suggestion

Authors shouldn't need to write different code to handle each scenario.
That way, even if an author only knows or cares about one scenario,
the code will handle all of them.
```

PR #221, opened 2020-07-07
Readability edits: Cascading Style Sheets section

alice:
***
<a href="https://pr-preview.s3.amazonaws.com/alice/design-principles/pull/221.html" title="Last updated on Sep 25, 2020, 12:07 AM UTC (752fade)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/221/d0ad23a...alice:752fade.html" title="Last updated on Sep 25, 2020, 12:07 AM UTC (752fade)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -778,27 +778,61 @@ that a property likely should have been inherited.
 
 <h3 id="css-computed-value">Choose the computed value type based on how the property should inherit</h3>
 
-Another important choice in the design of every CSS property
-is what its <a>computed value</a> is.
-This is an abstract description of the value that is inherited when the property inherits.
-It is also important for starting and animation of [=CSS Transitions=].
-This is most important for inherited properties, but it does matter for all properties.
+Choose the <a>computed value</a> of a CSS property
+based on how it will [=inherit=],
+and the other properties which may depend on it,
+whether or not it has been inherited.
 
-Note: Due to a somewhat unfortunate history,
-this is *not* the value that is returned by {{getComputedStyle()}},
-which is actually the <a>resolved value</a>,
-which can be the result of further computation starting from the computed value.
+<div class="note">
+Note: the <a>computed value</a> should not be confused with the [=resolved value=]
+which is returned from the {{getComputedStyle()}} method.
+</div>
 
-The definition of a CSS property often has interactions with other CSS properties.
-It is often possible for these interactions to happen at the [=computed value=] stage
-or at a later stage such as the [=used value=].
-Since inheritance operates on the computed value,
-whether an interaction should affect the value that descendants inherit
-is a major factor in the choice of where that interaction should be handled.
+Certain CSS properties are calculated with
+<a href="https://wiki.csswg.org/spec/property-dependencies">dependencies</a>
+on others:
+for example, the <a>computed value</a>
+of a property specified in 'lh' units
+depends on the 'line-height' property which applies to the element being considered.
+The <a>computed value</a> for a property should be chosen
+so that the dependent values make sense,
+regardless of whether they are inherited.
 
-<a href="https://wiki.csswg.org/spec/property-dependencies">The dependencies
-that do affect computed values</a>
-are required to not be circular.
+<div class="example">
+For example, the 'line-height' property may accept a <<number>> value,
+such as ''line-height: 1.4''.
+This value represents a multiple of the 'font-size',
+so if the 'font-size' is ''20px'',
+the [=actual value=] for the line height is ''28px''.
+
+However, the <a>computed value</a> in this case is the <<number>> ''1.4'',
+not the <<length>> ''28px''.
+
+The 'line-height' property can be inherited
+into elements that have a different 'font-size',
+and any property on those elements which depends on 'line-height'
+must take the relevant 'font-size' into account,
+rather than the 'font-size' for the element from which
+the 'line-height' value was inherited.
+
+  <pre highlight=html>
+&lt;body style="font-size: 20px; line-height: 1.4"&gt;
+
+  &lt;p&gt;This body text has a line height of 28px.&lt;/p&gt;
+
+  &lt;h2 style="font-size: 200%"&gt;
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -690,71 +690,67 @@ See also
...
-(though this image might not be fully opaque).
-The CSS painting model allows other things to draw on top of that background,
-including the backgrounds of descendant elements.
-But specifying ''background-image: none'' on a descendant element does not
-cause a hole to be poked in the ancestor element's background,
-since the entire effect comes from the property being used on a single element.
-Setting ''background-image: none'' is useful
-to override a declaration lower in the cascade on the same element,
-but this is a different feature from replacing the backgrounds that are drawn by ancestors.
+For example, the 'background-image' property is **not** inherited.
```
torgo, line 753: I'm not sure about dropping the explanation. Even with the above additional text I think this statement could still use an additional sentence of context reiterating the principle.

index.bs
```diff
@@ -874,6 +832,21 @@ These number values are generally the preferred values to use for 'line-height'
 because they inherit better than length values.
 </div>
 
+Issue: The paragraph below is hard to understand.
```
torgo, line 835: How about making this a clear recommendation to err on the side of reduced complexity? This is kind of implied but not explicitly stated. Something like “When making decisions about whether dependencies should affect computed values, try to not add undue complexity.“ ?

torgo (dismissed):
A few small comments and suggestions. Otherwise looks good.

index.bs
alice: There is still a whole paragraph below connecting it to the principle.

hober (approved):
Finally got a chance to look at this, and these changes look great to me. Thanks for improving the readability of this so much, @alice!

index.bs
alice, line 835:
```suggestion
```

index.bs
```diff
@@ -874,6 +832,21 @@ These number values are generally the preferred values to use for 'line-height'
...
+
+Making decisions about whether dependencies should affect computed values
+often involves tradeoffs of complexity.
+Sometimes it is less complex
+for the behavior to happen as part of the computation of the computed value;
+sometimes it is less complex
+for the behavior to happen in whatever handles the computed value.
+(This complexity might be in the specification, in implementations, or both.
+Complexity in implementations might affect how fast things run for end users.)
```
alice, line 844:
```suggestion
```

index.bs
```diff
@@ -874,6 +832,21 @@ These number values are generally the preferred values to use for 'line-height'
 because they inherit better than length values.
 </div>
 
+Issue: The paragraph below is hard to understand.
+
```
alice, line 836:
```suggestion
```

index.bs
```diff
@@ -874,6 +832,21 @@ These number values are generally the preferred values to use for 'line-height'
 because they inherit better than length values.
 </div>
```
alice, line 834:
```suggestion
```

alice (commented):
I'm removing the last paragraph, since I can't understand the advice it's trying to give. We might re-add it later with some concrete examples to illustrate the advice.

index.bs
```diff
@@ -780,92 +776,54 @@ This is the "code smell" mentioned above that suggests
...
+
+Certain CSS properties are calculated with
+<a href="https://wiki.csswg.org/spec/property-dependencies">dependencies</a>
+on others:
+for example, the <a>computed value</a>
+of a property specified in 'lh' units
+depends on the 'line-height' property which applies to the element being considered.
+The <a>computed value</a> for a property should be chosen
+so that the dependent values make sense,
+regardless of whether they are inherited.
```
dbaron, line 799: I'm not convinced that this is the case.  The `lh` unit could be defined to depend on something other than the computed value if that was necessary... at least if it's something after the computed value stage.  (This is the case for `ex`, `ch`, `cap`, and `ic`.)

It's also a substantially different point than what was here before.

I'm also a little queasy about the "dependent values" wording, because it sounds like it's introducing a term, even though it's not.

index.bs
```diff
@@ -780,92 +776,54 @@ This is the "code smell" mentioned above that suggests
...
-of walking up the ancestor chain looking for text decorations on ancestors.
-</div>
-
 <h3 id="css-computed-value">Choose the computed value type based on how the property should inherit</h3>
 
-Another important choice in the design of every CSS property
-is what its <a>computed value</a> is.
-This is an abstract description of the value that is inherited when the property inherits.
-It is also important for starting and animation of [=CSS Transitions=].
-This is most important for inherited properties, but it does matter for all properties.
```
dbaron, line 817: I think it's worth retaining these points.

index.bs
```diff
@@ -780,92 +776,54 @@ This is the "code smell" mentioned above that suggests
...
-often involves tradeoffs of complexity.
-Sometimes it is less complex
-for the behavior to happen as part of the computation of the computed value;
-sometimes it is less complex
-for the behavior to happen in whatever handles the computed value.
-(This complexity might be in the specification, in implementations, or both.
-Complexity in implementations might affect how fast things run for end users.)
-
-For more information about computed values, see
-[Computed Values Patterns](https://wiki.csswg.org/spec/computed-values).
```
dbaron, line 845: I'd like to try to get many of these points back in the document... although I admit they weren't very clear and lacked examples.  I'm hoping to try to write something soon...

dbaron (commented):
OK, I finally got a chance to look at this.  Most of it looks good, but I have a few comments, some of which are somewhat vague at this point and which I'd like to dig into further.  Hopefully I'll find some time to make a PR for at least some of these.

alice:
> OK, I finally got a chance to look at this. Most of it looks good, but I have a few comments, some of which are somewhat vague at this point and which I'd like to dig into further. Hopefully I'll find some time to make a PR for at least some of these.

Thanks for looking, let's try and get a PR up to address these ASAP.

PR #223, opened 2020-07-07
Readability edits: Event Design section

alice:
***
<a href="https://pr-preview.s3.amazonaws.com/alice/design-principles/pull/223.html" title="Last updated on Oct 7, 2020, 5:49 AM UTC (2820cb7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/223/4f5f94c...alice:2820cb7.html" title="Last updated on Oct 7, 2020, 5:49 AM UTC (2820cb7)">Diff</a>


index.bs
```diff
@@ -1300,253 +1300,359 @@ Promise-Using Specifications</a></strong> guideline.
...
+instead of directly on the relevant element interface(s).
+Similarly, add [=event handler IDL attributes=]
+to {{WindowEventHandlers}} rather than {{Window}}.
+</p>
 
 
 <h3 id="events-are-for-notification">Events are for notification</h3>
 
-Try to design DOM events to deliver after-the-fact notifications of changes. It may be tempting to try to trigger side-effects from the action of {{EventTarget/dispatchEvent()}}, but in general this is <a href="https://lists.w3.org/Archives/Public/public-webapps/2014AprJun/0510.html">strongly discouraged</a> as it requires changes to the DOM specification when added. Your design will proceed more quickly if you avoid this pattern.
+Issue: I can't understand this one well enough to rewrite it.
```
torgo, line 1340: I think it should definitely be re-written without the mailing list reference. How about “DOM events should deliver after-the-fact notifications of changes. They should not be designed to trigger additional actions.” Agree we need some examples.

index.bs
```diff
@@ -1300,253 +1300,359 @@ Promise-Using Specifications</a></strong> guideline.
...
-
-Finally, dispatching a deferred event that does not seem to require
-packaging extra state or defining a state machine for the algorithm as mentioned above,
-could mean that all of the state transitions have been completed and that the event
-is meant to signal completion of the algorithm. In this case, it's likely that instead
-of using an event to signal completion, the API should be designed to return and complete
-a {{Promise}} instead. See [[#one-time-events]].
+If using events causes problems,
+such as unavoidable [[#guard-against-recursion|recursion]],
+consider using an Observer pattern instead.
```
torgo, line 1516: link?

index.bs
dbaron: So an example for being "after-the-fact" is that when a `resize` event fires, the size of the element should be (as reported by `getBoundingClientRect`, etc.) the new size rather than the old size.  (And I think if multiple elements are changing size at the same time, probably all of their sizes should be the new sizes rather than the old sizes by the time the first `resize` event fires.)

I think an example for triggering additional actions would be that the script manually creating and dispatching a `resize` event shouldn't actually cause anything to resize.

index.bs
```diff
@@ -1300,253 +1300,359 @@ Promise-Using Specifications</a></strong> guideline.
...
-
-```js
-const filter = navigator.nfc.createReadFilter({
-  recordType: "json"
-});
-
-const onMessage = message => { … };
-filter.addEventListener('exchange', onMessage);
-```
+Note: This section has been removed. See [[#events-vs-observers]] instead.
```
alice, line 1315:
```suggestion
When creating an API which allows authors to start and stop a process,
don't combine this with subscribing to updates or notifications from the process.
Instead, use the existing event infrastructure for notifications, and create separate API controls to start/stop the underlying process.

See:
- [[#events-are-for-notification]]
- [[#events-vs-observers]
```

index.bs
alice: I attempted a rewrite of this in the latest change. PTAL?

index.bs
alice: I think the link would go to this section...

plinss (approved):

index.bs
```diff
@@ -1300,253 +1300,378 @@ Promise-Using Specifications</a></strong> guideline.
...
 
 <h3 id="events-are-for-notification">Events are for notification</h3>
 
-Try to design DOM events to deliver after-the-fact notifications of changes. It may be tempting to try to trigger side-effects from the action of {{EventTarget/dispatchEvent()}}, but in general this is <a href="https://lists.w3.org/Archives/Public/public-webapps/2014AprJun/0510.html">strongly discouraged</a> as it requires changes to the DOM specification when added. Your design will proceed more quickly if you avoid this pattern.
+Events shouldn't be used to trigger changes,
+only to deliver a notification that a change has already finished happening.
+
+<div class="example">
+When a window is resized,
+an event named [`resize`](https://drafts.csswg.org/cssom-view/#eventdef-window-resize)
```
alice, line 1361:
```suggestion
an event named <code>[resize](https://drafts.csswg.org/cssom-view/#eventdef-window-resize)</code>
```

index.bs
```diff
@@ -1300,253 +1300,378 @@ Promise-Using Specifications</a></strong> guideline.
...
 
 <h3 id="guard-against-recursion">Guard against potential recursion</h3>
 
-When designing a long-running or complicated algorithm that is initiated by an
-API call, events are appropriate for [[#events-are-for-notification|notifying]]
-user code of the ongoing process. However, they also introduce the possibility of unexpected
-re-execution of the current algorithm before it has finished! Because user code gets
-to run in the middle of the algorithm, recursion can happen if that user code triggers
-the algorithm again (directly or indirectly).
+Issue: This gets very into the weeds.
```
alice, line 1372:
```suggestion
```

index.bs
```diff
@@ -1300,253 +1300,378 @@ Promise-Using Specifications</a></strong> guideline.
...
```
alice, line 1373:
```suggestion
```

index.bs
```diff
@@ -1300,253 +1300,378 @@ Promise-Using Specifications</a></strong> guideline.
...
 
-    6. Set {{MyObject|this}}.[[{{started}}]] to `false`.
-</div>
+<h3 id="events-vs-observers">How to decide between Events and Observers</h3>
 
-Note: A caution about early termination: if the algorithm being terminated would go
-on to ensure some critical state consistency, be sure to also make the relevant adjustments
-in state before early termination of the algorithm. Not doing so can lead to inconsistent
-state and end-user-visible bugs when implemented as-specified.
+Issue: This gets very into the weeds.
```
alice, line 1523:
```suggestion
```

index.bs
```diff
@@ -1300,253 +1300,378 @@ Promise-Using Specifications</a></strong> guideline.
...
-    4. [=Fire an event=] named `"currentlyinprogress"` at {{MyObject|this}}.
+Where possible, use a plain {{Event}} with a specified `type`,
+and capture any state information in the {{Event/target}} object.
 
-    5. ...Finish up complex stuff...
+It's usually not necessary to create new subclasses of {{Event}}.
 
-    6. Set {{MyObject|this}}.[[{{started}}]] to `false`.
-</div>
+<h3 id="events-vs-observers">How to decide between Events and Observers</h3>
```
alice, line 1522:
```suggestion
```

torgo (approved):

dbaron:
@alice This broke the build due to another bad indent, and the errors are in fact more readable after #244.

PR #226, opened 2020-07-07
Readability edits: Other API Design Considerations section

alice:
***
<a href="https://pr-preview.s3.amazonaws.com/alice/design-principles/pull/226.html" title="Last updated on Oct 7, 2020, 6:04 AM UTC (7627724)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/226/d0ad23a...alice:7627724.html" title="Last updated on Oct 7, 2020, 6:04 AM UTC (7627724)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1976,16 +1976,6 @@ need to be careful when exposing these to the web platform.
 
 <h2 id="other-considerations">Other API Design Considerations</h2>
 
-<h3 id="design-resources">Privacy, Security, Accessibility, and Internationalization</h3>
-
-It is important not to neglect other aspects of API design such as privacy, security, accessibility, and internationalization.
-Please take advantage of these other excellent resources in your design process:
-
-* <a href="https://w3c.github.io/privacy-considerations/">Privacy Considerations for Web Protocols</a>
-* <a href="https://www.w3.org/TR/security-privacy-questionnaire/">Self-Review Questionnaire: Security and Privacy</a>
-* <a href="https://w3c.github.io/apa/fast/">Web Technology Accessibility Guidelines</a>
-* <a href="https://w3c.github.io/bp-i18n-specdev/">Internationalization Best Practices for Spec Developers</a> <!-- also at https://www.w3.org/TR/international-specs/ -->
-
 <h3 id="polyfills">Polyfills</h3>
 
 Polyfills can be hugely beneficial in helping to roll out new features to the web platform. The
@@ -2179,6 +2169,12 @@ Some useful advice on how to write specifications is available elsewhere:
     specifications: Kinds of statements</a> (Ian Hickson, 2006)
   * <a href="https://www.w3.org/TR/qaframe-spec/">QA Framework:
     Specification Guidelines</a> (W3C QA Working Group, 2005)
+  * <a href="https://w3c.github.io/privacy-considerations/">Privacy Considerations for Web Protocols</a>
+  * <a href="https://www.w3.org/TR/security-privacy-questionnaire/">Self-Review Questionnaire: Security and Privacy</a>
+  * <a href="https://w3c.github.io/apa/fast/">Web Technology Accessibility Guidelines</a>
+  * <a href="https://w3c.github.io/bp-i18n-specdev/">Internationalization Best Practices for Spec Developers</a> <!-- also at https://www.w3.org/TR/international-specs/ -->
+
+
 
 <h2 id="acks" class="no-num">Acknowledgments</h2>
 
--- a/index.bs
+++ b/index.bs
@@ -1978,11 +1978,14 @@ need to be careful when exposing these to the web platform.
 
 <h3 id="polyfills">Polyfills</h3>
 
-Polyfills can be hugely beneficial in helping to roll out new features to the web platform. The
-Technical Architecture Group <a href="https://www.w3.org/2001/tag/doc/polyfills/">finding on Polyfills and the Evolution of the Web</a> offers guidance
-that should be considered in the development of new features, notably:
+Issue: is this advice still current? When was the last time we saw a review with a speculative polyfill?
 
-* Being "polyfillable" is not essential but <strong>is</strong> beneficial
+Polyfills can be hugely beneficial
+in helping to roll out new features to the web platform.
+The Technical Architecture Group <a href="https://www.w3.org/2001/tag/doc/polyfills/">finding on Polyfills and the Evolution of the Web</a>
+offers guidance that should be considered in the development of new features, notably:
+
+* Being "polyfillable" isn't essential but <strong>is</strong> beneficial
 * [[#feature-detect]]
 * Polyfill development should be encouraged
 
--- a/index.bs
+++ b/index.bs
@@ -1993,7 +1993,7 @@ offers guidance that should be considered in the development of new features, no
 
 If your feature requires a manifest,
 investigate whether you can extend an existing manifest schema.
-
+New web features should be self-contained and self-describing and ideally should not require an additional manifest file.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1976,52 +1976,23 @@ need to be careful when exposing these to the web platform.
...
-* <a href="https://www.w3.org/TR/security-privacy-questionnaire/">Self-Review Questionnaire: Security and Privacy</a>
-* <a href="https://w3c.github.io/apa/fast/">Web Technology Accessibility Guidelines</a>
-* <a href="https://w3c.github.io/bp-i18n-specdev/">Internationalization Best Practices for Spec Developers</a> <!-- also at https://www.w3.org/TR/international-specs/ -->
-
 <h3 id="polyfills">Polyfills</h3>
 
-Polyfills can be hugely beneficial in helping to roll out new features to the web platform. The
-Technical Architecture Group <a href="https://www.w3.org/2001/tag/doc/polyfills/">finding on Polyfills and the Evolution of the Web</a> offers guidance
-that should be considered in the development of new features, notably:
+Issue: is this advice still current? When was the last time we saw a review with a speculative polyfill?
```
torgo, line 1981: I think it's current. eg there's a polyfill in development for the fold angle spec https://github.com/diekus/fold-angle/tree/master/polyfill

index.bs
```diff
@@ -2031,20 +2002,39 @@ Some of the existing manifest files include
 * <a href="https://wicg.github.io/origin-policy/">Origin Policy</a> which is used to set security policies.
 
 We encourage people to think of manifest files as extensible.
```
torgo, line 2004: Not sure what this means. Either they're extensible or they're not extensible.

index.bs
```diff
@@ -1976,52 +1976,23 @@ need to be careful when exposing these to the web platform.
...
-Note: By principle, existing manifests use lowercase, underscore-delimited names.
-There have been times where it has been beneficial to re-use dictionaries from a manifest in DOM APIs as well, which are camel-cased.
-One such example is the <a href="https://w3c.github.io/image-resource/">image resource</a>. For this reason, if a
-key can clearly be expressed as a single word, that is recommended.
-
-When designing new keys and values for a manifest, make sure they are needed (that is, they enable well-thought-out use-cases).
-Also, please check if a similar key exists. If an existing key/value pair does more or less what is needed,
-work with the existing spec to extend it to your use-case if possible.
+If your feature requires a manifest,
+investigate whether you can extend an existing manifest schema.
```
torgo, line 1996: Seems like some of the these above points were not captured in the edit. E.g. the thing about fetch timing. Also the list of different kinds of already existent manifest files.

alice: Four major existing manifest files are listed below - do we learn any more from the original text?

Regarding fetch timing - I guess I didn't understand the point it was making, and didn't feel that it added anything to the main point. Do we have an example of where a new manifest type was needed because of fetch timing?

index.bs
torgo: "We encourage people to extend existing manifest files." as proposed replacement text.

index.bs
torgo: @hober can we get your thoughts on this please?

index.bs
alice, line 1981:
```suggestion
```

index.bs
```diff
@@ -1976,52 +1976,23 @@ need to be careful when exposing these to the web platform.
...
-
-It is important not to neglect other aspects of API design such as privacy, security, accessibility, and internationalization.
-Please take advantage of these other excellent resources in your design process:
-
-* <a href="https://w3c.github.io/privacy-considerations/">Privacy Considerations for Web Protocols</a>
-* <a href="https://www.w3.org/TR/security-privacy-questionnaire/">Self-Review Questionnaire: Security and Privacy</a>
-* <a href="https://w3c.github.io/apa/fast/">Web Technology Accessibility Guidelines</a>
-* <a href="https://w3c.github.io/bp-i18n-specdev/">Internationalization Best Practices for Spec Developers</a> <!-- also at https://www.w3.org/TR/international-specs/ -->
-
 <h3 id="polyfills">Polyfills</h3>
```
alice, line 1980:
```suggestion
```

index.bs
alice, line 2004:
```suggestion
We encourage people to extend existing manifest files.
```

index.bs
alice, line 1996:
```suggestion
New web features should be self-contained and self-describing and ideally should not require an additional manifest file.
```

index.bs
```diff
@@ -2030,21 +2001,40 @@ Some of the existing manifest files include
...
-An example of something that was done as a monkey patch that is scheduled to be integrated into the web app manifest in
-a future level (post-CR):
+An example of something that was done as a monkey patch
+that is scheduled to be integrated into the web app manifest in a future level (post-CR):
 
 * https://wicg.github.io/web-share-target/#extension-to-the-web-app-manifest
+</div>
+
+However, if your feature requires a complex set of metadata specific to a functional domain,
+the creation of a new manifest may be justified.
```
alice, line 2029:
```suggestion
the creation of a new manifest may be justified.

You may need to make a new manifest file if the domain of the manifest file is different from the existing manifest files.
For example, if the fetch timing is different, or if the complexity of the manifest warrants it.  Application metadata should be added
to the Web App Manifest or be an extension of it.
Manifests designated to be used for specific applications or which require interoperability with non-browsers may need to take a different approach.
Payment Method Manifest, Publication Manifest, and Origin Policy are examples of these cases.

For example, if you have a single piece of metadata, even if the fetch timing is different than an existing manifest,
it is probably best to use an existing manifest (or ideally design the feature in such a way that a manifest is not required).
However, if your feature requires a complex set of metadata specific to a functional domain, the creation of a new manifest may be justified.
```

index.bs
```diff
@@ -1976,31 +1976,57 @@ need to be careful when exposing these to the web platform.
...
+offers guidance that should be considered in the development of new features, notably:
+
+* Being "polyfillable" isn't essential but <strong>is</strong> beneficial
 * [[#feature-detect]]
 * Polyfill development should be encouraged
 
 <h3 id="extend-manifests">Extend existing manifest files rather than creating new ones</h3>
 
+If your feature requires a manifest,
+investigate whether you can extend an existing manifest schema.
```
alice, line 1995:
```suggestion
investigate whether you can extend an existing manifest schema.

```

torgo (approved):

index.bs
```diff
@@ -1976,31 +1976,57 @@ need to be careful when exposing these to the web platform.
...
+* <a href="https://www.w3.org/TR/appmanifest/">Web App Manifest</a> which contains features related to web applications.
+* <a href="https://www.w3.org/TR/payment-method-manifest/">Payment Method Manifest</a> which is used for payment methods in the context of the web payment API
+* <a href="https://www.w3.org/TR/pub-manifest/">Publication Manifest</a> which is used by some web publications working group standards
+* <a href="https://wicg.github.io/origin-policy/">Origin Policy</a> which is used to set security policies.
+
+We encourage people to extend existing manifest files.
+Always try to get the changes into the original spec,
+or at least discuss the extension with the spec editors.
+Having this discussion is more likely to result in a better design
+and lead to something that better integrates with the platform.
```
alice, line 2008:
```suggestion
and lead to something that better integrates with the platform.

When designing new keys and values for a manifest, make sure they are needed (that is, they enable well-thought-out use-cases).	
Also, please check if a similar key exists. If an existing key/value pair does more or less what is needed,	
work with the existing spec to extend it to your use-case if possible.

```

Issue #172, opened 2020-05-02
add a principle on "no dataloss by default"


dbaron:
As a followup to https://github.com/w3ctag/design-principles/issues/126#issuecomment-610117172 we should probably have a principle (applying to both HTML and CSS) that there's no dataloss by default, as @fantasai [described](http://fantasai.inkedblade.net/weblog/2019/designing-css/#dataloss).

dbaron:
Discussed a bit in breakout A today; should link to minutes later.

dbaron:
I edited the above comment to link to the minutes.

Maybe a better way to describe this is that the default behavior of CSS properties and associated layout systems should always have the content readable/usable by default.  For example, in visual media, the default behavior should not have content clipped, overlapped, or unreachable by scrolling.

dbaron:
So we could perhaps add something like this:

### Content should be readable by default

When you design CSS properties or CSS layout systems (which are typically values of the `display` property), the default behavior should be that content is readable and usable.  For example, the default behavior of a layout system should not lead to content being clipped, content overlapping other content, or content being unreachable by scrolling.  These things should only happen if CSS features are used that are more explicitly choosing such a behavior (for example, `overflow: hidden` or `left: -40em`).  They should not happen by default as a result of something like `display: flex` or `position: relative`.

dbaron:
@atanassov suggested "viewable" rather than "readable", but there's also discussion in #240 about perceivable.

PR #240, opened 2020-10-05
Preserving content visibility by default (#172)

atanassov:
Addresses https://github.com/w3ctag/design-principles/issues/172

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/240.html" title="Last updated on Jan 27, 2021, 1:50 AM UTC (c61c0ff)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/240/a264592...c61c0ff.html" title="Last updated on Jan 27, 2021, 1:50 AM UTC (c61c0ff)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -514,19 +514,6 @@ See also:
 
 This section details design principles for features which are exposed via CSS.
 
-<h3 id="css-content-should-be-visible">Content should be readable by default</h3>
-
-Design CSS properties or CSS layout systems (which are typically values of the `display` property), 
-to preserve the content as readable and usable by default.  
-
-<div class="example">
-For example, the default behavior of a layout system should not lead to content being clipped, 
-content overlapping other content, or content being unreachable by scrolling. 
-These things should only happen if CSS features are used that are more explicitly choosing 
-such a behavior (for example, `overflow: hidden` or `left: -40em`). They should not happen by 
-default as a result of something like `display: flex` or `position: relative`.
-</div>
-
 <h3 id="css-property-separation">Separate CSS properties based on what should cascade separately</h3>
 
 Decide which values should be grouped together as CSS properties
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -514,6 +514,19 @@ See also:
 
 This section details design principles for features which are exposed via CSS.
 
+<h3 id="css-content-should-be-visible">Content should be readable by default</h3>
```
alice, line 517:
```suggestion
<h3 id="css-content-should-be-visible">Content should be perceivable by default</h3>
```

Maybe? Using WCAG language...

index.bs
```diff
@@ -514,6 +514,19 @@ See also:
...
+
+Design CSS properties or CSS layout systems (which are typically values of the `display` property),
```
alice, line 519:
```suggestion
Design CSS properties or CSS layout systems 
```

I'm not sure we need the parenthetical in this summary sentence. I think we can include it a little later on.

index.bs
```diff
@@ -514,6 +514,19 @@ See also:
 
 This section details design principles for features which are exposed via CSS.
 
+<h3 id="css-content-should-be-visible">Content should be readable by default</h3>
+
+Design CSS properties or CSS layout systems (which are typically values of the `display` property), 
+to preserve the content as readable and usable by default.
```
alice, line 520:
```suggestion
to preserve the content as perceivable and usable by default.  
```

Just being consistent. We don't need to go with this, though.

index.bs
```diff
@@ -514,6 +514,19 @@ See also:
 
 This section details design principles for features which are exposed via CSS.
 
+<h3 id="css-content-should-be-visible">Content should be readable by default</h3>
+
+Design CSS properties or CSS layout systems (which are typically values of the `display` property), 
+to preserve the content as readable and usable by default.  
+
+<div class="example">
+For example, the default behavior of a layout system should not lead to content being clipped,
```
alice, line 523: This doesn't seem like an example to me. Do we have specific examples of layout systems which demonstrate this principle?

index.bs
dbaron, line 517: I'd probably put this at the *end* of the CSS section rather than the beginning.

atanassov: Despite 'perceivable' it's not a term used by CSS and proved to be somewhat unfamiliar for non-native speakers. The preferred title ended up being: "Content should be viewable and accessible by default".

index.bs
atanassov: Similar to the previous issue - this should read: "to preserve content as viewable, accessible and usable by default"

index.bs
alice: I can live with that.

index.bs
atanassov: Well, we've kept this principle true in CSSWG over the years despite not having it written out explicitly. I'll change the wording to say that: "For example, the default behavior of all layout systems in CSS will not lead to content being clipped, content overlapping other content, or content being unreachable by scrolling. "

index.bs
atanassov: Done

alice (approved):

PR #244, opened 2020-10-06
Change gh-pages autogeneration from Travis CI to GitHub Actions.

dbaron:
This uses the GitHub Actions configuration documented in
https://github.com/w3ctag/process/blob/master/setting-up-github-repositories.md

I'm hoping that it will lead to easier-to-diagnose errors than the
Travis CI setup.

I haven't tested this.  (Is there a good way to do so other than to just try it?)

Requesting review from @hober, but cc @alice as well so she's aware of it.


hober (approved):
Looks right to me. Let's find out if it works!

PR #256, opened 2020-11-04
Update deploy.yml

plinss:
Skip manifest when updating bikeshed data

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -11,7 +11,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-python@v2
     - run: pip install bikeshed
-    - run: bikeshed update
+    - run: bikeshed update --skip-manifest
     - run: make
     - uses: peaceiris/actions-gh-pages@v3
       with:


alice (approved):

torgo (approved):

dbaron:
I'm wondering whether we'd be able to revert this.  It makes the deploy automation *much* slower.

Do you remember why we had to do this?  Was it to fix a transient issue?

plinss:
We can try. At the time, the manifest data was out of sync (as happens from time to time when Tab's generation system falls over) and the deploy was failing. This way is certainly much slower, but it always gets the correct data.

plinss:
The other approach we can use is to simply curl to the Bikeshed service API on https://api.csswg.org/bikeshed. It always has up to date Bikeshed code and spec data and skips all the local installation entirely.

dbaron:
(We might also try https://github.com/marketplace/actions/spec-prod... although I'm not sure which way it runs bikeshed.)

plinss:
Looks like spec-prod installs a local copy of Bikeshed and uses the manifest update. On the plus side, it also has auto-publishing and validation (and is maintained by someone else).

Issue #70, opened 2017-07-26
Improve guidance for using attributes vs methods


plinss:

dbaron:
Do we remember what this was about?  Is it related to https://github.com/w3ctag/design-principles/issues/46#issuecomment-318063402 ?

plinss:
IIRC this was referencing the issue of having:
```
interface Foo {
    attribute thing bar;
};
```
vs:
```
interface Foo {
    thing getBar();
};
```

Interfaces are inconsistent here and it would be useful to have advice on when to use one vs the other (and a preference for when it doesn't matter for consistency). Issue #46 may be related and help guide the direction.

kenchris:
In the Qt framework, we attempted to use something like attributes for things that were mostly getters, ie not side effects, or had to perform work. So a getBar() method could have side effects, like doing some set up and cache a value, but .bar couldn't. For methods actually doing work, like conversion we would prefix it with to* like toBar(), ie a toHTML()

domenic:
A developer asked me about this question for their API today and I was somewhat surprised that we had not documented it. At least there is an open issue :).

I'll note that, similar to what @kenchris says, this is a well-trodden path in other languages and frameworks. Although we want to be sure that any advice in the TAG design principles conforms to existing web APIs and thus helps enforce consistency, in my experience the web hasn't done anything particularly different from other platforms so far. So, we can likely draw inspiration from existing guidelines.

Here are two guidelines from .NET that I found kind of helpful: [1](https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-4.0/ms229054(v=vs.100)?redirectedfrom=MSDN), [2](https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/bzwdh01d(v=vs.71)?redirectedfrom=MSDN#cpconpropertyusageguidelinesanchor1).

dbaron:
> Here are two guidelines from .NET that I found kind of helpful: [1](https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-4.0/ms229054(v=vs.100)?redirectedfrom=MSDN), [2](https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/bzwdh01d(v=vs.71)?redirectedfrom=MSDN#cpconpropertyusageguidelinesanchor1).

[...]

Issue #195, opened 2020-05-28
Editorial: talk about methods and arguments, not functions and parameters


annevk:
For consistency with IDL.

cynthia:
I'll be the protein sed for this sometime next week.

PR #262, opened 2021-01-05
Added some text for "Attributes vs Methods"

torgo:
Initial commit with Sangwhan's first text, not incorporating Domenic's comments yet.

Original text by @cynthia. 

Referencing https://github.com/w3ctag/design-principles/issues/70

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/262.html" title="Last updated on May 19, 2021, 3:17 PM UTC (046e3f8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/262/f5a116e...046e3f8.html" title="Last updated on May 19, 2021, 3:17 PM UTC (046e3f8)">Diff</a>


annevk:
This text seems somewhat incorrect. A method does not make an expensive operation cheaper. And you can reveal the result of an expensive operation through an attribute as well (if you store it as a state somewhere and the attribute returns a promise).

cynthia:
> This text seems somewhat incorrect. A method does not make an expensive operation cheaper.

I don't think the text implies this, could you point out why it read that way for you?

> if you store it as state somewhere and the attribute returns a promise

If we have concrete examples where this would be preferred over a method that returns a promise I'd be happy to add this in the text.

annevk:
I mean, all the stuff that says "expensive" makes it read that way to me, e.g.,

> If a getter requires performing an expensive operation, it should be a method.

It's not like `offsetTop` would be better as a method.

Example: `ready` on https://w3c.github.io/ServiceWorker/#serviceworkercontainer-interface, iirc.

LeaVerou:
I'm a little skeptical about designing APIs around performance, which is basically letting implementation dictate UI, a common antipattern. An expensive operation today could be a cheap operation tomorrow, but we can't change APIs to match. It may be better to instead guide implementations to cache the result of expensive getters if possible, instead of exposing the difference via syntax.

annevk:
I think performance is important, but an operation doesn't become cheaper if it's a method instead of a getter, that was my point. If something is actually expensive, e.g., because it necessarily crosses a thread or process boundary, it needs to be event/promise-based.

LeaVerou:
> I think performance is important, but an operation doesn't become cheaper if it's a method instead of a getter, that was my point. If something is actually expensive, e.g., because it necessarily crosses a thread or process boundary, it needs to be event/promise-based.

I think the thinking behind the principle is that method calls are seen as potentially expensive, whereas attributes are seen as generally cheap and are not typically cached by calling code.

annevk:
I see, I wonder if there's research on that. There might be some truth to it with folks doing things like `innerHTML += ...` in a loop. 😊

cynthia:
> but an operation doesn't become cheaper if it's a method instead of a getter

as @LeaVerou noted above, that wasn't the implication here. (I figured for the intended audience of this document this bit would be obvious)

What this was trying to suggest is that the user (developer) expects the value to be there for an attribute, while a method could potentially take some time. (This isn't necessarily a web specific thing, but a general perception in programming.)

Let me try to reword this a bit.

> innerHTML += ... in a loop

Yeah, that...

annevk:
I think the problem in that case is the example. `offsetTop` should never have been able to force layout synchronously. `getOffsetTop()` would be equally terrible.

cynthia:
I'm thinking about adding these bits in the text:

- Regardless of it being an attribute or a method, getter-likes should not have side effects.
- A getter that is expected to perform an expensive operation should either be implemented as a method.
- Sometimes it may be desirable to use an attribute which is a Promise instead of a method.

As for the last item - when and why? I have examples but not much to go on the why part. (In particular, why an attribute Promise is desirable over a method which returns a Promise.) Feels like for long-lived objects this wouldn't be a good pattern.

LeaVerou:
I’m not sure about “A getter that is expected to perform an expensive operation should be implemented as a method.” for things that conceptually make sense as getters. Performance could change in the future, but it’s harder to change syntax. This feels like implementation details driving API decisions.

plinss:
I don't see any issues with an attribute that returns a promise, so long as it always returns the *same* promise.

annevk:
(I think it's okay to return a new promise if there was a state change of sorts, but yeah, it shouldn't return a new promise each time you invoke the getter.)

cynthia:
Discussed in today's breakout. I'm going to re-work some of this so that we can land this during the plenary.

plinss (changes requested):
To be in line with #195, this should say 'properties' instead of 'attributes'

cynthia:
>To be in line with #195, this should say 'properties' instead of 'attributes'

Done in https://github.com/w3ctag/design-principles/pull/262/commits/9f62db1581b0dd1d312cb1f8c615bb76c8b55166

index.bs
```diff
@@ -1167,6 +1167,30 @@ which might be expressed as a bitmask in another language,
 use a dictionary object instead.
 This object can be passed around as easily as a single bitmask value.
 
+<h3 id="properties-vs-methods">Properties vs. Methods</h3>
+
+Sometimes it is unclear whether to use an property or a method.
```
plinss, line 1172:
```suggestion
Sometimes it is unclear whether to use a property or a method.
```

plinss:
Actually, I'm re-thinking the attribute vs property thing. We *are* inconsistent throughout the doc, but WebIDL does call them attributes where JS calls them properties.

I suggest we do a careful audit and use the term `attribute` when talking about WebIDL interface definitions, and `property` everywhere else. Or vice-versa, I don't feel strongly, but we should make up a rule and stick with it. Needs more discussion.

We also have to be careful that we don't introduce confusion between CSS properties and JS properties.

index.bs
```diff
@@ -1167,6 +1167,30 @@ which might be expressed as a bitmask in another language,
...
+
+  * Property getters must not have any (observable) side effects.
+    If you have expected side effects, use a method.
+  * Property getters should not perform any expensive operations.
+    If a getter requires performing an expensive operation, it should be a method.
+    (A notable failure of the platform in this regard is getters like offsetTop performing layout; do not repeat this mistake.)
```
LeaVerou, line 1178: Like I wrote, I'm not sure about this point, performance should not influence API shape ideally, since performance concerns are often transient and APIs are much more long-lived. However the last point about `offsetTop` might be covered by the previous point about side effects?

cynthia:
Re: https://github.com/w3ctag/design-principles/pull/262#issuecomment-840214742

I think this should be tracked as a separate issue because the damage has already been done in the main document as well.

https://github.com/w3ctag/design-principles/issues/315

index.bs
```diff
@@ -1167,6 +1167,30 @@ which might be expressed as a bitmask in another language,
...
+Sometimes it is unclear whether to use a property or a method.
+
+  * Property getters must not have any (observable) side effects.
+    If you have expected side effects, use a method.
+  * Property getters should not perform any expensive operations.
+    If a getter requires performing an expensive operation, it should be a method.
+    (A notable failure of the platform in this regard is getters like offsetTop performing layout; do not repeat this mistake.)
+  * If the underlying object has not changed, property getters should return the same object each time it is called.
+    This means <code>obj.property === obj.property</code> must always hold.
+    Returning a new value from an property getter each time is not allowed.
```
plinss, line 1181:
```suggestion
    Returning a new value from a property getter each time is not allowed.
```

index.bs
```diff
@@ -1167,6 +1167,30 @@ which might be expressed as a bitmask in another language,
...
 This object can be passed around as easily as a single bitmask value.
 
+<h3 id="properties-vs-methods">Properties vs. Methods</h3>
+
+Sometimes it is unclear whether to use a property or a method.
+
+  * Property getters must not have any (observable) side effects.
+    If you have expected side effects, use a method.
+  * Property getters should not perform any expensive operations.
+    If a getter requires performing an expensive operation, it should be a method.
```
cynthia, line 1177:
```suggestion
  * Property getters are expected to represent the state of a given object.
    This means a getter should be able to efficiently return a value using the existing state.
    If a getter does not satisfy this, it should be a method.
```

index.bs
cynthia: offsetTop is a combination of both sins.

How does this look? https://github.com/w3ctag/design-principles/pull/262/files#r633975504

LeaVerou: Much better!

plinss (approved):

Issue #156, opened 2020-02-19
Taking debugging use cases into account in W3C specifications


captainbrosset:
The W3C TAG ethical web principles include a section about [the transparency of the web](https://w3ctag.github.io/ethical-web-principles/#transparent) and how _"we will always make sure it is possible to audit and inspect web applications"_.

I'm very thankful for this part of the design principles, and I think some W3C specifications took that at hearth and really went the extra mile in thinking about how a given piece of technology was going to be easy for people to debug.

One example is the Web Animations spec. Right from the starts, it lists [use cases](https://www.w3.org/TR/web-animations-1/#use-cases) including an _"Animation debugging"_ one.
This use case was taken into account right from the start to make sure it would be just as easy to debug animations (when things go wrong) as to create them in the first place:
_"The Web Animations programming interface may be used to inspect running animations"_.

I love this for 2 reasons:

* A new piece of technology is much easier to adopt and use on the long run if it comes with a way to inspect and debug the result of using this technology.
* Requiring (or at least suggesting) UAs to expose some form of debugging information means that web authors are able to create debugging tools themselves. Debugging tools are currently something that mostly browser vendors do. They often require privileged access to the rendering engine of the browser to even exist, and are therefore impossible for web authors to create.

The web was built on View Source but has evolved past it. Being able to access and read the source code is immensely valuable for the web, but it doesn't tell you the full story.
Complex CSS layout implementations in UAs tend to take a lot of decisions on behalf of web authors. Just reading the CSS code for a piece of layout won't be enough to solve a layout bug on a web page.

I would love for specification authors and editors to take debugging as a more important use case, more consistently and for specifications to include suggestions for how UAs should expose debugging information.

Flexbox is a good example of a CSS layout model that people sometimes have problems with and that could use tooling to be easier to adopt and use.
With a few CSS properties, web authors define the sizing rules that flex items should be governed by, but that doesn't tell them what the final size of an item is going to be and why.
Finding out the final size of an item can be done already by accessing the computed style of the element, or DOM APIs like `getBoundingClientRect`, but it won't tell you *why* that happened.

We built [specialized tooling for this in Firefox](https://developer.mozilla.org/en-US/docs/Tools/Page_Inspector/How_to/Examine_Flexbox_layouts) which required us to access privileged data from the layout engine otherwise not accessible to web authors.
[...]

PR #264, opened 2021-01-05
Added subsection for #156.

cynthia:
***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/264.html" title="Last updated on Jan 27, 2021, 1:12 AM UTC (90efc4f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/264/f5a116e...90efc4f.html" title="Last updated on Jan 27, 2021, 1:12 AM UTC (90efc4f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2039,6 +2039,27 @@ which meant converting the names to camel-cased version.
 One such example is the <a href="https://w3c.github.io/image-resource/">image resource</a>.
 For this reason, if a key can clearly be expressed as a single word, that is recommended.
 
+<h3 id="debuggability">Features should be developer-friendly</h3>
+
+Any new feature should be developer-friendly.
+While it is hard to quantify friendliness, at least consider the following points.
+
+Error messages must be meaningful.
+When a developer or user encounters an error,
+the message should be specific to that error case,
+and not overly generic.
+
+Ideally, the error message should have enough information
+to guide the developer in pinpointing where the problem is.
+
+Declarative features such as CSS,
+may require extra work in the implementation for debuggability.
+Defining this in the specification not only makes the feature more developer-friendly,
+it also ensures a consistent development experience for the users.
+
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2039,6 +2039,27 @@ which meant converting the names to camel-cased version.
...
+Any new feature should be developer-friendly.
+While it is hard to quantify friendliness, at least consider the following points.
+
+Error messages must be meaningful.
+When a developer or user encounters an error,
+the message should be specific to that error case,
+and not overly generic.
+
+Ideally, the error message should have enough information
+to guide the developer in pinpointing where the problem is.
```
annevk, line 2053: This is somewhat at odds with how we've been developing errors thus far, especially in networking. We tend to use a generic error. Developer consoles ideally give much more specific information, but that's not suitable to directly expose to the web at large.

cynthia: This is for what message an Exception for example should contain. And as the wording says - "ideally". I think It's fine to use a generic error, but I think the message inside that error should be a bit more specific - otherwise developers will search for that message and possibly end up choosing to opt for a suboptimal solution. (or do a lot of trial and error, also not great.)

annevk: I'm saying that would be problematic as it might expose information about users. We've also had problems with tooling parsing these messages and ending up depending on browser specifics. If anything I'd argue for making things more generic there.

cynthia: I think we might have been talking over each other's heads - @hober pointed this out and the content has been updated in https://github.com/w3ctag/design-principles/pull/264/commits/1a4d5e7411d84d2d42a5c9f2e7a927e36462684c

hober (approved):

index.bs
annevk: Thanks!

Issue #252, opened 2020-10-16
New principle: all new headers shall be structured headers


annevk:
To not expand on the number of custom header parsers in HTTP libraries, all new headers must follow https://tools.ietf.org/html/draft-ietf-httpbis-header-structure. This is already de facto enforced, but it would be good to have it documented.

cc @mnot (feel free to unsubscribe, but I thought you might want to know)

LeaVerou:
If we resolve to work on this, I wonder if this means we should add a new section for HTTP-related principles?

cynthia:
https://w3ctag.github.io/design-principles/#new-http-header-syntax

torgo:
Discussing today and we think we can close this.

PR #287, opened 2021-02-15
HTTP header syntax definition principle

ylafon:
See https://github.com/w3ctag/design-principles/issues/252

***
<a href="https://pr-preview.s3.amazonaws.com/ylafon/design-principles/pull/287.html" title="Last updated on May 11, 2021, 12:32 AM UTC (50416d1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/287/6922690...ylafon:50416d1.html" title="Last updated on May 11, 2021, 12:32 AM UTC (50416d1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2017,7 +2017,7 @@ New MIME types should have a specification and should be registered with the Int
 
 If you really need to define a new HTTP header, it is important that its syntax does not go against [the HTTP specification](https://tools.ietf.org/html/rfc7230#section-3.2).
 
-For more than basic values; If the new header must convey structured data, like lists or dictionnaries, typed values
+For more than basic values; If the new header must convey structured data, like lists or dictionaries, typed values
 like decimals, strings, booleans, then it is recommended to use the syntax defined in [Structured Field Values for HTTP](https://tools.ietf.org/html/rfc8941), to avoid having to write and maintain specific micro-parsers, or even worse
 something that would break those existing parsers. [[RFC8941]]
 
--- a/index.bs
+++ b/index.bs
@@ -2015,11 +2015,25 @@ New MIME types should have a specification and should be registered with the Int
 
 <h3 id="new-http-header-syntax">New HTTP Headers</h3>
 
-If you really need to define a new HTTP header, it is important that its syntax does not go against [the HTTP specification](https://tools.ietf.org/html/rfc7230#section-3.2).
+If you are defining a new HTTP header,
+its syntax mustn't go against
+[the HTTP specification](https://tools.ietf.org/html/rfc7230#section-3.2).
 
-If the new header must convey structured data, such as lists, dictionaries, or typed values
-like decimals, strings, or booleans, then the header should use the syntax defined in [Structured Field Values for HTTP](https://tools.ietf.org/html/rfc8941). This avoids consumers of the header having to write and maintain specific micro-parsers, or even worse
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2013,6 +2013,13 @@ due to security implications, and instead recommend enforcing strict MIME types
 
 New MIME types should have a specification and should be registered with the Internet Assigned Numbers Authority (IANA).
 
+<h3 id="new-http-header-syntax">New HTTP Headers</h3>
+
+If you really need to define a new HTTP header, it is important that its syntax does not go against [the HTTP specification](https://tools.ietf.org/html/rfc7230#section-3.2).
+
+For more than basic values; If the new header must convey structured data, like lists or dictionnaries, typed values
+like decimals, strings, booleans, then it is recommended to use the syntax defined in [Structured Field Values for HTTP](https://tools.ietf.org/html/rfc8941), to avoid having to write and maintain specific micro-parsers, or even worse
+something that would break those existing parsers. [[RFC8941]]
```
annevk, line 2022: I think we should make this stronger. Any new field should/must be a structured field.

index.bs
```diff
@@ -2013,6 +2013,13 @@ due to security implications, and instead recommend enforcing strict MIME types
 
 New MIME types should have a specification and should be registered with the Internet Assigned Numbers Authority (IANA).
 
+<h3 id="new-http-header-syntax">New HTTP Headers</h3>
+
+If you really need to define a new HTTP header, it is important that its syntax does not go against [the HTTP specification](https://tools.ietf.org/html/rfc7230#section-3.2).
+
+For more than basic values; If the new header must convey structured data, like lists or dictionnaries, typed values
```
LeaVerou, line 2020:
```suggestion
For more than basic values; If the new header must convey structured data, like lists or dictionaries, typed values
```

index.bs
```diff
@@ -2013,6 +2013,13 @@ due to security implications, and instead recommend enforcing strict MIME types
 
 New MIME types should have a specification and should be registered with the Internet Assigned Numbers Authority (IANA).
 
+<h3 id="new-http-header-syntax">New HTTP Headers</h3>
+
+If you really need to define a new HTTP header, it is important that its syntax does not go against [the HTTP specification](https://tools.ietf.org/html/rfc7230#section-3.2).
+
+For more than basic values; If the new header must convey structured data, like lists or dictionaries, typed values
+like decimals, strings, booleans, then it is recommended to use the syntax defined in [Structured Field Values for HTTP](https://tools.ietf.org/html/rfc8941), to avoid having to write and maintain specific micro-parsers, or even worse
+something that would break those existing parsers. [[RFC8941]]
```
plinss, line 2022:
```suggestion
If the new header must convey structured data, such as lists, dictionaries, or typed values
like decimals, strings, or booleans, then the header should use the syntax defined in [Structured Field Values for HTTP](https://tools.ietf.org/html/rfc8941). This avoids consumers of the header having to write and maintain specific micro-parsers, or even worse
something that would break those existing parsers. If the new header requires data that can't be represented by Structured Field Values, then either engage with IETF about extending the Structured Field Values syntax, or re-consider if an HTTP header is the right place to expose the data before inventing a new syntax. [[RFC8941]]
```

annevk: What else than structured data can it convey?

plinss: My concern was deeply nested structures that can't currently be represented by SFV.

cynthia (approved):

Issue #117, opened 2018-12-11
High level or low-level APIs?


travisleithead:
## Background

Was watching this [Kranky Geek video](https://www.youtube.com/watch?v=gV354kWh08o&feature=youtu.be&list=PL4_h-ulX5eNfaM0QM5r-PewWaY_zgLH7b&t=1226) (time index 20:00 - 30:00 mins) about WebRTC moving to potentially lower-level APIs with the majority of the "features" of the implementation of the WebRTC internals managed by Web Assembly (some native C app transpiled to WASM). This got me thinking about the general idea of porting native apps to the web using WASM and what it means for the design of Client Side APIs.

WASM programs must still run within the web sandbox--e.g., they are *not* granted special exception to access low-level OS/Driver functionality. From this POV, there is really nothing risky about this direction that web design might take. However, it begs the question of whether the web's current pattern of building relatively high-level APIs is the best approach.

The term "high level" must also be clarified. Over time, esp., with the adoption of the Extensible Web Manifesto, browser vendors have begun focusing on uncovering the "building blocks" that make other higher-level features work. Case-in-point, Web Components aim to explain how high-level markup languages work, enabling web applications to extend or create their own high-level markup languages for distribution on the web. The downside is that the code to power these experiences is not built-in to the browser, and must be downloaded by every client that visits the web site.

Is the approach to WebRTC as shown in the video just the natural evolution of the path we started down with the Extensible Web Manifesto, or is this a development that is beyond simply providing a new set of "low-level" APIs to explain RTC?

Common sense tells me that there should be a happy medium of API design that makes the most sense for the most developers.

This issue is opened to track the discussion around this topic and hopefully land on some recommendations we can provide via the design principles document for those designing APIs.

## Existing Discussion Questions

* If more and more application code is distributed via WASM to the web, how does the user agent continue to mediate the user's interest in terms of protecting privacy, security, etc.. Are the existing capability APIs and their permissions the right design approach?
* Is the web's potential growth limited when "view source" and copy-and-paste code is no longer possible (because WASM isn't really high-level enough to make sense of without source-maps or documentation).
* Is there specific guideance on what low-level APIs would make sense in this context? Currently the design-principles doc suggests text on porting ["high-level" native APIs](https://w3ctag.github.io/design-principles/#wrapper-apis) to the web, but does not cover the reverse of the spectrum.

dbaron:
discussed a bit in Breakout B today (will try to remember to link minutes later)

[...]

PR #291, opened 2021-03-02
Add text for low-level vs high-level APIs (related to #117)

LeaVerou:
This is my first, rough draft of text that addresses #117.

Ideally it needs more examples, but I figured I'd push the text while I brainstorm these.

A lot of this prose is based on [minutes](https://github.com/w3ctag/meetings/blob/gh-pages/2020/telcons/05-04-minutes.md#high-level-or-low-level-apis---hober-alice-dbaron-atanassov) from a past breakout, so I've invited all four participants of that conversation to review, including TAG alumni @alice & @dbaron.

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/291.html" title="Last updated on Jun 21, 2021, 12:11 PM UTC (0f338e9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/291/6922690...0f338e9.html" title="Last updated on Jun 21, 2021, 12:11 PM UTC (0f338e9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -287,7 +287,7 @@ A font picker (high level API)
 over a Font Enumeration API (low level API)
 as it addresses the bulk of use cases,
 without the fingerprinting concerns that accompany
-a general Font Enumeration API.
+a general Font Enumeration API, while preserving accessibility and privacy.
 
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -299,7 +299,7 @@ is not adequate for the use case at hand.
 
 Whenever possible, design high level APIs in terms of
 building blocks over lower level APIs.
-Note that it may not be possible to expose the lower level building blocks.
+Note that it isn't always possible to expose the lower level building blocks as API themselves.
 That may guide decisions on how high level the API needs to be.
 
 A well-layered solution should ensure continuity
--- a/index.bs
+++ b/index.bs
@@ -300,7 +300,15 @@ is not adequate for the use case at hand.
 Whenever possible, design high level APIs in terms of
 building blocks over lower level APIs.
 Note that it isn't always possible to expose the lower level building blocks as API themselves.
-That may guide decisions on how high level the API needs to be.
+
+Lower level building blocks cannot always be exposed as Web APIs.
+A few possible reasons for this are
+to preserve the user's security and privacy,
+or to avoid tying Web APIs to specific hardware implementations.
+However, high level APIs should be designed in terms of
+building blocks over lower level APIs
+whenever possible.
+This may guide decisions on how high level the API needs to be.
 
 A well-layered solution should ensure continuity
 of the ease-of-use vs power tradeoff curve
--- a/index.bs
+++ b/index.bs
@@ -297,10 +297,6 @@ from usage patterns over time.
 They also provide an escape hatch when the higher-level API
 is not adequate for the use case at hand.
 
-Whenever possible, design high level APIs in terms of
-building blocks over lower level APIs.
-Note that it isn't always possible to expose the lower level building blocks as API themselves.
-
 Lower level building blocks cannot always be exposed as Web APIs.
 A few possible reasons for this are
 to preserve the user's security and privacy,
--- a/index.bs
+++ b/index.bs
@@ -292,7 +292,7 @@ a general Font Enumeration API.
 </div>
 
 Low-level APIs afford authors room for experimentation
-so that what the high level APIs can organically emerge
+so that high level APIs can organically emerge
 from usage patterns over time.
 They also provide an escape hatch when the higher-level API
 is not adequate for the use case at hand.
--- a/index.bs
+++ b/index.bs
@@ -286,8 +286,11 @@ A font picker (high level API)
 [was recommended by the TAG](https://github.com/w3ctag/design-reviews/issues/399)
 over a Font Enumeration API (low level API)
 as it addresses the bulk of use cases,
-without the fingerprinting concerns that accompany
-a general Font Enumeration API, while preserving accessibility and privacy.
+while preserving user privacy,
+as it is free from the the fingerprinting concerns that accompany
+a general Font Enumeration API.
+A native font picker also comes with accessibility built-in,
+and provides consistency for end users.
 
 </div>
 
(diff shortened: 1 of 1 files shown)


LeaVerou:
I was a little unsure I fully understand this bit well (which is why it's basically worded the same as it was in the minutes):

> Whenever possible, design high level APIs in terms of building blocks over lower level APIs. Note that it may not be possible to expose the lower level building blocks. That may guide decisions on how high level the API needs to be.

So I reached out to @hober for clarification. The gist is that even when the low level blocks can't be exposed, it's still a good idea to define high level APIs in terms of low level ones. The EWM left some people with the impression that any underlying thing must be exposed as API if any platform feature were to depend on it, and we both agree this is incorrect. 
We also agreed that we should include some guidance on when these underlying things shouldn’t be exposed, otherwise it’s entirely subjective. The low hanging fruit there is security & privacy reasons, but are there others?

index.bs
```diff
@@ -269,6 +269,45 @@ See also:
...
+[was recommended by the TAG](https://github.com/w3ctag/design-reviews/issues/399)
+over a Font Enumeration API (low level API)
+as it addresses the bulk of use cases,
+without the fingerprinting concerns that accompany
+a general Font Enumeration API.
+
+</div>
+
+Low-level APIs afford authors room for experimentation
+so that what the high level APIs can organically emerge
```
hober, line 295:
```suggestion
so that high level APIs can organically emerge
```

index.bs
```diff
@@ -269,6 +269,45 @@ See also:
...
+from usage patterns over time.
+They also provide an escape hatch when the higher-level API
+is not adequate for the use case at hand.
+
+Whenever possible, design high level APIs in terms of
+building blocks over lower level APIs.
+Note that it may not be possible to expose the lower level building blocks.
```
hober, line 302:
```suggestion
Note that it isn't always possible to expose the lower level building blocks as API themselves.
```

index.bs
```diff
@@ -269,6 +269,45 @@ See also:
...
+</div>
+
+Low-level APIs afford authors room for experimentation
+so that what the high level APIs can organically emerge
+from usage patterns over time.
+They also provide an escape hatch when the higher-level API
+is not adequate for the use case at hand.
+
+Whenever possible, design high level APIs in terms of
+building blocks over lower level APIs.
```
hober, line 301: I think maybe this sentence should be moved after the next one? So that the reader knows *why* this is the case?

hober (approved):
Looks good to me. I noted several nits which may or may not be worth adopting.

index.bs
LeaVerou: Just pushed a change (somewhat larger than what you were suggesting), thoughts?

LeaVerou:
@hober any thoughts on my changes in response to your feedback? Let's try to get this merged soon 😄

index.bs
```diff
@@ -269,6 +269,49 @@ See also:
...
+However, high level APIs should be designed in terms of
+building blocks over lower level APIs
+whenever possible.
+This may guide decisions on how high level the API needs to be.
+
+A well-layered solution should ensure continuity
+of the ease-of-use vs power tradeoff curve
+and avoid sharp cliffs
+where a small amount of incremental use case complexity
+results in a disproportionately large increase of code complexity.
```
cynthia, line 313: I might remove the word "disproportionately" - "large" conveys the message well enough, and the word isn't friendly towards non-native speakers.

index.bs
```diff
@@ -269,6 +269,49 @@ See also:
...
+intervene in various ways [on behalf of the user](#priority-of-constituencies),
+such as to ensure accessibility, privacy, or usability.
+
+<div class=example id=font-enumeration>
+
+A font picker (high level API)
+[was recommended by the TAG](https://github.com/w3ctag/design-reviews/issues/399)
+over a Font Enumeration API (low level API)
+as it addresses the bulk of use cases,
+without the fingerprinting concerns that accompany
```
atanassov, line 289: I like the example but your leading point to it was about how good (high level API) are as they provide accessibility, privacy etc. You should add a sentence about Font Picker being accessible, privacy preserving etc.

index.bs
```diff
@@ -269,6 +269,49 @@ See also:
...
+building blocks over lower level APIs
+whenever possible.
+This may guide decisions on how high level the API needs to be.
+
+A well-layered solution should ensure continuity
+of the ease-of-use vs power tradeoff curve
+and avoid sharp cliffs
+where a small amount of incremental use case complexity
+results in a disproportionately large increase of code complexity.
+
```
atanassov, line 314: After reading the entire section few times I find myself still unsure about what the guidance is. My recollection from our discussions was that generally we recommend a layered approach to building API and starting by exposing the highest level possible. That allows UAs to take care of all the things you mentioned about a11y, privacy etc., validate scope and functionality and still have a path forward to allow extending by adding lower levels of APIs. It is also much easier to remove few APIs (high level presumably) vs many API.

LeaVerou:
After @atanassov's comments, I wonder if this should actually be merged with https://github.com/w3ctag/design-principles/pull/306 which is basically about the same thing but provides more clear guidance.

index.bs
```diff
@@ -269,6 +269,49 @@ See also:
...
+such as to ensure accessibility, privacy, or usability.
+
+<div class=example id=font-enumeration>
+
+A font picker (high level API)
+[was recommended by the TAG](https://github.com/w3ctag/design-reviews/issues/399)
+over a Font Enumeration API (low level API)
+as it addresses the bulk of use cases,
+without the fingerprinting concerns that accompany
+a general Font Enumeration API.
```
cynthia, line 290:
```suggestion
a general Font Enumeration API, while preserving accessibility and privacy.
```

index.bs
LeaVerou: There was already a mention of accessibility and privacy that right under your comment, but I reworded to emphasize. What do you think?

index.bs
torgo: @atanassov in discussion this morning we feel the best approach is to land the PR as written and then address the issues you've raised in a subsequent PR.

Issue #50, opened 2017-02-22
give advice on read-only lists


dbaron:
In a comment on #47, which created the [section on live versus static objects](https://w3ctag.github.io/design-principles/#live-vs-static), @domenic wrote:
> Maybe not as part of this PR or even this section, but a common point of confusion is the case of "read only lists" (where only the implementation can change them). Currently we tend to recommend people return static snapshots as arrays (for methods, when they change frequently) or frozen arrays (for attributes, when they change infrequently). It's not the best solution, compared to some kind of real read-only array type, but we tend to be going in that direction. Notably we are moving away from creating various fake-array types like NodeList or StyleSheetList for this purpose.

torgo:
We agreed we should keep this open but lower priority.

hober:
This continues to be something that people run into; see WICG/construct-stylesheets#45 for instance.

dbaron:
So it's not clear to me what we need to recommend about `sequence<T>` versus `FrozenArray<T>` versus `ObservableArray<T>`... or for that matter how stable and widely adopted the latter two are.

It also seems that the notion of returning static snapshots that the initial comment here suggests was the *problem* in WICG/construct-stylesheets#45 because that case required an API for mutation, so I think we have at least some evidence that static snapshots *aren't* a good API for mutation.  I think this aligns with [the existing advice on live versus static objects](https://w3ctag.github.io/design-principles/#live-vs-static) which I think is relevant here (particularly the bit about how things that are the way state is mutated should be live).

Is the intent of `ObservableArray<T>` that it can replace all of these custom array classes in the case where mutation *is* needed?  That's what it looks like from a quick skim.

marcoscaceres:
Sorry if this is covered already (it's kinda hinted at in the OP)... there is also cases I've seen where `FrozenArray<T>` have been specified to be passed as method arguments:

```
  // PaymentDelegation is just a dictionary or whatever
  Promise<void> enableDelegations(FrozenArray<PaymentDelegation> delegations);
```

What's the right thing to use above? A `FrozenArray` or `sequence`?

domenic:
Always sequence for method arguments.

dbaron:
There appears to be a good bit of useful background for this in heycam/webidl#796.

atanassov:
Upon further discussion with @dbaron as part of our Cork virtual f2f the following recommendation should be considered and made.
What are the recommended list types used for:
1. Function parameters - from comment above it should be `sequence<T>`
[...]

PR #292, opened 2021-03-03
Add advice about read-only lists

atanassov:
The added recommendation is about the types of static lists to be considered for API design.
Addresses https://github.com/w3ctag/design-principles/issues/50

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/292.html" title="Last updated on May 13, 2021, 12:32 AM UTC (a0b49a5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/292/e5aafa3...a0b49a5.html" title="Last updated on May 13, 2021, 12:32 AM UTC (a0b49a5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -989,6 +989,13 @@ converting {{undefined}} to the type of the parameter.
 For example, if a boolean parameter isn't set,
 it must default to false.
 
+When deciding between different list data types for your API, 
+and if possible, consider using the following static list types:
+
+* Method list parameters should be of type [=sequence types|sequence&ltT&gt=]
+* Method return values should be of type [=sequence types|sequence&ltT&gt=]
+* Attributes should be of type [=observable array type|ObservableArray&ltT&gt=]
+
 See also:
 
 * [[#prefer-dict-to-bool]]
--- a/index.bs
+++ b/index.bs
@@ -990,7 +990,7 @@ For example, if a boolean parameter isn't set,
 it must default to false.
 
 When deciding between different list data types for your API, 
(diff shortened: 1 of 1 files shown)


LeaVerou:
What happened to ReadOnlyArray? Was it folded into ObservableArray as a special case?

cynthia:
@plinss and I discussed this in a breakout during the F2F. We think the ObservableArray recommendation is a bit strange - the observable properties are definitely not desirable by every case where an attribute is needed, and there are cases where read-only is desirable (e.g. in an attribute) - but this PR doesn't seem to cover that.

Also, @plinss mentioned he wasn't so sure about the return being restricted to only sequences. There are cases where other array types are more desirable, but this feels a bit restrictive.

annevk:
@LeaVerou I'm not aware of a `ReadOnlyArray` proposal, but https://heycam.github.io/webidl/#idl-observable-array does indeed satisfy that use case (it's basically up to API designers to decide what kind of operations they permit).

@domenic was it considered to have default implementations of delete and set an indexed value so that the readonly case would be consistent across APIs? That does strike me as something that is missing, but I might also be missing something. 😊

plinss (approved):

plinss (approved):

atanassov:
Had another round of CR with @plinss and @cynthia and arrived with the final wordsmithing addressing all concerns for this PR. Thank you all for the time.

cynthia (approved):

PR #296, opened 2021-03-17
Fix Bikeshed syntax in casing rules table

domenic:


domenic:
Hmm it looks like `autocompleteerror` was removed from HTML ages ago...

annevk:
Maybe replace with `canplaythrough` or `securitypolicyviolation`.

cynthia (approved):

PR #306, opened 2021-04-16
Incorporate Alan Kay's maxim into 'Prefer simple solutions'.

### LINKED ISSUE #299: Simple things should be easy, complex things should be possible
From [this week's plenary](https://github.com/w3ctag/meetings/blob/gh-pages/2021/telcons/03-29-minutes.md#high-level-vs-low-level)

> Lea: I think a good API isn't one point in the continuum of low level and high level, it progressively reveals more abilities as you need them. Can start very hgih level, sensible defaults, simple things are easy, and progressively you can customise what it does for complex use cases. That is the ideal. Simple things are easy and complex things are possible.
> 
> Dan: I like that wording. Maybe we do need a simple things should be easy principle.

Also related to #117 

Actually, maybe it should be "Common things should be easy, complex things should be possible".
The common case is not necessarily simple (e.g. see Sanitizer API), but that's the one we should optimize for.

hober:
Fixes #299.

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/306.html" title="Last updated on Apr 16, 2021, 9:22 PM UTC (ebffc6f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/306/98bb0d4...ebffc6f.html" title="Last updated on Apr 16, 2021, 9:22 PM UTC (ebffc6f)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/306.html" title="Last updated on Jan 10, 2022, 5:45 PM UTC (900accb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/306/98bb0d4...900accb.html" title="Last updated on Jan 10, 2022, 5:45 PM UTC (900accb)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -278,6 +278,16 @@ though their solutions may be more complicated to use.
 As Alan Kay said,
 "simple things should be simple, complex things should be possible."
 
+Do note however that while common cases are often simple,
+commonality and complexity are not always correlated.
+
+<div class=example>
+The Sanitizer API provides a shortcut for the common case of preventing XSS,
+even though the rules necessary for that are not simple at all
+and requires extensive security knowledge. 
+It also permits simpler types of filtering, but with more configuration.
+</div>
+
 See also:
 
 * [[LEAST-POWER]]
--- a/index.bs
+++ b/index.bs
@@ -284,7 +284,7 @@ commonality and complexity are not always correlated.
 <div class=example>
 Sanitizing HTML to prevent XSS attacks is a complex process
 that requires extensive security knowledge,
-however the Sanitizer API provides a shortcut for this common use case. 
+however the [Sanitizer API](https://wicg.github.io/sanitizer-api/) provides a shortcut for this common use case. 
 It also permits simpler types of filtering, but with more configuration.
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -282,7 +282,9 @@ Do note however that while common cases are often simple,
 commonality and complexity are not always correlated.
 
 <div class=example>
-Sanitizing HTML to prevent XSS attacks is a complex process that requires extensive security knowledge, however the Sanitizer API provides a shortcut for this common use case. 
+Sanitizing HTML to prevent XSS attacks is a complex process
+that requires extensive security knowledge,
+however the Sanitizer API provides a shortcut for this common use case. 
 It also permits simpler types of filtering, but with more configuration.
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -265,11 +265,18 @@ although they may be harder to find.
 Simpler features are easier for user agents to implement and test,
 more likely to be interoperable,
 and easier for authors to understand.
+If you expect authors will use your feature often,
+it's especially important to make it easy to use.
 
 Make sure that your <a href="#priority-of-constituencies">user needs</a>
 are well-defined.
 This allows you to avoid scope creep,
 and make sure that your API does actually meet the needs of all users.
+Of course,
+complex or rare use cases are also worth solving,
+though their solutions may be more complicated to use.
+As Alan Kay said,
+"simple things should be simple, complex things should be possible."
 
 See also:
 
--- a/index.bs
+++ b/index.bs
@@ -282,9 +282,7 @@ Do note however that while common cases are often simple,
 commonality and complexity are not always correlated.
 
 <div class=example>
-The Sanitizer API provides a shortcut for the common case of preventing XSS,
-even though the rules necessary for that are not simple at all
-and requires extensive security knowledge. 
+Sanitizing HTML to prevent XSS attacks is a complex process that requires extensive security knowledge, however the Sanitizer API provides a shortcut for this common use case. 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -265,11 +265,18 @@ although they may be harder to find.
 Simpler features are easier for user agents to implement and test,
 more likely to be interoperable,
 and easier for authors to understand.
+If you expect authors will use your feature often,
+it's especially important to make it easy to use.
```
LeaVerou, line 269: I think we should recommend that each API should try to make its common cases easy to use regardless of how frequently the API is used in general. It's much easier for an API designer to know which use cases of their API are common, than to know how frequently an API is used compared to other APIs. 
That said, of course if API designers actually _know_ that their feature will be used often in general they should strive even more to make common cases easy, but like I said, that's often hard to gauge.

index.bs
```diff
@@ -265,11 +265,18 @@ although they may be harder to find.
...
 
 Make sure that your <a href="#priority-of-constituencies">user needs</a>
 are well-defined.
 This allows you to avoid scope creep,
 and make sure that your API does actually meet the needs of all users.
+Of course,
+complex or rare use cases are also worth solving,
+though their solutions may be more complicated to use.
+As Alan Kay said,
+"simple things should be simple, complex things should be possible."
```
LeaVerou, line 279: It's generally good to cite pre-existing sources to further ground our principles, and this is probably the most well-known quote on this. However, as I mentioned in the issue, I think the guidance should be that **common cases** should be easy, which is not necessarily the same as simple cases. E.g. think of the Sanitizer API we recently reviewed. The common case is not simple at all, it requires extensive security knowledge to take care of all the different rules to prevent XSS. However, *that* is the case to optimize for, not for simple types of filtering that are much less useful.

torgo:
We agreed that since we merged #291 we need to reword this before merging it.

index.bs
LeaVerou, line 269:
```suggestion
It is especially important to design your feature so that 
the most common use cases are easy to accomplish.
```

index.bs
LeaVerou, line 279:
```suggestion
"simple things should be simple, complex things should be possible."

Do note however that while the common cases are often simple,
commonality and complexity are not always correlated.

<div class=example>
The Sanitizer API provides a shortcut for the common case of preventing XSS,
even though the rules necessary for that are not simple at all
and requires extensive security knowledge. 
It also permits simpler types of filtering, but with more configuration.
</div>
```

index.bs
```diff
@@ -265,11 +265,28 @@ although they may be harder to find.
...
+
+Do note however that while common cases are often simple,
+commonality and complexity are not always correlated.
+
+<div class=example>
+The Sanitizer API provides a shortcut for the common case of preventing XSS,
+even though the rules necessary for that are not simple at all
+and requires extensive security knowledge.
```
plinss, line 287:
```suggestion
Sanitizing HTML to prevent XSS attacks is a complex process that requires extensive security knowledge, however the Sanitizer API provides a shortcut for this common use case. 
```

index.bs
```diff
@@ -265,11 +265,26 @@ although they may be harder to find.
...
+complex or rare use cases are also worth solving,
+though their solutions may be more complicated to use.
+As Alan Kay said,
+"simple things should be simple, complex things should be possible."
+
+Do note however that while common cases are often simple,
+commonality and complexity are not always correlated.
+
+<div class=example>
+Sanitizing HTML to prevent XSS attacks is a complex process that requires extensive security knowledge, however the Sanitizer API provides a shortcut for this common use case.
```
plinss, line 285:
```suggestion
Sanitizing HTML to prevent XSS attacks is a complex process
that requires extensive security knowledge,
however the Sanitizer API provides a shortcut for this common use case. 
```

index.bs
```diff
@@ -265,11 +265,28 @@ although they may be harder to find.
...
+complex or rare use cases are also worth solving,
+though their solutions may be more complicated to use.
+As Alan Kay said,
+"simple things should be simple, complex things should be possible."
+
+Do note however that while common cases are often simple,
+commonality and complexity are not always correlated.
+
+<div class=example>
+Sanitizing HTML to prevent XSS attacks is a complex process
```
torgo, line 285:
```suggestion
[Sanitizing HTML](https://wicg.github.io/sanitizer-api/) to prevent XSS attacks is a complex process
```

index.bs
```diff
@@ -265,11 +265,28 @@ although they may be harder to find.
...
+that requires extensive security knowledge,
+however the Sanitizer API provides a shortcut for this common use case.
```
torgo, line 287:
```suggestion
however the [Sanitizer API](https://wicg.github.io/sanitizer-api/) provides a shortcut for this common use case. 
```

torgo (approved):
lgtm

plinss (approved):

Issue #185, opened 2020-05-13
Security / Cryptography Principle?


torgo:

torgo:
Came out of webcrypto review... We discussed briefly on the call October 5 2020. We need a very carefully worded principle here that can basically say "new crypto algorithms that are proposed to be added to the web platform should be rigorously review by international security experts..."

torgo:
We discussed on 4-1-2020 in breakout the idea of adding a general section on crypto in the web - with some discussion of desirable traits of crypto algorithms and APIs...

torgo:
Related info https://www.chromestatus.com/feature/4913922408710144

PR #310, opened 2021-05-11
Add a new principle for crypto.

cynthia:
We didn't have any mention of crypto, so here it goes: #185.

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/310.html" title="Last updated on Sep 15, 2021, 8:05 AM UTC (e4b1c08)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/310/2f553a7...e4b1c08.html" title="Last updated on Sep 15, 2021, 8:05 AM UTC (e4b1c08)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2130,16 +2130,10 @@ part of the specification is [Web Animations](https://www.w3.org/TR/web-animatio
 <h3 id="crypto">Use the best crypto, and expect it to evolve</h3>
 
 Use only cryptographic algorithms
-that have been impartially reviewed by security experts,
+that have been impartially reviewed by security experts,
 and make sure your choice of algorithm is proven, and up-to-date.
-
 Not only do they become obsolete or insecure,
 cryptographic protocols and algorithms also evolve quickly.
-Therefore, features should be designed
-so that they are usable in an algorithm-independent manner.
-For example, [WebCrypto](https://www.w3.org/TR/WebCryptoAPI/#scope-algorithms)
-is designed so that the underlying algorithm
-is a parameter to a generic API.
 
 <h2 id="spec-writing">Writing good specifications</h2>
 
--- a/index.bs
+++ b/index.bs
@@ -2127,6 +2127,20 @@ it also ensures a consistent development experience for the users.
 A good example where debuggability was defined as
 part of the specification is [Web Animations](https://www.w3.org/TR/web-animations-1/#use-cases).
 
+<h3 id="crypto">Use the best crypto, and expect it to evolve</h3>
+
+Use only cryptographic algorithms
+that have been impartially reviewed by security experts,
+and make sure your choice of algorithm is proven, and up-to-date.
+
+Not only do they become obsolete or insecure,
+cryptographic protocols and algorithms also evolve quickly.
+Therefore, features should be designed
+so that they are usable in an algorithm-independent manner.
+For example, [WebCrypto](https://www.w3.org/TR/WebCryptoAPI/#scope-algorithms)
+is designed so that the underlying algorithm
+is a parameter to a generic API.
+
 <h2 id="spec-writing">Writing good specifications</h2>
 
 This document mostly covers API design for the Web,


annevk:
Drive-by: might be good for @sleevi or @mozkeeler to look at this before it lands.

index.bs
```diff
@@ -2127,6 +2127,20 @@ it also ensures a consistent development experience for the users.
...
+that have been impartially reviewed by security experts,
+and make sure your choice of algorithm is proven, and up-to-date.
+
+Not only do they become obsolete or insecure,
+cryptographic protocols and algorithms also evolve quickly.
+Therefore, features should be designed
+so that they are usable in an algorithm-independent manner.
+For example, [WebCrypto](https://www.w3.org/TR/WebCryptoAPI/#scope-algorithms)
+is designed so that the underlying algorithm
+is a parameter to a generic API.
```
sleevi, line 2142: So I know this is meant well, but it can have a few different interpretations, some of which remain deeply debated within the cryptographic community.

I wouldn't necessarily call WebCrypto a hallmark of good design; its generic approach certainly confounds both polyfillability and feature detection in very unfortunate ways. The original design was something closer to Apple's Security Transforms API, except that Streams was vastly more immature than Promises (at the time), and much less reliable. The present incarnation bears similarity to Windows' CryptoAPI/CNG or PKCS#11, but that's not without known compromises - such APIs (of grouping algorithms into "shapes") tend to be less adaptable to evolving cryptography (PKCS#11 has shown this greatly). In particular, the current design of WebCrypto fundamentally makes it problematic for performance, which was what the original Streams'-based API tried for.

However, setting aside those design-level questions (that is, do you make square holes [knowing how they'll be (mis)used](https://www.tiktok.com/@tired_actor/video/6958183750198938885)), there's also the question about implementation, and the question of "cryptographic agility"

That is, one interpretation of this advice would be to assume that web features should support the "pluggable encryption" version of cryptographic agility, in which the developer can mix and match primitives, or algorithms, to their hearts content, negotiating with their peer if needed. This is the model that JW{K/T/S} took, to [predictably disastrous results](https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/). Similarly, as part of TLS 1.3 effort, the idea of "mix and match crypto" was explicitly phased out, in favor of holistic combinations of algorithms that captured all the necessary security state. Further, it explicitly limited the options to a well-tested set, and even then, it was and is not without controversy. Equally still, the [TLS space](https://ssl-config.mozilla.org/) is [notably complex](https://github.com/ssllabs/research/wiki/SSL-Server-Rating-Guide) and hard for developers to get right.

Many secure protocols, such as Signal, are secure explicitly because they do not allow cryptographic agility/negotiation: they treat the cryptographic properties as fundamental and intrinsically part of the "version" of the protocol, such that if you want to change algorithms, you're really negotiating a new version of the protocol.

While hardly scientific, [this poll](https://twitter.com/FiloSottile/status/1386751406758105089) from Golang's cryptography lead perhaps best captures some of the various complexities here within the cryptographic community.

As such, the language is unclear what, precisely, is meant by this. For example, is version negotiation sufficient? Are we suggesting opinionated APIs (that remove developer choice and, say, explicitly require a particular algorithm) are less desirable than abstract APIs that make the complex balancing the developer's fault?

Ultimately, this principle feels very out of place in this document, given how little is spent on security and protocols, of which this is ostensibly a small subset. It's unclear if this is talking about API design, protocols used by features, or something else, and it's difficult to succinctly distill very actionable advice, since the trade-offs are complicated even for the "simple" things.

However, as the person responsible for WebCrypto's evolving shapes, all of which felt (and still feel) awful from a developer experience perspective, despite repeated efforts to engage browsers' more versed design guides, it does feel wrong to point to that as a good model :)

plinss, line 2142: Another s/paramater/argument/

index.bs
```diff
@@ -2127,6 +2127,20 @@ it also ensures a consistent development experience for the users.
...
+Use only cryptographic algorithms
+that have been impartially reviewed by security experts,
+and make sure your choice of algorithm is proven, and up-to-date.
+
+Not only do they become obsolete or insecure,
+cryptographic protocols and algorithms also evolve quickly.
+Therefore, features should be designed
+so that they are usable in an algorithm-independent manner.
+For example, [WebCrypto](https://www.w3.org/TR/WebCryptoAPI/#scope-algorithms)
+is designed so that the underlying algorithm
```
rhiaro, line 2141:
```suggestion
<h3 id="crypto">Use the best crypto, and expect it to evolve</h3>

Use only cryptographic algorithms
that have been impartially reviewed by security experts,
and make sure your choice of algorithm is proven, and up-to-date.
Not only do they become obsolete or insecure,
cryptographic protocols and algorithms also evolve quickly.
```

Suggestion to remove the part that @sleevi raised concerns with (https://github.com/w3ctag/design-principles/pull/310/files#r630611213) but keep the main essence of the principle.

torgo (approved):
lgtm

cynthia:
New text LGTM.

PR #313, opened 2021-05-13
Add consistency section, closes #285

### LINKED ISSUE #285: New design principle: Consistency
Consistency is currently mentioned in [Naming](https://w3ctag.github.io/design-principles/#naming-consistency), but I think it's broader than naming and should really have its own section, which naming could reference.

A few thoughts:

I think we probably all agree that all else being equal, consistency with the rest of the Web Platform should be prioritized. However since typically all else are *not* equal, there are tradeoffs to be made, and we should provide guidance about how to resolve these dilemmas.

* Since the Web Platform has gradually evolved over time, there are often multiple conflicting precedents. We should provide some guidance about which precedent to choose in these cases. It seems that a reasonable guideline might be to choose based on prevalence, API ergonomics, and to prioritize newer APIs over older ones. Not sure about the relative order of these three factors.
* There is often a tension between API ergonomics and internal consistency, when precedent is very clearly of poor usability. In some cases it makes sense to break consistency to improve usability, but the improvement in usability should be very significant to justify this.
* There is often a tension between [internal and external consistency](https://www.nngroup.com/articles/consistency-and-standards/#articleBody:~:text=A%20Continuum%20of%20Consistency). That might materialize in three layers: consistency within the technology (e.g. CSS) vs consistency with the rest of the Web Platform vs consistency with the rest of the world. One example is Lab colors: It would be more consistent with CSS to use `<percentage>` for L (0%-100%), but more consistent with the color science world to use a `<number>` (0-100). There was a lot of heated debate, which [resolved](https://drafts.csswg.org/css-color-4/#specifying-lab-lch) in favor of consistency within CSS, i.e. percentages.

This is partly related to the [Pave the Cowpaths HTML Design principle](https://www.w3.org/TR/html-design-principles/#pave-the-cowpaths). Do we want to merge them?

LeaVerou:
As discussed in 18b.

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/313.html" title="Last updated on May 13, 2021, 7:13 PM UTC (65ee986)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/313/1c64d38...65ee986.html" title="Last updated on May 13, 2021, 7:13 PM UTC (65ee986)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -283,6 +283,53 @@ Naming APIs well makes it much easier for authors to use them correctly.
 See the more detailed <a href="#naming-is-hard">Naming principles</a> section
 for specific guidance on naming.
 
+<h3 id="consistency">Consistency</h3>
+
+It is good practice to consider precedent in the design of your API
+and to try to be consistent with it.
+
+There is often a tension between API ergonomics and consistency,
+when existing precedent is of poor usability.
+In some cases it makes sense to break consistency to improve usability,
+but the improvement should be very significant to justify this.
+
+<!-- TODO: example? -->
+
+Since the Web Platform has gradually evolved over time,
+there are often multiple conflicting precedents
+which are mutually exclusive.
+You can weigh which precdent to follow by taking into account
+prevalence (all else being equal, follow the more popular precedent),
+API ergonomics (all else being equal, follow the more usable precedent),
+and API age (all else being equal, follow the newer precedent).
+
+<!-- TODO: example? -->
+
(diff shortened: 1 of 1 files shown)


torgo (approved):
It looks good to me as is. I wonder if we should also add an example about WebXR or WebRTC or some similar API where there is an existing "outside domain" / community of practice that has established naming conventions.

PR #317, opened 2021-05-19
Add non-"fully active" document handling guide

rakina:
As discussed in https://github.com/w3ctag/design-reviews/issues/628, we want to improve support for BFCache/non-fully active documents, as we noticed that a lot of API specifications do not handle it well/explicitly.

This change adds various guides and things to be aware of related to non-"fully active" documents, including gives various examples. We hope this will help future APIs to handle BFCache/on-"fully active" documents correctly by default.

cc @hober @domenic @fergal

Note that there's another PR for the "Security & Privacy Questionnaire" doc at https://github.com/w3ctag/security-questionnaire/pull/128

***
<a href="https://pr-preview.s3.amazonaws.com/rakina/design-principles/pull/317.html" title="Last updated on Aug 16, 2021, 10:03 AM UTC (eadf0d9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/317/1af4f0b...rakina:eadf0d9.html" title="Last updated on Aug 16, 2021, 10:03 AM UTC (eadf0d9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -59,6 +59,10 @@ urlPrefix: https://url.spec.whatwg.org/; spec: URL
     url: #percent-encode; type: abstract-op; text: percent-encode
 urlPrefix: https://www.w3.org/TR/payment-request/; spec: payment-request
     url: #dfn-state; type: dfn; text: [[state]];
+urlPrefix: https://w3c.github.io/geolocation-api/; spec: Geolocation
+    url: #watchposition-method; type: interface; for: Geolocation; text: watchPosition()
+urlPrefix: https://w3c.github.io/gamepad/; spec: Gamepad
+    url: #event-gamepadconnected; type: event; text: gamepadconnected
 </pre>
 
 <style>
@@ -525,49 +529,51 @@ See also:
 <h3 id="support-non-fully-active">Support non-"fully active" documents</h3>
 
 After a user navigated away from a document,
-the document might stay around in a non-"[=Document/fully active=]" state,
+the document might stay around in a non-[=Document/fully active=] state,
 and might be reused when the user navigates back to the entry holding the document.
-In the past, many APIs have missed specifying support for non-"[=Document/fully active=]" documents,
+In the past, many APIs have missed specifying support for non-[=Document/fully active=] documents,
 making them hard to support in various user agents' implementation of "back/forward cache" (BFCache).
 
 To avoid this happening with your API,
-you should specify support for non-"[=Document/fully active=]" documents by following these guidelines:
+you should specify support for non-[=Document/fully active=] documents by following these guidelines:
 
-<h4 id="gate-fully-active">Gate actions with "[=Document/fully active=]" checks</h4>
+<h4 id="gate-fully-active">Gate actions with [=Document/fully active=] checks</h4>
 
 When performing actions that might update the state of a document,
-be aware that the document might not be "[=Document/fully active=]"
+be aware that the document might not be [=Document/fully active=]
 and is considered as "non-existent" from the user's perspective.
 This means they should not receive updates or perform actions.
 
 In many cases,
-anything that happens while the document is not "[=Document/fully active=]"
+anything that happens while the document is not [=Document/fully active=]
 should be treated as if it never happened,
 and not "queue" the updates and send them after the document becomes fully activated.
 If it makes sense to send the latest state update (and only the latest state,
-not a queue of state changes) to the document after it becomes "[=Document/fully active=]" again,
+not a queue of state changes) to the document after it becomes [=Document/fully active=] again,
 consider the "listen for changes" pattern below.
 <div class="example">
   APIs that periodically send information updates,
-  such as Geolocation API's "[watch position](https://w3c.github.io/geolocation-api/#watchposition-method)"
+  such as Geolocation API's {{Geolocation/watchPosition()}}
   should not send updates if the document is no longer fully active.
-  It also should not queue those updates to arrive later,
-  and only resume sending updates when the document becomes active with the latest location.
+  They also should not queue those updates to arrive later,
+  and only resume sending updates when the document becomes active
+  with the latest information.
 </div>
 
-<h4 id="listen-fully-active">Listen for changes to "[=Document/fully active=]" status</h4>
+<h4 id="listen-fully-active">Listen for changes to [=Document/fully active=] status</h4>
 
-When a document goes from "[=Document/fully active=]" to non-"[=Document/fully active=]",
+When a document goes from [=Document/fully active=] to non-[=Document/fully active=],
 it should be treated similarly to the way discarded documents are treated.
 The document must not retain exclusive access to shared resources
 and must ensure that no new requests are issued
 and that connections that allow for new incoming requests are terminated. 
-When a document goes from non-"[=Document/fully active=]" to "[=Document/fully active=]" again,
+When a document goes from non-[=Document/fully active=] to [=Document/fully active=] again,
 it can restore connections if appropriate.
   
 While web authors can manually do cleanup (e.g. release the resources, sever connections)
 from within the {{pagehide}} event and restore them from the {{pageshow}} event themselves,
-doing this automatically from the API design would reduce a lot of overhead.
+doing this automatically from the API design allows the document to be kept alive after navigation by default,
+and is more likely to lead to well-functioning web applications.
 
 <div class="example">
   APIs that create live connections can pause/close the connection and possibly resume/reopen it later.
@@ -578,22 +584,22 @@ doing this automatically from the API design would reduce a lot of overhead.
 <div class="example">
   APIs that hold non-exclusive resources
   may be able release the resource when the document becomes not fully active,
-  and re-acquire them when it becomes "[=Document/fully active=]" again (Screen Wake Lock API is already [doing](https://w3c.github.io/screen-wake-lock/#handling-document-loss-of-full-activity) the first part).
+  and re-acquire them when it becomes [=Document/fully active=] again (Screen Wake Lock API is already [doing](https://w3c.github.io/screen-wake-lock/#handling-document-loss-of-full-activity) the first part).
 </div>
 Note: this might not be appropriate for all types of resources,
 e.g. if an exclusive lock is held,
-we cannot just release it and reacquire when "[=Document/fully active=]"
+we cannot just release it and reacquire when [=Document/fully active=]
 since another page could then take that lock.
 If there is an API to signal to the page that this has happened,
 it may be acceptable but beware that if the only time this happens is with BFCache,
 then it's likely many pages are not prepared for it. If it is not possible to support BFCache,
-follow the "Discard non-"[=Document/fully active=]" documents for situations that can't be supported" pattern described below.
+follow the [[#discard]] pattern described below.
   
-Additionally, when a document becomes "[=Document/fully active=]" again,
+Additionally, when a document becomes [=Document/fully active=] again,
 it can be useful to update it with the current state of the world,
-if anything has changed while it is in the non-"[=Document/fully active=]" state.
+if anything has changed while it is in the non-[=Document/fully active=] state.
 However, care needs to be taken with events that occurred while in the BFCache.
-When not "[=Document/fully active=]", for some cases, all events should be dropped,
+When not [=Document/fully active=], for some cases, all events should be dropped,
 in some the latest state should be delivered in a single event,
 in others it may be appropriate to queue events or deliver a combined event.
 The correct approach is case by case and should consider privacy,
@@ -601,26 +607,27 @@ correctness, performance and ergonomics.
 
 <div class="example">
   The {{gamepadconnected}} event
-  can be sent to a document that becomes "[=Document/fully active=]" again
-  if a gamepad is connected while the document is not "[=Document/fully active=]".
+  can be sent to a document that becomes [=Document/fully active=] again
+  if a gamepad is connected while the document is not [=Document/fully active=].
   If the gamepad was repeatedly connected and disconnected,
   only the final connected event should be delivered.
+  (This is not specified yet, see [issue](https://github.com/w3c/gamepad/issues/149))
 </div>
 <div class="example">
   For geolocation or other physical sensors,
-  no information about what happened while not-"[=Document/fully active=]" should be delivered.
-  The events should simply resume from when the document became "[=Document/fully active=]"
+  no information about what happened while not [=Document/fully active=] should be delivered.
+  The events should simply resume from when the document became [=Document/fully active=].
 </div>
 <div class="example">
   For network connections or streams,
-  the data received while not-"[=Document/fully active=]" should be delivered only
-  when the document becomes "[=Document/fully active=]" again,
+  the data received while not [=Document/fully active=] should be delivered only
+  when the document becomes [=Document/fully active=] again,
   but whereas a stream might have created many events with a small amount of data each,
   it could be delivered as smaller number of events with more data in each.
 </div>
 
-<h4 id="omit-non-fully-active">Omit non-"[=Document/fully active=]" documents from APIs that span multiple documents</h4>
-Non-"[=Document/fully active=]" documents should not be observable,
+<h4 id="omit-non-fully-active">Omit non-[=Document/fully active=] documents from APIs that span multiple documents</h4>
+Non-[=Document/fully active=] documents should not be observable,
 so APIs should treat them as if they no longer exist.
 They should not be visible to the "outside world" through document-spanning APIs
 (e.g. {{Clients/matchAll()|clients.matchAll()}}, {{Window/opener|window.opener}}).
@@ -632,13 +639,14 @@ Note: This should be rare since cross-document-spanning APIs are themselves rela
 </div>
 <div class="example">
   {{Clients/matchAll()|clients.matchAll()}}
-  currently does not distinguish between "[=Document/fully active=]"
-  and non-"[=Document/fully active=]" clients,
-  but correct implementations should only return "[=Document/fully active=]" clients.
+  currently does not distinguish between [=Document/fully active=]
+  and non-[=Document/fully active=] clients,
+  but correct implementations should only return [=Document/fully active=] clients.
+  (See [issue](https://github.com/w3c/ServiceWorker/issues/1594))
 </div>
 
-<h4 id="discard">Discard non-"[=Document/fully active=]" documents for situations that can't be supported</h4>
-If supporting non-"[=Document/fully active=]" documents is not possible for certain cases,
+<h4 id="discard">Discard non-[=Document/fully active=] documents for situations that can't be supported</h4>
+If supporting non-[=Document/fully active=] documents is not possible for certain cases,
 explicitly specify it by discarding the document if the situation happens after the user navigated away,
 or setting the document's [salvageable](https://html.spec.whatwg.org/multipage/browsing-the-web.html#concept-document-salvageable)
 bit to false if the situation happens before or during the navigation away from the document,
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -522,6 +522,158 @@ See also:
 * [Security and privacy are essential](https://www.w3.org/2001/tag/doc/ethical-web-principles/#privacy)
 * [What data does this specification expose to an origin?](https://www.w3.org/TR/security-privacy-questionnaire/#underlying-platform-data)
 
+<h3 id="support-non-fully-active">Support non-"fully active" documents</h3>
+
+After a user navigated away from a document,
+the document might stay around in a non-"[=Document/fully active=]" state,
```
domenic, line 528: I don't think there's a need to add "scare quotes" around every usage of the term. Maybe the first one, but in other cases just a normal hyperlink suffices.

index.bs
```diff
@@ -522,6 +522,158 @@ See also:
...
+In many cases,
+anything that happens while the document is not "[=Document/fully active=]"
+should be treated as if it never happened,
+and not "queue" the updates and send them after the document becomes fully activated.
+If it makes sense to send the latest state update (and only the latest state,
+not a queue of state changes) to the document after it becomes "[=Document/fully active=]" again,
+consider the "listen for changes" pattern below.
+<div class="example">
+  APIs that periodically send information updates,
+  such as Geolocation API's "[watch position](https://w3c.github.io/geolocation-api/#watchposition-method)"
```
domenic, line 552: I would suggest `{{Geolocation/watchPosition()}}` (with no scare quotes). You may need to update the anchors block like https://github.com/jeremyroman/alternate-loading-modes/blob/00e0523192c13a6d58721c8e73aa5f9b1599eafa/index.bs#L52-L56

index.bs
```diff
@@ -522,6 +522,158 @@ See also:
...
+  should not send updates if the document is no longer fully active.
+  It also should not queue those updates to arrive later,
```
domenic, line 554: The previous sentence is talking about "APIs", with geolocation as one example. But this sentence seems to shift to singular tense and talk about "latest location". It'd be good to reconcile them.

index.bs
```diff
@@ -522,6 +522,158 @@ See also:
...
+When a document goes from "[=Document/fully active=]" to non-"[=Document/fully active=]",
+it should be treated similarly to the way discarded documents are treated.
+The document must not retain exclusive access to shared resources
+and must ensure that no new requests are issued
+and that connections that allow for new incoming requests are terminated. 
+When a document goes from non-"[=Document/fully active=]" to "[=Document/fully active=]" again,
+it can restore connections if appropriate.
+  
+While web authors can manually do cleanup (e.g. release the resources, sever connections)
+from within the "pagehide" event and restore them from the "pageshow" event themselves,
```
domenic, line 569:
```suggestion
from within the {{pagehide}} event and restore them from the {{pageshow}} event themselves,
```

index.bs
```diff
@@ -522,6 +522,158 @@ See also:
...
+doing this automatically from the API design would reduce a lot of overhead.
```
domenic, line 570: I don't think "reduce overhead" is very helpful here. Maybe "is better for web developers" or "is more likely to lead to well-functioning web applications" or something like that?

index.bs
```diff
@@ -522,6 +522,158 @@ See also:
...
+  and re-acquire them when it becomes "[=Document/fully active=]" again (Screen Wake Lock API is already [doing](https://w3c.github.io/screen-wake-lock/#handling-document-loss-of-full-activity) the first part).
+</div>
+Note: this might not be appropriate for all types of resources,
+e.g. if an exclusive lock is held,
+we cannot just release it and reacquire when "[=Document/fully active=]"
+since another page could then take that lock.
+If there is an API to signal to the page that this has happened,
+it may be acceptable but beware that if the only time this happens is with BFCache,
+then it's likely many pages are not prepared for it. If it is not possible to support BFCache,
+follow the "Discard non-"[=Document/fully active=]" documents for situations that can't be supported" pattern described below.
```
domenic, line 590: You could use the `[[#discard]]` syntax to get a nice autolink.

index.bs
```diff
@@ -522,6 +522,158 @@ See also:
...
+if anything has changed while it is in the non-"[=Document/fully active=]" state.
+However, care needs to be taken with events that occurred while in the BFCache.
+When not "[=Document/fully active=]", for some cases, all events should be dropped,
+in some the latest state should be delivered in a single event,
+in others it may be appropriate to queue events or deliver a combined event.
+The correct approach is case by case and should consider privacy,
+correctness, performance and ergonomics.
+
+<div class="example">
+  The [gamepadconnected](https://w3c.github.io/gamepad/#event-gamepadconnected) event
```
domenic, line 603:
```suggestion
  The {{gamepadconnected}} event
```

index.bs
```diff
@@ -522,6 +522,158 @@ See also:
...
+  when the document becomes "[=Document/fully active=]" again,
+  but whereas a stream might have created many events with a small amount of data each,
+  it could be delivered as smaller number of events with more data in each.
+</div>
+
+<h4 id="omit-non-fully-active">Omit non-"[=Document/fully active=]" documents from APIs that span multiple documents</h4>
+Non-"[=Document/fully active=]" documents should not be observable,
+so APIs should treat them as if they no longer exist.
+They should not be visible to the "outside world" through document-spanning APIs
+(e.g. [clients.matchAll()](https://w3c.github.io/ServiceWorker/#clients-matchall), window.opener).
```
domenic, line 626:
```suggestion
(e.g. {{Clients/matchAll()|clients.matchAll()}}, {{Window/opener|window.opener}}).
```

index.bs
```diff
@@ -522,6 +522,158 @@ See also:
...
+
+Note: This should be rare since cross-document-spanning APIs are themselves relatively rare.
+
+<div class="example">
+  [BroadcastChannel](https://html.spec.whatwg.org/multipage/web-messaging.html#broadcasting-to-other-browsing-contexts:fully-active)
+  checks for "[=Document/fully active=]" before sending messages to other browsing contexts.
```
domenic, line 632:
```suggestion
  {{BroadcastChannel}} [checks](https://html.spec.whatwg.org/multipage/web-messaging.html#broadcasting-to-other-browsing-contexts:fully-active) for [=Document/fully active=] before sending messages to other browsing contexts.
```

index.bs
```diff
@@ -522,6 +522,158 @@ See also:
...
+</div>
+<div class="example">
+  clients.matchAll()'s [spec](https://w3c.github.io/ServiceWorker/#clients-matchall)
```
domenic, line 635:
```suggestion
  {{Clients/matchAll()|clients.matchAll()}}
```

index.bs
```diff
@@ -522,6 +522,158 @@ See also:
...
+  currently does not distinguish between "[=Document/fully active=]"
+  and non-"[=Document/fully active=]" clients,
+  but correct implementations should only return "[=Document/fully active=]" clients.
```
domenic, line 638: Can we link to a spec issue about getting this fixed?

index.bs
```diff
@@ -522,6 +522,158 @@ See also:
...
+to cause it to be automatically discarded after navigation.
+
+Note: this should be rare and probably should only be used when retrofitting old APIs,
+as new APIs should always strive to work well with BFCache.
+
+<div class="example">
+  WebSockets [sets the salvageable bit to false](https://html.spec.whatwg.org/#unloading-documents:concept-document-salvageable-7) during unload.
+</div>
+<div class="example">
+  Calling [clients.claim()](https://w3c.github.io/ServiceWorker/#clients-claim)
```
domenic, line 655:
```suggestion
  Calling {{Clients/claim()|clients.claim()}}
```

index.bs
```diff
@@ -522,6 +522,158 @@ See also:
...
+  should not wait for non-"[=Document/fully active=]" clients,
+  instead it should cause the non-"[=Document/fully active=]" client documents to be discarded.
```
domenic, line 657: These "shoulds" are confusing. Does it currently wait for non-fully active clients? I can't find that at a glance. Is there a spec issue about the behavior here?

index.bs
rakina: OK, added this and also mentioned "bfcache by default" a little bit

index.bs
rakina: Oooh, that's neat, thanks!

index.bs
rakina: Linked to https://github.com/w3c/ServiceWorker/issues/1594 now

index.bs
rakina: Linked to the issue, do you think we should rewrite it like the matchAll() one above?

domenic: Either phrasing works; now that you added "this is currently not specified" it's clear.

index.bs
```diff
@@ -522,6 +526,162 @@ See also:
...
+and is considered as "non-existent" from the user's perspective.
+This means they should not receive updates or perform actions.
+
+In many cases,
+anything that happens while the document is not [=Document/fully active=]
+should be treated as if it never happened,
+and not "queue" the updates and send them after the document becomes fully activated.
+If it makes sense to send the latest state update (and only the latest state,
+not a queue of state changes) to the document after it becomes [=Document/fully active=] again,
+consider the "listen for changes" pattern below.
```
domenic, line 553: Maybe use `[[#listen-fully-active]]`

domenic (commented):
Looks good from my perspective! Of course I am not on the TAG so their review is what really matters :)

index.bs
rakina: Added, thanks!

index.bs
```diff
@@ -522,6 +526,162 @@ See also:
...
+making them hard to support in various user agents' implementation of "back/forward cache" (BFCache).
+
+To avoid this happening with your API,
+you should specify support for non-[=Document/fully active=] documents by following these guidelines:
+
+<h4 id="gate-fully-active">Gate actions with [=Document/fully active=] checks</h4>
+
+When performing actions that might update the state of a document,
+be aware that the document might not be [=Document/fully active=]
+and is considered as "non-existent" from the user's perspective.
```
mfalken, line 544: Do we intend prerendering pages to be considered fully active?

In same-origin prerendering at least, we plan to give them access to storage/shared state, and they receive updates. So from that perspective, they should be considered fully active, but the text above suggests that any document that is "non-existent" to the user should be not fully active. cc @domenic 

This also affects how we deal with https://github.com/w3c/ServiceWorker/issues/1594 as the current plan is for the service worker to not see bfcached pages but does see prerendered pages.

rakina: I discussed this with @domenic a while back (see [Google-internal link](https://docs.google.com/document/d/1E93GazamR7FP_w_uBHwMtGeixCeGeiiJGkijpUYHybs/edit?resourcekey=0-ZWf7pJL-rGTUvyMhXlvK_Q&disco=AAAAIPLiOCY)), and I think prerendered page will be fully active but has its "is prerendering" bit set or something set or something like that, since we expect it to behave differently than bfcached pages more often than not.

Perhaps we should add a note here that there are other kinds of documents that are "non-existent" but also "fully active", specifically mentioning prerendered pages and link to https://jeremyroman.github.io/alternate-loading-modes/ ?

mfalken: Thanks I see. Yep adding a note would address my question, thanks!

rakina: Added, thanks!

rakina:
@hober, do you think you/another TAG member can review this and https://github.com/w3ctag/security-questionnaire/pull/128? :)

mfalken:
Do you plan to introduce the "fully active" concept to dedicated workers as well?

It's tricky because dedicated workers can be created by shared workers, and shared workers don't necessarily have a single owner document. So a shared worker can be owned by both a non-fully active document and a fully active document.

I suppose for dedicated workers we want this type of behavior:
- (Document) -> (Dedicated Worker 1) -> (Dedicated Worker 2) --> both dedicated workers are fully active iff document is fully active
- (Document) -> (Shared Worker) -> (Dedicated Worker) --> dedicated worker is fully active even if document is not fully active

And shared workers and service workers are always considered fully active. So maybe it looks like this:

- dedicated worker is fully active iff its creator is fully active
- shared worker is always fully active
- service worker is always fully active

At least for https://github.com/w3c/ServiceWorker/issues/1594 this will come in handy because otherwise we to define the above in the spec to handle the case where `client` is a document vs dedicated worker vs shared worker etc.

index.bs
```diff
@@ -522,6 +526,168 @@ See also:
...
+  {{Clients/matchAll()|clients.matchAll()}}
+  currently does not distinguish between [=Document/fully active=]
+  and non-[=Document/fully active=] clients,
+  but correct implementations should only return [=Document/fully active=] clients.
+  (See [issue](https://github.com/w3c/ServiceWorker/issues/1594))
+</div>
+
+<h4 id="discard">Discard non-[=Document/fully active=] documents for situations that can't be supported</h4>
+If supporting non-[=Document/fully active=] documents is not possible for certain cases,
+explicitly specify it by discarding the document if the situation happens after the user navigated away,
```
mfalken, line 656: Link to https://html.spec.whatwg.org/#discard-a-document here?

index.bs
```diff
@@ -59,6 +59,10 @@ urlPrefix: https://url.spec.whatwg.org/; spec: URL
     url: #percent-encode; type: abstract-op; text: percent-encode
 urlPrefix: https://www.w3.org/TR/payment-request/; spec: payment-request
     url: #dfn-state; type: dfn; text: [[state]];
+urlPrefix: https://w3c.github.io/geolocation-api/; spec: Geolocation
+    url: #watchposition-method; type: interface; for: Geolocation; text: watchPosition()
```
marcoscaceres, line 63: These shouldn't be needed, as they are exported from that spec.

domenic: Nothing is exported from ReSpec EDs since those all happen client side which the database can't pick up.

marcoscaceres: The spec is on TR and evergreen, so should be picked up. If it's not, that's a Bikeshed-data/Shepherd issue (not a Respec issue).  

> Nothing is exported from ReSpec EDs since those all happen client side which the database can't pick up.

You are right, but that's like saying that a bikeshed ".bs" file is not picked up because the spec editor didn't run `bikeshed`. If you spot at spec that is not being snapshotted, please tell those folks to create a snapshot of it (exactly as one would for a .bs spec).

domenic: TR can't be trusted to be evergreen, so as a matter of personal policy I don't encourage people to link to TR (and as a matter of WHATWG policy we never do from WHATWG specs).

index.bs
rakina: Done, thanks!

index.bs
marcoscaceres: I think we should discuss these assumptions (elsewhere, sorry for taking us off topic): You are not wrong, but they are not entirely true either.

torgo (approved):
lgtm however one question below.

index.bs
hober: Removed in fb99f17, thanks @marcoscaceres.

index.bs
```diff
@@ -522,6 +524,168 @@ See also:
...
+
+<h4 id="gate-fully-active">Gate actions with [=Document/fully active=] checks</h4>
+
+When performing actions that might update the state of a document,
+be aware that the document might not be [=Document/fully active=]
+and is considered as "non-existent" from the user's perspective.
+This means they should not receive updates or perform actions.
+
+Note: It is possible for a [=Document/fully active=] document to be perceived as "non-existent" by users,
+such as when the document is [displaying prerendered content](https://jeremyroman.github.io/alternate-loading-modes/).
```
torgo, line 546: Hi - is there a more authoritative reference for this concept that could be linked here – e.g. something in standards track?

index.bs
domenic: :(

index.bs
rakina: cc @domenic @jeremyroman

domenic: I don't think there is: prerendering is not well defined anywhere, despite being implemented in Safari stable and behind a flag in Chrome, so that is why we're trying to create a spec that defines it well.

index.bs
```diff
@@ -522,6 +524,168 @@ See also:
 * [Security and privacy are essential](https://www.w3.org/2001/tag/doc/ethical-web-principles/#privacy)
 * [What data does this specification expose to an origin?](https://www.w3.org/TR/security-privacy-questionnaire/#underlying-platform-data)
 
+<h3 id="support-non-fully-active">Support non-"fully active" documents</h3>
+
+After a user navigated away from a document,
+the document might stay around in a non-[=Document/fully active=] state,
+and might be reused when the user navigates back to the entry holding the document.
```
marcoscaceres, line 531: Maybe we be more clear about this being about the BFCache? 

```suggestion
the document might be cached in a non-[=Document/fully active=] state,
and might be reused when the user navigates back to the entry holding the document, which makes navigation fast for users. 
In browsers, this is known as the back/forward cache, or "<abbr title="back/forward cache">BFCache</abbr>" for short. 
```

index.bs
```diff
@@ -522,6 +524,168 @@ See also:
...
+In the past, many APIs have missed specifying support for non-[=Document/fully active=] documents,
+making them hard to support in various user agents' implementation of "back/forward cache" (BFCache).
```
marcoscaceres, line 533:
```suggestion
making them hard to support in various user agents to cache pages in the BFCache, effectively making the user experience of navigating back and forth less optimal.
```

index.bs
marcoscaceres: I wonder if we should instead point to https://html.spec.whatwg.org/#link-type-prerender ? Given that what's at the link above is supposed to replace what's in HTML anyway?

marcoscaceres (commented):
This is starting to look really good! 

Personally, I'd still like to see some guidance provided around developers holding references to objects that should have been GC'd when a document becomes not fully active (i.e., all my stolen `contentWindow` objects from  `<iframe>`s that then go an try to use APIs). 

Although an edge case, it comes up as a source of interop issues - like it did in the Permissions API.

index.bs
```diff
@@ -59,6 +59,8 @@ urlPrefix: https://url.spec.whatwg.org/; spec: URL
     url: #percent-encode; type: abstract-op; text: percent-encode
 urlPrefix: https://www.w3.org/TR/payment-request/; spec: payment-request
     url: #dfn-state; type: dfn; text: [[state]];
+urlPrefix: https://w3c.github.io/gamepad/; spec: Gamepad
+    url: #event-gamepadconnected; type: event; text: gamepadconnected
```
marcoscaceres, line 63: I believe this can go... bikeshed should be able to auto link it. 

```suggestion
```

index.bs
rakina: https://html.spec.whatwg.org/#link-type-prerender doesn't really explain too much but I guess that's the most official link we can get for now. Changed it to that, but I'm fine with the old link as well.

rakina:
Thanks for the reviews!

> Personally, I'd still like to see some guidance provided around developers holding references to objects that should have been GC'd when a document becomes not fully active (i.e., all my stolen contentWindow objects from <iframe>s that then go an try to use APIs).

Right, I think is kinda covered by some of the points here like "check if document is fully active", but I've also added a note calling out the non-bfcache cases explicitly.

index.bs
domenic: I'd strongly prefer the old link as the HTML section is so vague as to be useless; it should honestly be removed from HTML.

index.bs
domenic: This is the same problem as before where doing so will link to the TR version :(. But last time @hober overrode our work to link to the ED so it seems unlikely we're going to get a good outcome here.

index.bs
marcoscaceres: hehe, I have to agree. I guess we can always update this document once HTML gets updated with something better.

marcoscaceres:
I want to add my ✅ to this guidance from @rakina. It's been tremendously useful and, with @rakina's help, we've made (and are continuing to make!) a bunch of improvements to various specs as a direct result. 

I think a lot of specs will benefit from this.

index.bs
rakina: OK, reverted to old link then :)

rakina:
Thanks @marcoscaceres and all, all the discussions that happened here and lots of other threads was really helpful in ensuring we have a usable guide! @torgo, hopefully this is mergeable now :)

hober (approved):
Thanks for all of your hard work on this! It's great! :)

PR #326, opened 2021-07-07
Add HTML section, add attribute re-use subsection

### LINKED ISSUE #281: HTML: Re-use existing attribute names for similar concepts
New elements should try to re-use existing attribute names for similar concepts.

Examples:
- The `multiple` attribute is used in both `<select>` and `<input type=file>`
- The `open` attribute was introduced in `<details>` and re-used for `<dialog>`
- The `for` attribute is used in both `<label>` and `<output>`

(I did not include the numerous cases of the same attribute being used across multiple elements for the same thing, e.g. `src` or `type`)

LeaVerou:
Resolves #281.

I also added an HTML section, since it will be needed for a variety of future principles, even though this particular one could conceivably also go under Naming.

Not sure if I should scope the new section to be more general than just HTML (e.g. "HTML and other markup languages of the Web Platform") or spell out the acronym like the next section title does.

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/326.html" title="Last updated on Aug 2, 2021, 4:06 PM UTC (c9cbf71)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/326/1a842fa...c9cbf71.html" title="Last updated on Aug 2, 2021, 4:06 PM UTC (c9cbf71)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -575,6 +575,49 @@ See also:
 * [Security and privacy are essential](https://www.w3.org/2001/tag/doc/ethical-web-principles/#privacy)
 * [What data does this specification expose to an origin?](https://www.w3.org/TR/security-privacy-questionnaire/#underlying-platform-data)
 
+<h2 id="css">HTML</h2>
+
+This section details design principles for features which are exposed via HTML.
+
+<h3 id="attribute-reuse">Re-use attribute names for similar functionality</h3>
+
+If you are adding a feature that is specified through an HTML attribute,
+check if there is an existing attribute name on another element
+that specifies similar functionality.
+Re-using an existing attribute name means authors can utilize existing knowledge,
+maintains [consistency](#consistency) across the language,
+and keeps its vocabulary small.
+
+<div class="example">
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -575,6 +575,49 @@ See also:
 * [Security and privacy are essential](https://www.w3.org/2001/tag/doc/ethical-web-principles/#privacy)
 * [What data does this specification expose to an origin?](https://www.w3.org/TR/security-privacy-questionnaire/#underlying-platform-data)
 
+<h2 id="css">HTML</h2>
```
LeaVerou, line 578:
```suggestion
<h2 id="html">HTML</h2>
```

kenchris:
LGTM, but I think we should state that behavior should obviously be similar - if it differs too much, another name might be better

kenchris (approved):

torgo (approved):
lgtm

torgo:
As agreed on call today.

PR #356, opened 2022-01-25
New principle: Where possible APIs should be available in DedicatedWorker

### LINKED ISSUE #325: Guidance about exposing on Window vs Worker only (and other contexts)
This came up in https://github.com/w3ctag/design-reviews/issues/612#issuecomment-862713258 and it was requested that the TAG reaches consensus on guidance on this.

kenchris:
Closes #325

***
<a href="https://pr-preview.s3.amazonaws.com/kenchris/design-principles/pull/356.html" title="Last updated on Jan 27, 2022, 7:34 AM UTC (6f63fe3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/356/51c4da3...kenchris:6f63fe3.html" title="Last updated on Jan 27, 2022, 7:34 AM UTC (6f63fe3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2536,6 +2536,24 @@ offers guidance that should be considered in the development of new features, no
 * [[#feature-detect]]
 * Polyfill development should be encouraged
 
+<h3 id="consider-dedicatedworker-support">Where possible APIs should be made available in DedicatedWorker</h3>
+
+When exposing a feature, please consider whether it makes sense to expose the feature
+to DedicatedWorker as well.
+A DedicatedWorker is a non specialized worker type that allows developers to run code
+off the main thread.
+
+Many features could work out of the box on a DedicatedWorker and not enabling the feature
+there could limit the ability for users to run their code in a non blocking manner.
+
+Certain challenges can exist when trying to enable a feature on DedicatedWorker,
+especially if the feature requires user input by asking for permission,
+or showing a picker or selector.
+Even though this might discourage spec authors to support workers,
+we still recommend designing the feature with DedicatedWorker support in mind,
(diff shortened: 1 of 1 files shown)


annevk (commented):
Can this advice be made more generic to cover all types of workers?

I suspect folks also might want guidance around worklets and `Exposed=*`.

index.bs
```diff
@@ -2536,6 +2536,24 @@ offers guidance that should be considered in the development of new features, no
 * [[#feature-detect]]
 * Polyfill development should be encouraged
 
+<h3 id="consider-dedicatedworker-support">Where possible APIs should be made available in DedicatedWorker</h3>
+
+When exposing a feature, please consider whether it makes sense to expose the feature
+to DedicatedWorker as well.
+A DedicatedWorker is a non specialized worker type that allows developers to run code
+off the main thread.
```
cynthia, line 2544: I think we can just reference back to DedicatedWorker and not explain this ourselves.

kenchris: But then devs will have to open multiple links, maybe do both?

cynthia (approved):

kenchris:
@annevk if you have suggestions, please share. I think we want to land something now and then fix in follow ups as we get consensus

index.bs
cynthia, line 2544:
```suggestion
```

annevk:
I filed #359 as a follow-up. I'm a bit surprised this was just landed without some more time for back-and-forth discussion. At least in the past updates to this document were done quite carefully.

PR #357, opened 2022-01-25
Describe what to do when a feature is temporarily only available on some platforms.

jyasskin:
@chrishtr wanted to document guidance on this situation, and I thought it would make sense to put it here rather than in a Chromium-specific place. Let me know if you want different guidance or think this isn't the right place after all.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/design-principles/pull/357.html#devices-platforms" title="Last updated on Apr 6, 2023, 6:36 AM UTC (479f53f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/357/e81cae6...jyasskin:479f53f.html" title="Last updated on Apr 6, 2023, 6:36 AM UTC (479f53f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -275,8 +275,8 @@ and allows users to choose their own font and font size without causing text to 
 
 Sometimes features aren't yet implemented on some platforms
 despite working on others.
-Design these cases so that
-code can gracefully fail or be polyfilled where the feature isn't yet implemented
+In these cases, it should be possible for
+code to gracefully fail or be polyfilled.
 also works on the missing platforms.
 See [[#feature-detect]].
 
--- a/index.bs
+++ b/index.bs
@@ -275,7 +275,7 @@ and allows users to choose their own font and font size without causing text to 
 
 Sometimes features aren't yet implemented on some platforms
 despite working on others.
-In these cases, it should be possible for
+In these cases, features should be designed such that it is possible for
 code to gracefully fail or be polyfilled.
 See [[#feature-detect]].
 
--- a/index.bs
+++ b/index.bs
@@ -277,7 +277,6 @@ Sometimes features aren't yet implemented on some platforms
 despite working on others.
 In these cases, it should be possible for
 code to gracefully fail or be polyfilled.
-also works on the missing platforms.
 See [[#feature-detect]].
 
 <h3 id="leave-the-web-better">Leave the web better than you found it</h3>
--- a/index.bs
+++ b/index.bs
@@ -276,7 +276,7 @@ and allows users to choose their own font and font size without causing text to 
 Sometimes features aren't yet implemented on some platforms
 despite working on others.
 Design these cases so that
-code written to work on browsers where the feature isn't yet implemented
+code can gracefully fail or be polyfilled where the feature isn't yet implemented
 also works on the missing platforms.
 See [[#feature-detect]].
 
--- a/index.bs
+++ b/index.bs
@@ -273,6 +273,13 @@ so that it works across screen sizes,
 and allows users to choose their own font and font size without causing text to overflow.
 </div>
 
+Sometimes features aren't yet implemented on some platforms
+despite working on others.
+Design these cases so that
+code written to work on browsers where the feature isn't yet implemented
+also works on the missing platforms.
+See [[#feature-detect]].
+
 <h3 id="leave-the-web-better">Leave the web better than you found it</h3>
 
 As you add new capabilities to the web platform, do so in a way that improves
--- a/index.bs
+++ b/index.bs
@@ -273,7 +273,7 @@ so that it works across screen sizes,
 and allows users to choose their own font and font size without causing text to overflow.
 </div>
 
-Sometimes features aren't yet implemented on some platforms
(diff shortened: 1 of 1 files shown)


cynthia (approved):
LGTM. Thanks for the PR!

index.bs
```diff
@@ -263,6 +263,13 @@ so that it works across screen sizes,
 and allows users to choose their own font and font size without causing text to overflow.
 </div>
 
+Sometimes features aren't yet implemented on some platforms
+despite working on others.
+Design these cases so that
+code written to work on browsers where the feature isn't yet implemented
+also works on the missing platforms.
+See [[#feature-detect]].
+
```
cynthia, line 272: Might be useful to have an example here.

index.bs
```diff
@@ -263,6 +263,13 @@ so that it works across screen sizes,
 and allows users to choose their own font and font size without causing text to overflow.
 </div>
 
+Sometimes features aren't yet implemented on some platforms
+despite working on others.
+Design these cases so that
+code written to work on browsers where the feature isn't yet implemented
```
cynthia, line 269:
```suggestion
code can gracefully fail or be polyfilled where the feature isn't yet implemented
```

index.bs
```diff
@@ -263,6 +263,13 @@ so that it works across screen sizes,
...
+also works on the missing platforms.
```
cynthia, line 270: ...and thanks to the suggestion above this bit no longer makes sense. Hmm.

index.bs
```diff
@@ -263,6 +263,13 @@ so that it works across screen sizes,
 and allows users to choose their own font and font size without causing text to overflow.
 </div>
 
+Sometimes features aren't yet implemented on some platforms
+despite working on others.
+Design these cases so that
+code can gracefully fail or be polyfilled where the feature isn't yet implemented
+also works on the missing platforms.
```
rhiaro, line 270:
```suggestion
```

index.bs
```diff
@@ -263,6 +263,13 @@ so that it works across screen sizes,
 and allows users to choose their own font and font size without causing text to overflow.
 </div>
 
+Sometimes features aren't yet implemented on some platforms
+despite working on others.
+Design these cases so that
+code can gracefully fail or be polyfilled where the feature isn't yet implemented
```
rhiaro, line 269:
```suggestion
In these cases, it should be possible for
code to gracefully fail or be polyfilled.
```

index.bs
```diff
@@ -263,6 +263,12 @@ so that it works across screen sizes,
 and allows users to choose their own font and font size without causing text to overflow.
 </div>
 
+Sometimes features aren't yet implemented on some platforms
```
torgo, line 266:
```suggestion
Sometimes features aren't yet implemented on some implementations
```

plinss:
@jyasskin @chrishtr There's some debate among the TAG about exactly what you're asking for here, and we want to make sure we're addressing the right thing.

My interpretation is that some features depend on hardware or OS capabilities, so even when implemented in a given engine, may not always be available depending on where the engine is running. It should be possible to detect that situation (vs not implemented at all) and possibly polyfill.

plinss:
Gentle ping @jyasskin @chrishtr

jyasskin:
Unfortunately, we don't remember which particular feature we were thinking of either. I'm pretty sure it was the situation @plinss describes, where the same engine, running in different places, might or might not provide the API. [Barcode Detection](https://developer.mozilla.org/en-US/docs/Web/API/Barcode_Detection_API#browser_compatibility) is an example of that, although it's also an example of a feature that's only implemented in one engine so far.

My original text said,
> Design these cases so that code written to work on browsers where the feature isn't yet implemented also works on the missing platforms.

That is, there are always some old browsers that don't have the feature yet because they predate the time the feature shipped anywhere, even if all of the engines ship about the same time. Developers have to write code to detect the feature on those old browsers. That same code, when running in a browser version that does implement the feature in some places, should also distinguish between platforms (e.g. MacOS) where the feature exists and other platforms (e.g. Windows) where the feature doesn't.

dbaron:
I think the existing document already provide some guidance on this, in [2.5 New features should be detectable](https://w3ctag.github.io/design-principles/#feature-detect):
> Provide a way for authors to programmatically detect whether your feature is available, so that web content may gracefully handle the feature not being present.
> 
> An existing feature may not be available on a page for a number of reasons. Two of the more common reasons are because it hasn’t been implemented yet, or because it’s only available in [secure contexts](https://w3ctag.github.io/design-principles/#secure-context).
> 
> Authors shouldn’t need to write different code to handle each scenario. That way, even if an author only knows or cares about one scenario, the code will handle all of them.

index.bs
torgo:
```suggestion
Sometimes features aren't yet implemented on some implementations or platforms
```

index.bs
torgo: Created an issue: https://github.com/w3ctag/design-principles/issues/428

torgo (approved):
approved as discussed in today's call

Issue #131, opened 2019-06-20
Discourage overloading


annevk:
In particular the kind where subsequent arguments (both in terms of the amount of them and in terms of their type) depend on a prior argument. We have some of this in the platform and generally dictionaries or some such would have been better.

dbaron:
Do you have an example or two in your head already?

annevk:
* https://xhr.spec.whatwg.org/#interface-formdata
* https://html.spec.whatwg.org/multipage/canvas.html#canvasdrawpath
* https://html.spec.whatwg.org/multipage/canvas.html#canvasdrawimage

kenchris:
In a sense, optional arguments are related and could be seen as overloading, and that is quite common on the web platform.

Could you explain the reasoning behind this suggestion?

js-choi:
I’d also like the term “overloading” to be made more precise. Ideally, it would exclude type overloading functions’ individual parameters.

(For example, in JavaScript, [TC39 is considering adding type overloading to `Math.pow`](https://github.com/tc39/proposal-bigint-math), making it return BigInts when given BigInt arguments. This is an attempt to match the precedent of JavaScript’s existing math operators such as `**`, whose parameters are already type overloaded.)

annevk:
@kenchris I thought OP made it quite clear what kind of overloading was meant and it explicitly excludes ordinary optional arguments. So I'm not sure what clarification you are seeking.

@js-choi that's interesting. You could express most of that with unions in Web IDL so that seems fine, though you cannot really express the full intent of the argument and return value having to match in type, which might be nice for code generation and such.

LeaVerou:
@annevk We are seeking clarification on *why* OP sees this kind of overloading as harmful.

annevk:
I see, because it's not very ideomatic, results in rather complex bindings, and more importantly makes it quite confusing what the actual signature of a method is when reading and writing code. It stems from an era when dictionaries were not yet a thing.

kenchris:
@littledan Hi Dan, do you know of people on the TC39 has discussed overloading in JavaScript and/or have opinions on this topic?

LeaVerou:
[...]

PR #372, opened 2022-05-19
New principle: on overloading, for #131

rhiaro:
WIP

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/372.html" title="Last updated on Jul 4, 2022, 6:39 PM UTC (33081c2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/372/025c56e...33081c2.html" title="Last updated on Jul 4, 2022, 6:39 PM UTC (33081c2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1549,6 +1549,33 @@ See also:
 
 * [[#naming-is-hard]]
 
+<h3 id="overloading">Use overloading wisely</h3>
+
+If the signature or behaviour of a method will be significantly different
+depending on the arguments passed
+you should define a separate method
+rather than overload a single method.
+
+Overloading a method such that one of its arguments can be
+a single thing or an array is useful.
+Similarly, passing a dictionary of options,
+or using optional arguments,
+are good patterns.
+
+However, if subsequent arguments for a method have to change
+because of the first argument passed,
+or if a method has different behaviour depending on the types of inputs,
+this can hinder code readability,
+and make discovering API functionality more difficult for authors,
+and should be avoided.
+
+<div class="example">
+TODO
+https://xhr.spec.whatwg.org/#interface-formdata
+https://html.spec.whatwg.org/multipage/canvas.html#canvasdrawpath
+https://html.spec.whatwg.org/multipage/canvas.html#canvasdrawimage
+"For the examples I gave I would expect either dictionaries or separate method names to be used."
+
 <h3 id="constructors">Classes should have constructors when possible</h3>
 
 Make sure that any class that's part of your API has a constructor,
--- a/index.bs
+++ b/index.bs
@@ -1568,13 +1568,6 @@ this can hinder code readability,
 and make discovering API functionality more difficult for authors,
 and should be avoided.
 
-<div class="example">
-TODO
-https://xhr.spec.whatwg.org/#interface-formdata
-https://html.spec.whatwg.org/multipage/canvas.html#canvasdrawpath
-https://html.spec.whatwg.org/multipage/canvas.html#canvasdrawimage
-"For the examples I gave I would expect either dictionaries or separate method names to be used."
-
 <h3 id="constructors">Classes should have constructors when possible</h3>
 
 Make sure that any class that's part of your API has a constructor,
--- a/index.bs
+++ b/index.bs
@@ -1558,7 +1558,7 @@ rather than overload a single method.
 
 Overloading a method such that one of its arguments can be
 either a single value or an array of values is useful.
-Similarly, passing a dictionary of options,
+Similarly, passing a dictionary of options
 or using optional arguments,
 are good patterns because they allow for more flexibility.
 
--- a/index.bs
+++ b/index.bs
@@ -1559,7 +1559,6 @@ rather than overload a single method.
 Overloading a method such that one of its arguments can be
 either a single value or an array of values is useful.
 Similarly, passing a dictionary of options
-or using optional arguments,
 are good patterns because they allow for more flexibility.
 
 However, if subsequent arguments for a method have to change
--- a/index.bs
+++ b/index.bs
@@ -1557,7 +1557,7 @@ you should define a separate method
 rather than overload a single method.
 
 Overloading a method such that one of its arguments can be
-a single thing or an array is useful.
+either a single value or an array of values is useful.
 Similarly, passing a dictionary of options,
 or using optional arguments,
 are good patterns.
--- a/index.bs
+++ b/index.bs
@@ -1559,7 +1559,7 @@ rather than overload a single method.
 Overloading a method such that one of its arguments can be
 either a single value or an array of values is useful.
 Similarly, passing a dictionary of options
-are good patterns because they allow for more flexibility.
+is a good pattern because it allows for more flexibility.
 
 However, if subsequent arguments for a method have to change
 because of the first argument passed,
--- a/index.bs
+++ b/index.bs
@@ -1553,7 +1553,7 @@ See also:
 
 If the behaviour of a method will be significantly different
 depending on the arguments passed
-you should define a separate method
+you should usually define a separate method
 rather than overload a single method.
 
 Overloading a method such that one of its arguments can be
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1549,6 +1549,33 @@ See also:
...
+If the signature or behaviour of a method will be significantly different
+depending on the arguments passed
+you should define a separate method
+rather than overload a single method.
+
+Overloading a method such that one of its arguments can be
+a single thing or an array is useful.
+Similarly, passing a dictionary of options,
+or using optional arguments,
+are good patterns.
```
LeaVerou, line 1563: Using a dictionary of options or optional arguments is not overloading I believe. Are you trying to describe the `addEventListener` case? I can try to reword if so.

index.bs
```diff
@@ -1549,6 +1549,33 @@ See also:
...
+because of the first argument passed,
+or if a method has different behaviour depending on the types of inputs,
+this can hinder code readability,
+and make discovering API functionality more difficult for authors,
+and should be avoided.
+
+<div class="example">
+TODO
+https://xhr.spec.whatwg.org/#interface-formdata
+https://html.spec.whatwg.org/multipage/canvas.html#canvasdrawpath
```
LeaVerou, line 1575: Looking at these:

```webidl
undefined fill(optional CanvasFillRule fillRule = "nonzero");
undefined fill(Path2D path, optional CanvasFillRule fillRule = "nonzero");

undefined clip(optional CanvasFillRule fillRule = "nonzero");
undefined clip(Path2D path, optional CanvasFillRule fillRule = "nonzero");

boolean isPointInPath(unrestricted double x, unrestricted double y, optional CanvasFillRule fillRule = "nonzero");
boolean isPointInPath(Path2D path, unrestricted double x, unrestricted double y, optional CanvasFillRule fillRule = "nonzero");

boolean isPointInStroke(unrestricted double x, unrestricted double y);
boolean isPointInStroke(Path2D path, unrestricted double x, unrestricted double y);
```
Is the problem here really that these should have been different methods? Or improper argument ordering? Or that the alternate signatures should instead have been Path2D methods?
I'm unconvinced having a bunch of similar but slightly different methods like `isPointInStrokePath` would have been a good design.

index.bs
```diff
@@ -1549,6 +1549,33 @@ See also:
...
+https://html.spec.whatwg.org/multipage/canvas.html#canvasdrawimage
```
LeaVerou, line 1576: Yikes, this one is really bad. We could also use it as an example of what not to do for https://w3ctag.github.io/design-principles/#prefer-dict-to-bool

index.bs
```diff
@@ -1549,6 +1549,33 @@ See also:
 
 * [[#naming-is-hard]]
 
+<h3 id="overloading">Use overloading wisely</h3>
+
+If the signature or behaviour of a method will be significantly different
+depending on the arguments passed
+you should define a separate method
```
hober, line 1556:
```suggestion
you should usually define a separate method
```

index.bs
```diff
@@ -1549,6 +1549,33 @@ See also:
 
 * [[#naming-is-hard]]
 
+<h3 id="overloading">Use overloading wisely</h3>
+
+If the signature or behaviour of a method will be significantly different
```
hober, line 1554:
```suggestion
If the behaviour of a method will be significantly different
```

'signature' is redundant here

index.bs
hober: (This, as with almost every principle, isn't absolute.)

index.bs
hober: Optional booleans are generally a smell, though.

index.bs
hober: I actually like this pattern in the Canvas 2D API. there's an implicit current path that gets used unless you pass in a path. And passing in a path as the first argument feels right too, since it's that path that the operation acts on.

hober (changes requested):
Some of my comments are nits that don't need to be addressed, but I think we might not have consensus on some of this.

index.bs
```diff
@@ -1549,6 +1549,33 @@ See also:
...
+depending on the arguments passed
+you should define a separate method
+rather than overload a single method.
+
+Overloading a method such that one of its arguments can be
+either a single value or an array of values is useful.
+Similarly, passing a dictionary of options,
+or using optional arguments,
+are good patterns.
```
torgo, line 1563:
```suggestion
are good patterns because they allow for more flexibility.
```

index.bs
hober: Also numeric arguments, as @plinss just pointed out in this breakout.

index.bs
```diff
@@ -1549,6 +1549,33 @@ See also:
...
+However, if subsequent arguments for a method have to change
+because of the first argument passed,
+or if a method has different behaviour depending on the types of inputs,
+this can hinder code readability,
+and make discovering API functionality more difficult for authors,
+and should be avoided.
+
+<div class="example">
+TODO
+https://xhr.spec.whatwg.org/#interface-formdata
```
hober, line 1574: Maybe the filename should have been in an options bag, but I actually like the overloading here. Form data can be stringy or a blob.

index.bs
hober: Yeah, maybe this is a better example for that principle than this one?

hober:
@plinss, @torgo, and I took a look at this PR in Tokyo. We're not sure about most of the examples in the TODO examples div, so we're gonna drop it, land the PR without those examples, and file a followup issue to add examples to this principle.

index.bs
```diff
@@ -1549,6 +1549,33 @@ See also:
...
+
+If the behaviour of a method will be significantly different
+depending on the arguments passed
+you should usually define a separate method
+rather than overload a single method.
+
+Overloading a method such that one of its arguments can be
+either a single value or an array of values is useful.
+Similarly, passing a dictionary of options,
+or using optional arguments,
```
hober, line 1562:
```suggestion
```

index.bs
```diff
@@ -1549,6 +1549,33 @@ See also:
...
+<h3 id="overloading">Use overloading wisely</h3>
+
+If the behaviour of a method will be significantly different
+depending on the arguments passed
+you should usually define a separate method
+rather than overload a single method.
+
+Overloading a method such that one of its arguments can be
+either a single value or an array of values is useful.
+Similarly, passing a dictionary of options,
```
hober, line 1561:
```suggestion
Similarly, passing a dictionary of options
```

index.bs
hober, line 1562:
```suggestion
```

index.bs
```diff
@@ -1549,6 +1549,33 @@ See also:
...
+or using optional arguments,
+are good patterns because they allow for more flexibility.
```
hober, line 1563:
```suggestion
is a good pattern because it allows for more flexibility.
```

hober (changes requested):
This drops the "or using optional arguments" since both bool optional arguments and numeric optional arguments can be problematic.

hober (approved):

torgo (approved):
lgtm

LeaVerou:
@hober

> I actually like this pattern in the Canvas 2D API. there's an implicit current path that gets used unless you pass in a path. And passing in a path as the first argument feels right too, since it's that path that the operation acts on.

In terms of use cases, both of these should be accommodated: both specifying an explicit path, and implicitly using the last specified path. But that doesn’t mean this is the best design to accommodate this.

Typically, when you end up passing the same context around on multiple procedural functions, it's a smell that these should have really been methods on that context, in this case `Path2D`. Then `CanvasRenderingContext2D` could still have the non-path methods. Or `CanvasRenderingContext2D` could provide a way to get the current path, and then we only need these methods on `Path2D`, instead of having two signatures. That would be lengthier though, so perhaps this may not be the best example for this principle? Unless we frame it as neither good or bad overloading but "look, there are tradeoffs to be made".

PR #386, opened 2022-07-27
Start developer version

LeaVerou:
***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/386.html" title="Last updated on Jan 24, 2024, 9:33 AM UTC (6754594)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/386/7627102...6754594.html" title="Last updated on Jan 24, 2024, 9:33 AM UTC (6754594)">Diff</a>

### DIFF
--- a/dev/README.md
+++ b/dev/README.md
@@ -1,0 +1,13 @@
+To generate developer version `cd` to the root of this repo, then run:
+
+```
+bikeshed spec index.bs dev/index.html  --md-Text-Macro="DEV on"
+```
+
+To exclude and include parts of the document use [Bikeshed conditional inclusion](https://tabatkins.github.io/bikeshed/#conditional):
+
+```bikeshed
+<div include-if="text macro: DEV">HELLO HELLO TEST</div>
+
+<div exclude-if="text macro: DEV">HELLO HELLO TEST</div>
+```
--- a/index.bs
+++ b/index.bs
@@ -1046,7 +1046,7 @@ and because the <{img}> element is an empty element that does not permit child e
 A space-separated syntax would not have been possible, as each list item includes multiple values.
 
 <!--
-TODO: link to the "microsyntaxes are usually bad" principle once it's been written.
+TODO: link to the "microsyntaxes are usually bad" principle once it's been written. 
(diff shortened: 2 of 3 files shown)


dev/README.md
```diff
@@ -0,0 +1,13 @@
+To generate developer version `cd` to the root of this repo, then run:
```
hober, line 1: The Makefile should be updated to have a rule that does this, and these instructions should simply say something like "cd to the root of this repo and type `make dev`" or something.

hober (changes requested):
Please [update the Makefile](https://github.com/w3ctag/design-principles/pull/386/files#r1172101424)

dev/README.md
plinss: Given that we're using Bikeshed, which is Python based, I'd rather not add a nodejs dependency.

LeaVerou:
Per breakout discussion today: 

- [x] Action item for @LeaVerou to ping @tabatkins to fix Bikeshed bug
- [ ] Action item for @plinss to explore using Jinja + make to produce the two versions, so we don't depend on Bikeshed. Then we can also have more fine grained conditionals, whereas bikeshed can only do whole elements, so it may be a better solution altogether.

index.bs
```diff
@@ -1035,7 +1046,7 @@ and because the <{img}> element is an empty element that does not permit child e
 A space-separated syntax would not have been possible, as each list item includes multiple values.
 
 <!--
-TODO: link to the "microsyntaxes are usually bad" principle once it's been written. 
+TODO: link to the "microsyntaxes are usually bad" principle once it's been written.
```
cynthia, line 1049:
```suggestion
TODO: link to the "microsyntaxes are usually bad" principle once it's been written. 
```

LeaVerou:
Closing this since we resolved to use a Python-based workflow rather than Bikeshed's inclusion/exclusion capabilities. @plinss you said you'd explore using Jinja to preprocess the document, did you make any progress on that? If are able to just add it to the spec, we can take it from there.

PR #408, opened 2022-12-20
Warn against getter throwing exception since it is a side effect

wolenetz:
Also includes some bikeshed indentation error fixes.

***
<a href="https://pr-preview.s3.amazonaws.com/wolenetz/design-principles/pull/408.html" title="Last updated on Apr 20, 2023, 2:26 AM UTC (2d154fd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/408/69b2ff0...wolenetz:2d154fd.html" title="Last updated on Apr 20, 2023, 2:26 AM UTC (2d154fd)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -704,7 +704,7 @@ following categories:
 * Makes cross-document interaction/resource sharing possible
   (e.g. holding locks)
 * May malfunction when a document is
-    kept in a non-[=Document/fully active=] BFCached state
+  kept in a non-[=Document/fully active=] BFCached state
     (instead of getting destroyed)
     after the user navigates away from it
     or gets restored
--- a/index.bs
+++ b/index.bs
@@ -700,12 +700,12 @@ If your feature does **anything**
 that falls into the
 following categories:
 * Interacts with a document from the "outside"
-  (e.g. sends information to a document)
+    (e.g. sends information to a document)
 * Makes cross-document interaction/resource sharing possible
-  (e.g. holding locks)
+    (e.g. holding locks)
 * May malfunction when a document is
-  kept in a non-[=Document/fully active=] BFCached state
-  (instead of getting destroyed)
+    kept in a non-[=Document/fully active=] BFCached state
+    (instead of getting destroyed)
     after the user navigates away from it
     or gets restored
     (e.g. expects that a state saved in the document
@@ -714,7 +714,7 @@ following categories:
 Specify how your feature works
 with non-[=Document/fully active=] BFCached documents,
 following the guidelines in the
-[Supporting BFCached Documents](https://w3ctag.github.io/bfcache-guide/) guide.  
+[Supporting BFCached Documents](https://w3ctag.github.io/bfcache-guide/) guide.
 
 Note: It is possible for a document
 to become non-[=Document/fully active=]
--- a/index.bs
+++ b/index.bs
@@ -1361,7 +1361,7 @@ whether to use an attribute or a method.
 
 - Attribute getters should not throw exceptions.
     Exceptions in the context of getters are side effects.
-    Updating existing APIs to include getters that throw exception must be avoided.
+    Updating existing getters to throw exceptions must be avoided as existing API users may enumerate or wrap the API and not expect the new exception, breaking backwards compatibility.
 
 Note: For example, existing API users may enumerate or wrap the API and not expect
 the new exception, breaking backwards compatibility of the API change.
--- a/index.bs
+++ b/index.bs
@@ -702,7 +702,7 @@ following categories:
 * Interacts with a document from the "outside"
   (e.g. sends information to a document)
 * Makes cross-document interaction/resource sharing possible
-    (e.g. holding locks)
+  (e.g. holding locks)
 * May malfunction when a document is
     kept in a non-[=Document/fully active=] BFCached state
     (instead of getting destroyed)
--- a/index.bs
+++ b/index.bs
@@ -1362,10 +1362,6 @@ whether to use an attribute or a method.
 - Attribute getters should not throw exceptions.
     Exceptions in the context of getters are side effects.
     Updating existing getters to throw exceptions must be avoided as existing API users may enumerate or wrap the API and not expect the new exception, breaking backwards compatibility.
-
-Note: For example, existing API users may enumerate or wrap the API and not expect
-the new exception, breaking backwards compatibility of the API change.
-
 - Attribute getters should not perform any blocking operations.
     If a getter requires performing a blocking operation,
     it should be a method.
--- a/index.bs
+++ b/index.bs
@@ -1359,6 +1359,13 @@ whether to use an attribute or a method.
 - Attribute getters must not have any (observable) side effects.
     If you have expected side effects, use a method.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1359,6 +1359,13 @@ whether to use an attribute or a method.
 - Attribute getters must not have any (observable) side effects.
     If you have expected side effects, use a method.
 
+- Attribute getters should not throw exceptions.
+    Exceptions are side effects.
```
annevk, line 1363: I'm not sure it's widely accepted that exceptions are side effects. E.g., I'm pretty sure typical side-effect-free functions are still allowed to do input validation.

I can kinda see how they can cause side effects in JavaScript, but only if you have listeners in place that end up modifying state in response to the exception.

index.bs
```diff
@@ -1359,6 +1359,13 @@ whether to use an attribute or a method.
...
+    Updating existing APIs to include getters that throw exception must be avoided.
```
annevk, line 1364: I think it can still be reasonable if that only happens in a new state that was not accessible before the update, but this is an interesting observation.

annevk (commented):
To be clear, in general I agree that you probably do not want your getters to throw, but I think in practice all platform object getters can throw. For instance, if their this doesn't have the correct platform object brand.

cc @domenic

annevk:
(Sorry for not calling out these caveats earlier. I guess they didn't really occur to me when you filed the issue.)

index.bs
cynthia, line 1364:
```suggestion
    Updating existing APIs to include getters that throw an exception must be avoided.
```

index.bs
cynthia, line 1363:
```suggestion
    Exceptions in the context of getters are side effects.
```

cynthia (approved):

index.bs
LeaVerou, line 1364:
```suggestion
    Updating existing getters to throw exceptions must be avoided.
```

index.bs
```diff
@@ -1359,6 +1359,13 @@ whether to use an attribute or a method.
...
+
+Note: For example, existing API users may enumerate or wrap the API and not expect
+the new exception, breaking backwards compatibility of the API change.
+
```
LeaVerou, line 1368:
```suggestion
```

(see change above)

index.bs
LeaVerou: I think this is beside the point, because exceptions being side effects  is *not* the reason getters should not throw. Getters should behave like regular JS properties, and these do not throw.
Also, typically, invalid state throws when *set*, it's not a good design to allow writes to go through only to realize you have invalid state when someone tries to read it.

atanassov:
```suggestion
    Exceptions in the context of getters are considered side effects.
```

LeaVerou:
```suggestion
    Getters should [behave like regular data properties](#attributes-like-data), and regular data properties do not throw exceptions when read. Furthermore, invalid state should generally be avoided by rejecting *writes*, not when data is *read*.
```

index.bs
```diff
@@ -1359,6 +1359,9 @@ whether to use an attribute or a method.
 - Attribute getters must not have any (observable) side effects.
     If you have expected side effects, use a method.
 
+- Attribute getters should not throw exceptions.
+    Exceptions in the context of getters are side effects.
```
LeaVerou, line 1363:
```suggestion
    Getters should [behave like regular data properties](#attributes-like-data), and regular data properties do not throw exceptions when read. Furthermore, invalid state should generally be avoided by rejecting *writes*, not when data is *read*.
```

index.bs
```diff
@@ -1359,6 +1359,9 @@ whether to use an attribute or a method.
...
+    Updating existing getters to throw exceptions must be avoided as existing API users may enumerate or wrap the API and not expect the new exception, breaking backwards compatibility.
```
cynthia, line 1364:
```suggestion
    Updating existing getters to throw exceptions must be avoided as existing API users may enumerate or wrap the API and not expect the new exception, breaking backwards compatibility.

```

index.bs
```diff
@@ -700,12 +700,12 @@ If your feature does **anything**
 that falls into the
 following categories:
 * Interacts with a document from the "outside"
-  (e.g. sends information to a document)
+    (e.g. sends information to a document)
```
cynthia, line 703:
```suggestion
  (e.g. sends information to a document)
```

index.bs
```diff
@@ -700,12 +700,12 @@ If your feature does **anything**
...
 * Makes cross-document interaction/resource sharing possible
-  (e.g. holding locks)
+    (e.g. holding locks)
```
cynthia, line 705:
```suggestion
  (e.g. holding locks)
```

index.bs
```diff
@@ -700,12 +700,12 @@ If your feature does **anything**
...
 * May malfunction when a document is
-  kept in a non-[=Document/fully active=] BFCached state
-  (instead of getting destroyed)
+    kept in a non-[=Document/fully active=] BFCached state
```
cynthia, line 707:
```suggestion
  kept in a non-[=Document/fully active=] BFCached state
```

index.bs
```diff
@@ -700,12 +700,12 @@ If your feature does **anything**
...
+    (instead of getting destroyed)
```
cynthia, line 708:
```suggestion
  (instead of getting destroyed)
```

index.bs
```diff
@@ -714,7 +714,7 @@ following categories:
 Specify how your feature works
 with non-[=Document/fully active=] BFCached documents,
 following the guidelines in the
-[Supporting BFCached Documents](https://w3ctag.github.io/bfcache-guide/) guide.  
+[Supporting BFCached Documents](https://w3ctag.github.io/bfcache-guide/) guide.
```
cynthia, line 717:
```suggestion
[Supporting BFCached Documents](https://w3ctag.github.io/bfcache-guide/) guide.  
```

index.bs
hober: Given @annevk's comment, perhaps must should be watered down a bit?

index.bs
```diff
@@ -1359,6 +1359,10 @@ whether to use an attribute or a method.
 - Attribute getters must not have any (observable) side effects.
     If you have expected side effects, use a method.
 
+- Attribute getters should not throw exceptions.
+    Getters should [behave like regular data properties](#attributes-like-data), and regular data properties do not throw exceptions when read. Furthermore, invalid state should generally be avoided by rejecting *writes*, not when data is *read*.
+    Updating existing getters to throw exceptions must be avoided as existing API users may enumerate or wrap the API and not expect the new exception, breaking backwards compatibility.
```
hober, line 1364:
```suggestion
    Updating existing getters to throw exceptions should be avoided as existing API users may enumerate or wrap the API and not expect the new exception, breaking backwards compatibility.
```

hober: Per @annevk's comment.

torgo (approved):
We discussed in today's call and we agreed to merge without Lea's proposed change for now.

PR #409, opened 2022-12-23
Editorial tweaks

gibson042:
* Fix some typos.
* Change some phrasing for readability.
* Merge sections describing expectations of attributes and properties vs. methods.
* Update a section title to imperative form.

***
<a href="https://pr-preview.s3.amazonaws.com/gibson042/design-principles/pull/409.html" title="Last updated on Dec 24, 2022, 5:29 PM UTC (bde8146)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/409/69b2ff0...gibson042:bde8146.html" title="Last updated on Dec 24, 2022, 5:29 PM UTC (bde8146)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2515,7 +2515,7 @@ Browsers support MIME type sniffing for legacy reasons, but we recommend enforci
 Due to security implications, we do not recommend extending
 the [pattern matching algorithm](https://mimesniff.spec.whatwg.org/#image-type-pattern-matching-algorithm).
 
-Each new MIME type should have a specification and should be registered with the Internet Assigned Numbers Authority (IANA).
+Each new MIME type should have a specification and should be registered with the <abbr title="Internet Assigned Numbers Authority">IANA</abbr>.
 
 <h3 id="new-http-header-syntax">Conform new HTTP Headers to standards</h3>
 
@@ -2534,7 +2534,7 @@ or even worse,
 something that would break those existing parsers.
 If the new header requires data
 that can't be represented by Structured Field Values,
-then either engage with the Internet Engineering Task Force (IETF) about
+then either engage with the <abbr title="Internet Engineering Task Force">IETF</abbr> about
 extending the Structured Field Values syntax,
 or re-consider if an HTTP header is a right place
 to expose the data before inventing a new syntax. [[RFC8941]]
--- a/index.bs
+++ b/index.bs
@@ -1203,8 +1203,8 @@ the property won't be updated until the next task runs.
 Ensure your JavaScript Web APIs don't provide a way
 for an author to know the timing of garbage collection.
 
-The timing of garbage collection is different in different user agents,
-and may change over time as user agents work on improving performance.
+The timing of garbage collection differs across user agents,
+and may change over time as implementers work on improving performance.
 If an API exposes the timing of garbage collection,
 it can cause programs to behave differently
 in different contexts.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1691,31 +1651,6 @@ which might be expressed as a bitmask in another language,
 use a dictionary object instead.
 This object can be passed around as easily as a single bitmask value.
 
-<h3 id="properties-vs-methods">Use properties and methods appropriately</h3>
```
LeaVerou, line 1694: What happened to this section?

LeaVerou (changes requested):
See comment about missing section but LGTM otherwise! Thank you!

index.bs
gibson042: Merged into #attributes-like-data along with #attributes-vs-methods, eliminating the redundancy of three sections describing appropriate behavior for attributes/properties vs. methods and lamenting their violation by `offsetTop`.

index.bs
```diff
@@ -2599,7 +2534,7 @@ or even worse,
 something that would break those existing parsers.
 If the new header requires data
 that can't be represented by Structured Field Values,
-then either engage with IETF about
+then either engage with the Internet Engineering Task Force (IETF) about
```
LeaVerou, line 2537: I would use an `<abbr>` here

index.bs
LeaVerou: I agree there is redundancy, and thank you for spotting it! However, right now it feels like a lot of the more concrete guidance was dropped in the process (e.g. the parts about what should equal what). Could you try to fold some of that in?

gibson042: I believe all the points are still there.
* "_Getters must not have any observable side effects._" is currently in all three sections and remains in the proposed merge.
* "_Getters should not perform any {complex,blocking,[inefficient]} operations._" is currently in all three sections and remains in the proposed merge.
* A variation of "_Ensure that `obj.attribute === obj.attribute` is always true. Don't create a new value each time the getter is called._" is currently in all three sections remains in the proposed merge with the least passive phrasing.
* "_[ensure that] given `obj.attribute = x`, [a subsequent] `obj.attribute === x` {is,should be} true._" is currently in all three sections and remains in the proposed merge. And I have now also incorporated the elaborations currently absent from [#attributes-like-data](https://w3ctag.github.io/design-principles/#attributes-like-data).
* Guidance that static objects should be new each time (and therefore returned from methods rather than attributes) is currently absent from [#attributes-like-data](https://w3ctag.github.io/design-principles/#attributes-like-data) but is incorporated in the proposed merge.
* Guidance to use a method if any of the stated prescriptions cannot be satisfied is currently in all three sections and remains in the proposed merge (as the first content after the list of them).
* An observation that `offsetTop` violates the second point and should have been a method is currently in all three sections and remains in the proposed merge (in its clearest form and with outbound links).

index.bs
gibson042: Done, and also for IANA.

index.bs
LeaVerou: @cynthia what do you think?

index.bs
LeaVerou: Thanks!

LeaVerou:
Hi @gibson042, we were just reviewing this in a breakout in our Tokyo F2F but noticed it needs rebasing. Could you please rebase it? Thanks!

cynthia:
Please hold off on the rebase. We discussed this during our F2F and concluded this changes too much at once. For that reason, we don't think we can take this whole PR as is. I'll slowly incorporate some of the pure editorial changes and typos.

cynthia:
Apologies for the long delay, and we appreciate the contribution - but I think it's more work to resolve the conflicts and land this at this point. We'll incorporate the suggestions in a different patch and attribute it accordingly, but likely in a more conservative form. Will follow up in a week or two with a revised PR incorporating this.

PR #435, opened 2023-04-21
Make 6.1 more inclusive of non-Web JS APIs

LeaVerou:
Per our discussion today

More context in https://github.com/w3ctag/design-principles/issues/436

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/435.html" title="Last updated on Apr 21, 2023, 5:48 AM UTC (536a252)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/435/0559332...536a252.html" title="Last updated on Apr 21, 2023, 5:48 AM UTC (536a252)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1252,13 +1252,12 @@ the interaction with garbage collection.
 
 <h2 id="api-surface">JavaScript API Surface Concerns</h2>
 
-<h3 id="attributes-like-data">Attributes should behave like data properties</h3>
+<h3 id="accessors-like-data" oldids="attributes-like-data">Accessors should behave like data properties</h3>
 
-[[!WEBIDL]] attributes should act like simple JavaScript object properties.
+JavaScript accessors (specified as [[!WEBIDL]] attributes with separate getter and setter methods in Web APIs)
+should act like simple JavaScript object properties.
 
-In reality, IDL attributes are implemented as accessor properties
-with separate getter and setter methods.
-To make them act like JavaScript object properties:
+This includes:
 
 * Getters must not have any observable side effects.
 * Getters should not perform any complex operations.
@@ -1268,7 +1267,7 @@ To make them act like JavaScript object properties:
     <code highlight="js">obj.attribute === x</code> is true.
     (This may not be possible if some kind of conversion is necessary for `x`.)
 
-If you were thinking about using an attribute,
+If you were thinking about using an accessor,
 but it doesn't behave this way,
 you should probably use a method instead.
 


index.bs
```diff
@@ -1252,13 +1252,12 @@ the interaction with garbage collection.
 
 <h2 id="api-surface">JavaScript API Surface Concerns</h2>
 
-<h3 id="attributes-like-data">Attributes should behave like data properties</h3>
+<h3 id="accessors-like-data" oldids="attributes-like-data">Accessors should behave like data properties</h3>
 
-[[!WEBIDL]] attributes should act like simple JavaScript object properties.
+JavaScript accessors (specified as [[!WEBIDL]] attributes with separate getter and setter methods in Web APIs)
```
hober, line 1257: I think the original sentence is clearer.

hober (changes requested):

index.bs
```diff
@@ -1252,13 +1252,12 @@ the interaction with garbage collection.
...
+should act like simple JavaScript object properties.
```
LeaVerou, line 1258:
```suggestion
JavaScript accessors 
should act like simple JavaScript object properties.
In [[!WEBIDL]], these are specified as attributes with separate getter and setter methods.
```

plinss: The phrase "these are specified as attributes with separate getter and setter methods" implies that the getters and setters are explicit in the WebIDL, and they're generally implied.

Perhaps something like: "In [[!WEBIDL]] these are specified as attributes, which are implemented as getter and setter methods."

cynthia:
I'd probably defer landing this until we have some discussion as per https://github.com/w3ctag/design-principles/issues/436#issuecomment-1517352926.

index.bs
LeaVerou: Web IDL attributes are used for both (writable) data properties as well as accessors. My understanding of what this sentence was saying is that in WebIDL, to specify an accessor, you specify an attribute with separate getter and setter.

cynthia:
Just as a blunt question - is the concept of "attributes" confusing to non-web facing JS developers? I'd imagine that the concept itself is used in other languages to a point where most people would know what we are talking about with "attributes", but "accessors" are less common. Personally, I think the existing text is more clear.

(Silly anecdote, there are 229M results for "javascript attributes" vs 330K results for "javascript accessors" - but that includes a bunch of stuff related to HTML attributes accessed via JS)

LeaVerou:
"Attributes" is a natural language word as well, used for all sorts of things, whereas I believe accessors is a term of art specific to JS, so I'm not surprised at the huge disparity in search result numbers.
Also, I don't think the part that alienates is the word "attributes", but the whole phrase "WebIDL attributes".
That said, there's probably a better phrasing, I'll try to brainstorm something

index.bs
LeaVerou:
```suggestion
Any syntax that looks like JavaScript properties (such as accessors or [[!WEBIDL]] attributes)
```

@cynthia @hober what about that?

cynthia:
Closing for the time being (for the record, @cynthia and @hober were strongly opposed to this), to be revisited if we want to change the audience/direction of the principles.

PR #441, opened 2023-05-09
Give more guidance about how to monkeypatch well.

jyasskin:
This catches the things I've noticed people doing suboptimally, but I'm happy to change the advice or add more things.

@annevk @domenic, does this catch the things that most annoy you?

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/design-principles/pull/441.html#cant-avoid-monkey-patching" title="Last updated on Jun 5, 2023, 8:23 PM UTC (170af2e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/441/feac005...jyasskin:170af2e.html" title="Last updated on Jun 5, 2023, 8:23 PM UTC (170af2e)">Diff</a>


index.bs
```diff
@@ -2979,11 +2979,22 @@ As such, monkey patching (wrongly!) presupposes that underlying functionality ca
...
+    > 1. The first new step.
+
+    or
+
+    > Replace step 7.1 of [=HTTP-network fetch=], "If connection is failure, then return a network error." with "If connection is failure, do something else."
+ 1. Keep monkey patches short. If you're adding more than a couple of steps, define a separate, self-contained algorithm in your specification, and have the monkey patch call the new algorithm.
+ 1. The text of your monkey patch should be suitable to paste literally into the upstream algorithm. This implies that control flow like "return" or "abort" will return or abort from the whole upstream algorithm, not just the monkey patched section.
+ 1. File an issue against the existing specification that asks to merge your monkey patch. This will notify the maintainers of the upstream specification that there's a change incoming that might benefit from their review. They might even be able to tell you how to hook into an existing extension point or quickly create one so that you can remove your monkey patch sooner than you expected.
```
annevk, line 2995: I would say that this notifies the relevant community.

This is also a point where you might get pushback. Perhaps we should make that clear?

annevk (commented):
Generally looks good to me. Seems like a good place as well, especially compared to Infra.

index.bs
```diff
@@ -2979,11 +2979,22 @@ As such, monkey patching (wrongly!) presupposes that underlying functionality ca
...
-
- 1. File a bug, or create a fix, for the existing specification. This will help you understand what the problem is and how to fix it, and maybe even alleviate you from even creating the monkey patch in the first place.
- 1. In your spec, clearly mark the monkey patch as a temporary workaround (e.g., using a big red box) and make it clear to readers that you intend to add your patch to the existing specification. Make sure you link to the existing specification and the bug or fix that you filed.
- 1. Work together with the authors of the existing specification on a solution. If the existing specification is no longer maintained... Congratulations! you are now the editor of that existing specification. If necessary, add suitable extension points to the existing specification to better handle your changes.
+Sometimes monkey patching is unavoidable (e.g., while a new specification is incubating in a community group). If you find yourself needing to monkey patch an existing specification, make sure you:
```
LeaVerou, line 2982: It's not just that. E.g. in the CSS WG we often write specs this way in the very early stages where e.g. Level 3 is still being iterated on, but we're also working on a Level 4 draft.

ljharb:
I'm a little confused; the term "monkey patch" usually refers, ime, to runtime JS altering environment builtins. If it's here referring to specs modifying other specs, nearly every proposal does this, using ins/del tags in the ecmarkup, and it's the responsibility of proposal champions to coordinate with other proposals if their changes would conflict - and the first one to hit stage 4 "wins" in terms of who has to "rebase".

LeaVerou:
> If it's here referring to specs modifying other specs, 

It is

> nearly every proposal does this, using ins/del tags in the ecmarkup, and it's the responsibility of proposal champions to coordinate with other proposals if their changes would conflict - and the first one to hit stage 4 "wins" in terms of who has to "rebase".

This is about not allowing mature specs to do this. Proposals and drafts can still do it (though the current PR should be reworded to account for the CSS WG and TC39 processes around drafts and proposals).

ljharb:
Makes sense, thanks for clarifying.

Doesn't HTML do this to the 262 spec? ecma402 certainly does this to ecma262.

I can't speak for TC39 but I suspect we'd all *prefer* that other specs use only host hooks to integrate, but I'm not sure we consider it a "problem" warranting explicit guidelines (i think HTML may be one of the only non-tc39 specs that sits on top of 262, modulo C++'s regexes)

index.bs
jyasskin: I tried to phrase it positively as "This community may suggest better ways to accomplish your goals. Take them seriously."

index.bs
```diff
@@ -2979,11 +2979,22 @@ As such, monkey patching (wrongly!) presupposes that underlying functionality ca
...
+
+ 1. In your spec, clearly mark the monkey patch as a proposed change to another specification using language like:
+    > `# Monkey patch HTTP fetch`
+ 1. Identify the steps you want to change using both their number and a quote from their text. This allows you to update your monkey patch as the upstream specification changes.
```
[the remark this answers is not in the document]
jyasskin: Good point. I've reworded along those lines, but I don't yet have a good example of such a non-numbered monkey patch. If you think it's worth including one here, does one come to mind for you?

jyasskin (commented):
I'll need more details about how the CSSWG uses monkeypatches in order to integrate that working mode into this text. I'll ask @tabatkins for his take, but I'd also appreciate other people's explanations and especially suggested textual changes.

Thanks for the quick reviews!

index.bs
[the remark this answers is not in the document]
annevk: Going from one- to two-argument `has()` could qualify I suppose: https://github.com/whatwg/url/commit/bfb9157186c237078cb1ac4998607d88242abe35.

index.bs
```diff
@@ -2981,9 +2981,24 @@ Wikipedia also describes some <a href="https://en.wikipedia.org/wiki/Monkey_patc
...
- 1. In your spec, clearly mark the monkey patch as a temporary workaround (e.g., using a big red box) and make it clear to readers that you intend to add your patch to the existing specification. Make sure you link to the existing specification and the bug or fix that you filed.
- 1. Work together with the authors of the existing specification on a solution. If the existing specification is no longer maintained... Congratulations! you are now the editor of that existing specification. If necessary, add suitable extension points to the existing specification to better handle your changes.
+ 1. In your spec, clearly mark the monkey patch as a proposed change to another specification that is only temporarily in this specification, using language like:
+    ```markdown
+    # Modify HTTP fetch
+
+    <p class="issue">Merge this into [[Fetch]] once it has enough consensus.</p>
+    ```
+ 1. Identify the place you want to change using a quote from its text. This allows you to update your monkey patch as the upstream specification changes. Include the current step numbers if you're changing something that's numbered, but remember that numbers aren't sufficient by themselves.
```
fantasai, line 2990: > remember that numbers aren't sufficient by themselves.

Might want to explain why (because the numbers often change as the origin spec is edited to add/remove steps).

fantasai:
> I'll need more details about how the CSSWG uses monkeypatches

CSSWG doesn't often do this kind of detailed monkey-patching into the middle of an algorithm, but we do very often use delta specs as @LeaVerou mentions. Here's a few examples:

https://www.w3.org/TR/2019/WD-css-grid-2-20191203/
https://www.w3.org/TR/2019/WD-css-text-4-20191113/
https://www.w3.org/TR/2023/WD-scroll-animations-1-20230428/#timeline-ranges
https://www.w3.org/TR/2021/WD-css-fonts-5-20211221/
https://www.w3.org/TR/2022/WD-css-color-5-20220628/
https://www.w3.org/TR/2021/WD-css-sizing-4-20210520/

You'll notice none of these use numbered references or quoted text to identify patch insertion points, but rather identify the feature that's being extended and then explain how it's being extended... so they're not quite “monkey-patching”.

There's also been a few instances where we needed to modify CSS2 behavior, without rewriting it, and in those cases we described the changes but not as a patch, e.g. https://www.w3.org/TR/2023/WD-css-align-3-20230217/#justify-block

fantasai:
Wrt defining monkey-patches, it some cases it might be helpful to use diff markup (ins/del/blockquote) rather than describing the insertions and deletions in English, so I would suggest that should be allowed.

index.bs
jyasskin: Done, thanks.

index.bs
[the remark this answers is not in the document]
jyasskin: I've adapted the `URLSearchParams.has()` change into an [example](https://pr-preview.s3.amazonaws.com/jyasskin/design-principles/pull/441.html#example-monkeypatch-quote-text-change-idl), thanks! Let me know if it looks like what you'd want an actual spec to do.

jyasskin (commented):
Thanks for the CSS examples, @fantasai; they were really helpful! I think some of these, like where you're just adding new properties or values that are independent from the other values, seem more like uses of a ["publicly defined extension point"](https://pr-preview.s3.amazonaws.com/jyasskin/design-principles/pull/441.html#monkey-patching) and so not strictly monkeypatches. But I made examples out of the ["over-constrained"](https://pr-preview.s3.amazonaws.com/jyasskin/design-principles/pull/441.html#example-monkeypatch-quote-text-replace-def) change and [the format change to font-size-adjust](https://pr-preview.s3.amazonaws.com/jyasskin/design-principles/pull/441.html#example-monkeypatch-quote-text-change-css).

index.bs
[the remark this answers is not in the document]
annevk: Looks good for proposed IDL changes.

index.bs
```diff
@@ -2981,9 +2981,69 @@ Wikipedia also describes some <a href="https://en.wikipedia.org/wiki/Monkey_patc
...
+    <p class="issue">Merge this into [[Fetch]] once it has enough consensus.</p>
+    ```
+
+    </div>
+
+ 1. Identify the place you want to change using a quote from its text or by linking to a defined term if you're replacing that term's whole definition. Include the current step numbers if you're changing something that's numbered, but remember that numbers aren't sufficient by themselves because they change when other steps are added or removed. If you're changing several places within a definition, it can be helpful to paste the entire definition into a `<blockquote>` and use `<ins>` and `<del>` to describe what's changing.
+
+    <div class="example" id="example-monkeypatch-quote-text-before-step">
+
+    Before step 4.2 of [=HTTP fetch=], «If request's redirect mode is "follow", then set request's service-workers mode to "none".», insert the following steps:
```
domenic, line 3000: Use normal or curly-quotes instead of French (?)/Infra-list quotes?

index.bs
```diff
@@ -2981,9 +2981,69 @@ Wikipedia also describes some <a href="https://en.wikipedia.org/wiki/Monkey_patc
...
+      boolean has(USVString name<ins>, optional USVString value</ins>);
+    };
+    </pre>
+
+    </div>
+    <div class="example" id="example-monkeypatch-quote-text-change-css">
+
+    Modify the definition of <a spec="css-fonts-4" property>font-size-adjust</a> as follows:
+
+    <blockquote>
```
domenic, line 3025: Blockquote doesn't seem necessary here

domenic (approved):
Generally looks good, just found a few editorial nits!

index.bs
jyasskin: French, to avoid ambiguity with the quotes around the redirect modes. But they're not that ambiguous, so I can just use normal quotes. Done.

index.bs
jyasskin: Sure, removed.

jyasskin:
It looks like the non-TAG feedback has died down. How else should this PR change before it gets merged? Does the TAG have an ETA to review it?

LeaVerou:
> It looks like the non-TAG feedback has died down. How else should this PR change before it gets merged? Does the TAG have an ETA to review it?

Hi Jeffrey,
We usually work on Design Principles PRs during Design Principles week (once a month), but we can bump things up if there's any reason they are more urgent (even a weak reason is ok).

jyasskin:
No rush. I wondered about the timeline because I'm citing this to a new spec writer, but I can cite the PR as easily as the published document.

index.bs
```diff
@@ -2981,9 +2981,65 @@ Wikipedia also describes some <a href="https://en.wikipedia.org/wiki/Monkey_patc
...
+    Animation type: <ins>discrete if the keywords differ, otherwise </ins>by computed value type
+    </pre>
+
+    </div>
+
+ 1. Keep monkey patches short. If you're modifying an algorithm by adding more than a couple of steps, define a separate, self-contained algorithm in your specification, and have the monkey patch call the new algorithm.
+ 1. If you're replacing or adding steps in an algorithm, write the new steps so an editor can paste them verbatim into the upstream algorithm. This implies that control flow like "return" or "abort" will return or abort from the whole upstream algorithm, not just the monkey patched section.
+ 1. File an issue against the existing specification that asks to merge your monkey patch. This will notify the relevant community that there's a change incoming that might benefit from their review. This community may suggest better ways to accomplish your goals. Take them seriously. They might also be able to tell you how to hook into an existing extension point or quickly create one so that you can remove your monkey patch sooner than you expected.
```
LeaVerou, line 3040: This doesn’t really apply if the specification is at a very early stage (e.g. stage 1 TC39 proposals)

LeaVerou (changes requested):
@jyasskin we looked at this PR with the group today. We generally agree with everything, but the point I commented on gave us some pause. Ideally, the relevant community should be notified as early as possible, but opening an issue to merge a potentially unvetted and unreviewed monkey patch seems a little premature (e.g. imagine an issue being opened for every stage 0 TC39 proposal). We'd like to see these separated, and the guidance about opening the issue to use wording that makes it clear that this applies when the monkey patch has received enough vetting and consensus that it's ready to consider merging.

If you're able to make these changes by Wednesday, we may even be able to merge this week. If not, no worries!

jyasskin:
I've updated the item you were concerned about to say that the patch's community should review the proposal before filing an issue with the upstream specification, and that this issue should ask for review rather than asking for merging. I think we do want the issue filed before the patch has "enough vetting and consensus that it's ready to consider merging", but you're right that we don't want it filed as early as my initial text said. Step 7 ("When your work has enough support to merge"), then, becomes the point at which people would update the issue to request merging.

The text I have might be a bit confusing by mentioning two different communities (the patch's community, and the specification's community). I think it works well enough, but if y'all have ideas for better wording, I'm all ears.

index.bs
```diff
@@ -3037,7 +3037,7 @@ Sometimes monkey patching is unavoidable (e.g., early in the design phase of a n
...
- 1. File an issue against the existing specification that asks to merge your monkey patch. This will notify the relevant community that there's a change incoming that might benefit from their review. This community may suggest better ways to accomplish your goals. Take them seriously. They might also be able to tell you how to hook into an existing extension point or quickly create one so that you can remove your monkey patch sooner than you expected.
+ 1. Once your feature has been reviewed within our own community, and there seems to be consensus that it's a good idea, file an issue against the existing specification that asks the upstream community to review your monkey patch. This community may suggest better ways to accomplish your goals. Take them seriously. They might also be able to tell you how to hook into an existing extension point or quickly create one so that you can remove your monkey patch sooner than you expected.
```
bathos: was this meant to say “Once your feature has been reviewed within _your_ own”?

LeaVerou:
Looks good, we resolved to merge! Thank you!

jyasskin:
Thanks!

PR #452, opened 2023-09-07
fix caps in reference

ylafon:
Note that those should be harmonised


ylafon:
Closing it as duplicate of #449

PR #460, opened 2023-11-30
More detail on why events should fire before promises

jakearchibald:
***
<a href="https://pr-preview.s3.amazonaws.com/jakearchibald/design-principles/pull/460.html" title="Last updated on Apr 1, 2024, 8:51 PM UTC (7b95818)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/460/7fe598f...jakearchibald:7b95818.html" title="Last updated on Apr 1, 2024, 8:51 PM UTC (7b95818)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1477,7 +1477,7 @@ make that argument optional and specify the default value.
 
 <div class="example">
 {{EventTarget/addEventListener()}} takes an optional boolean `useCapture` argument.
-Thie defaults to `false`, meaning that
+This defaults to `false`, meaning that
 the event should be dispatched to the listener in the bubbling phase by default.
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -1805,7 +1805,7 @@ If a Promise-based asynchronous algorithm dispatches events,
 it should dispatch them before the Promise resolves,
 rather than after.
 
-When a promise is resolved, a microtask is queued to run its reaction callbacks.
+When a promise is resolved, a <a href=https://html.spec.whatwg.org/multipage/webappapis.html#microtask>microtask</a> is queued to run its reaction callbacks.
 Microtasks are processed when the JavaScript stack empties.
 <a href=https://dom.spec.whatwg.org/#dispatching-events>Dispatching an event</a> is synchronous, 
 which involves the JavaScript stack emptying between each listener.
--- a/index.bs
+++ b/index.bs
@@ -1807,7 +1807,8 @@ rather than after.
 
 When a promise is resolved, a microtask is queued to run its reaction callbacks.
 Microtasks are processed when the JavaScript stack empties.
-Dispatching an event can involve the JavaScript stack emptying between listeners.
+<a href=https://dom.spec.whatwg.org/#dispatching-events>Dispatching an event</a> is synchronous, 
+which involves the JavaScript stack emptying between each listener.
 As a result, if a promise is resolved before dispatching a related event,
 the promise reaction callbacks will be invoked between the first and second listeners of the event.
 
--- a/index.bs
+++ b/index.bs
@@ -1812,8 +1812,8 @@ which involves the JavaScript stack emptying between each listener.
 As a result, if a promise is resolved before dispatching a related event,
 the promise reaction callbacks will be invoked between the first and second listeners of the event.
 
-Dispatching the event first prevents this interleaving,
-ensuring that all event listeners are invoked before any of the related promise reaction callbacks.
+Dispatching the event first prevents this interleaving.
+All event listeners are then invoked before any promise reaction callbacks.
 
 <h3 id="dont-invent-event-like">Don’t invent your own event listener-like infrastructure</h3>
 
--- a/index.bs
+++ b/index.bs
@@ -1810,7 +1810,8 @@ Microtasks are processed when the JavaScript stack empties.
 <a href=https://dom.spec.whatwg.org/#dispatching-events>Dispatching an event</a> is synchronous, 
 which involves the JavaScript stack emptying between each listener.
 As a result, if a promise is resolved before dispatching a related event,
-the promise reaction callbacks will be invoked between the first and second listeners of the event.
+any microtasks that are scheduled in reaction to a promise 
+will be invoked between the first and second listeners of the event.
 
 Dispatching the event first prevents this interleaving.
 All event listeners are then invoked before any promise reaction callbacks.
--- a/index.bs
+++ b/index.bs
@@ -1477,7 +1477,7 @@ make that argument optional and specify the default value.
 
 <div class="example">
 {{EventTarget/addEventListener()}} takes an optional boolean `useCapture` argument.
-This defaults to `false`, meaning that
+Thie defaults to `false`, meaning that
 the event should be dispatched to the listener in the bubbling phase by default.
 </div>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1477,7 +1477,7 @@ make that argument optional and specify the default value.
 
 <div class="example">
 {{EventTarget/addEventListener()}} takes an optional boolean `useCapture` argument.
-This defaults to `false`, meaning that
+Thie defaults to `false`, meaning that
```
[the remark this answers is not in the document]
jakearchibald: Ugh. I edited the file on Github, and for some reason it decided to undo my change in https://github.com/w3ctag/design-principles/pull/459

jakearchibald, line 1480:
```suggestion
This defaults to `false`, meaning that
```

cynthia (approved):

index.bs
```diff
@@ -1799,17 +1799,20 @@ Follow the <a href="https://www.w3.org/2001/tag/doc/promises-guide#one-time-even
...
-This guarantees that once the Promise resolves,
-all effects of the algorithm have been applied.
-For example, if an author changes some state
-in reaction to an event which the Promise dispatches,
-they can be sure that all of the state is consistent if the Promise is resolved.
+When a promise is resolved, a microtask is queued to run its reaction callbacks.
+Microtasks are processed when the JavaScript stack empties.
+Dispatching an event can involve the JavaScript stack emptying between listeners.
+As a result, if a promise is resolved before dispatching a related event,
+the promise reaction callbacks will be invoked between the first and second listeners of the event.
```
martinthomson, line 1812: Hey @jakearchibald, sorry to perform necromancy on this one.  We think that it's a distinct improvement, but have some requests.

Is there a reference you could use for this, or an example?  This took me several reads to understand properly.  It's accurate, but the effect is surprising.

The gap, I think, is that while you explain how promises can end up interleaved with event handlers (is it a handler or a listener?) there is a gap regarding the execution model.  That doesn't need to be here, but a reference would help.

(Questions I couldn't answer is what the effect on bubbling events.  And maybe what model event listener dispatch takes.)

jakearchibald: I'm not sure what's missing, or what reference I'd use. Let me try and explain it in more detail, and you can tell me which bit(s), if any, give you an "aha!" moment, and I'll work that into the description.

Pending promise reaction callbacks are processed in microtasks. There are many places that microtasks are processed, but it's always when the JS stack is empty. The most-encountered place that microtasks are processed is https://html.spec.whatwg.org/multipage/webappapis.html#clean-up-after-running-script, which is called whenever the JavaScript stack empties. This algorithm gets called whenever script or callbacks are executed, eg https://webidl.spec.whatwg.org/#invoke-a-callback-function.

Imagine the following spec:

1. Resolve promise.
2. Dispatch event.
3. Log "hello".

Resolving a promise is async, so the promise callbacks aren't called as part of step 1.

However, dispatching an event is synchronous. _All_ the related callbacks for the event (listeners, handlers, capturing, bubbling) are called before we get to step 3.

Let's say there are two such callbacks to call as part of step 2. Each listener/handler callback is [invoked](https://dom.spec.whatwg.org/#concept-event-listener-inner-invoke), which calls [call a user object's operation](https://webidl.spec.whatwg.org/#call-a-user-objects-operation), which calls [clean up after running script](https://html.spec.whatwg.org/multipage/webappapis.html#clean-up-after-running-script) when each callback is complete.

Because [clean up after running script](https://html.spec.whatwg.org/multipage/webappapis.html#clean-up-after-running-script) causes microtasks to be processed, the order callbacks are called is:

1. The first event listener callback
2. All currently queued reaction callbacks for the promise
3. The second event listener callback

That seems bad/confusing.

If the spec was:

1. Dispatch event.
2. Resolve promise.
3. Log "hello".

Then all event callbacks will have been called before we get to step 2. That means the order of callbacks would be:

1. The first event listener callback
2. The second event listener callback
3. …potentially some other stuff…
4. All currently queued reaction callbacks for the promise

I've added "potentially some other stuff", since it's possible for other things to happen before the next microtask checkpoint. There's a microtask checkpoint at the end of every task. There are checkpoints in the rendering steps too. And again, whenever the JS stack empties.

martinthomson: Thanks for indulging me @jakearchibald. That is very helpful.

> dispatching an event is synchronous

This is the missing piece.  (To whoever thought this was a good idea: I hope you understand why this makes me very sad and feel at least a little remorse.  Of course I forgive you, because I understand that you don't always know better when you ship something.)

My expectation was that - if we can use threading terms - you would signal the main JS thread when you want to fire an event and it would be enqueued for processing.  That it is synchronous is a little horrifying, frankly, but I can see how it ended up that way.

annevk: @martinthomson To be clear, end user input is definitely scheduled in that way (see the event loop and task concepts in the HTML Standard). Event dispatch is a small "synchronous" part of that (and is defined in the DOM Standard).

jakearchibald: Yeah, I think we're running into sync vs async being kinda fluffy terms when you actually dig into them.

Some events are supposed to fire at the 'same time' because they're related to the same thing, like `input` and `change` on some inputs. This is only possible because dispatching events is synchronous. Although a task could be queued to make those calls, making it feel async depending on where you're standing.

`event.dispatchEvent` is definitely synchronous though, which is how you can tell if the event was cancelled in the next line.

index.bs
```diff
@@ -1799,17 +1799,20 @@ Follow the <a href="https://www.w3.org/2001/tag/doc/promises-guide#one-time-even
...
 If a Promise-based asynchronous algorithm dispatches events,
 it should dispatch them before the Promise resolves,
 rather than after.
 
-This guarantees that once the Promise resolves,
-all effects of the algorithm have been applied.
-For example, if an author changes some state
-in reaction to an event which the Promise dispatches,
-they can be sure that all of the state is consistent if the Promise is resolved.
+When a promise is resolved, a microtask is queued to run its reaction callbacks.
```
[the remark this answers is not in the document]
martinthomson: I think so.  Though when you say "all event-related callbacks" is that really the case?  Because the text then says that the second callback goes after the promise, because the promise reactions run when the stack drains.

(I guess I'm still not 100% on all of this.  A reference might still be necessary.)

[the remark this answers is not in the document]
jakearchibald: > Though when you say "all event-related callbacks" is that really the case?

Yes.

> Because the text then says that the second callback goes after the promise, because the promise reactions run when the stack drains.

Right, but that doesn't prevent running "all event-related callbacks" as part of the dispatch, it just additionally involves the promise reaction callbacks.

Dispatching the event involves calling callbacks, which means the JS stack populates and empties, which means microtasks run, which means promise callbacks run 

> (I guess I'm still not 100% on all of this. A reference might still be necessary.)

By reference, do you mean a link to another spec? Would any of the links I've provided so far work as this reference?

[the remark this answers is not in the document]
martinthomson: OK, so I'm following you now, but none of this detail is necessarily obvious.  Citing one of those documents might help fill in some details if people (like me) find themselves struggling.

Dispatching an event is synchronous.  All event callbacks are invoked in turn.  In comparison, resolving a promise creates a microtask for each callback.  Microtasks do not run immediately, but run when the stack drains.  Mixing these patterns creates a potentially surprising interaction.  Invoking any callback means populating and then draining the stack.  When the stack drains, any microtasks that are dependent on promises are invoked.  Thus, if an event is dispatched after a promise resolves, the first event callback will be invoked.  However, when the callback returns, the stack drains and then callbacks for the promise run as microtasks.  Those microtasks run before the second and subsequent event callbacks.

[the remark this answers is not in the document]
jakearchibald: > resolving a promise creates a microtask for each callback.

It creates a microtask to run all currently queued callbacks. All of the callbacks will run in the same microtask. Although, I'm not 100% sure that difference is observable.

> When the stack drains, any microtasks that are dependent on promises are invoked

Yes, but _all_ microtasks run, not just promise-related ones. Microtasks queued for, say, mutation observers, will also run.

I'll link up the definitions for dispatch and resolve when I get time (I don't have a bikeshed environment on my new machine, so I'll need to set that up)

LeaVerou:
Somewhat orthogonal to the actual text in the PR, I think it would be great to list a few examples of events and related promises to make what we're talking about more concrete. What are some popular examples of this in the web platform? (either where the guidance is already followed or where it is not)

LeaVerou (approved):

Issue #344, opened 2021-10-07
General principle around time measurement units/formats for various domains (JS, HTTP, HTML, etc)?


csharrison:
This was brought up in our discussion of a new HTML attribute in the Attribution Reporting API (https://github.com/WICG/conversion-measurement-api/issues/132)

The existing TAG principles (https://w3ctag.github.io/design-principles/#milliseconds) are focused on Javascript APIs. Should those principles be extended to all Web Platform APIs including things like HTML attributes, HTTP headers, etc?

annevk:
Probably not? At least HTTP uses delta-seconds, HTML uses date and time strings, and both make sense to me given the domain.

I'm not even sure if all APIs ought to use milliseconds. I believe media uses seconds (as floats) pretty consistently (at least in the media element APIs). @padenot might have thoughts here.

annevk:
Yeah, from other changes it does seem like the document is trying to cover the various domains so that would make sense.

cc @LeaVerou

kenchris:
We should definitely remove the text saying to always use milliseconds. It is quite common in DOM APIs though, like setTimeout, setInterval, WebAnimations, Vibrate etc

We also have a few polling APIs using frequency instead (like Generic Sensors).

Telemetry APIs like Performance.now() often use DOMHighResTimestamp which makes sense in that domain.

Maybe we can remove the milliseconds example and instead say to stay consistent with other APIs in the same domain, like the telemetry domain uses DOMHighResTimestamp, etc

LeaVerou:
We discussed this today in our virtual f2f. It doesn't appear there is one universal guidance here (like what is currently in the spec), APIs across the Web Platform do different things, and often for good reason.

- `HTMLMediaElement` APIs use seconds
- `window.performance` uses `DOMHighResTimeStamp`
- [Web Codecs use microseconds as a long](https://www.w3.org/TR/webcodecs/#dom-videoframe-timestamp)
- Web Animations use milliseconds
- ...

Therefore, the guidance needs to be more nuanced than a blanket "use milliseconds and `DOMHighResTimeStamp` that it is now, and possibly include a list of precedents, so that API Designers can choose what to be compatible with based on their use case.

LeaVerou:
@kenchris will take a stab at some initial text, and then we can iterate.

csharrison:
[...]

PR #467, opened 2024-01-24
Add principle on use of delta seconds

torgo:
Addresses #344...

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/467.html" title="Last updated on Jan 24, 2024, 10:06 AM UTC (d5a9561)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/467/0a66b86...d5a9561.html" title="Last updated on Jan 24, 2024, 10:06 AM UTC (d5a9561)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2334,6 +2334,21 @@ and there is no way to make them immutable.
         on the Web IDL Bugzilla
 </div>
 
+<h3 id="use-delta-seconds">Use Delta-Seconds over HTTP-Date</h3>
+
+For HTTP header-based APIs, using Delta-seconds is prefered over HTTP-Date for a few reasons:
+
+* Parsing is inconsistent, especially around error handling 
+(see <a href="https://cache-tests.fyi/#expires-parse">tests</a>)
+* parsing is relatively expensive
+* To adjust for clock sync, you need to have a Date header available (which
+means you need to parse two dates)
+* Because of that, potential drift is doubled to two seconds
+
+Delta-seconds are much more straightforward. Developers sometimes ask for sub second 
+resolution in HTTP headers, but this is typically not accommodated for in specs, 
+since the RTT for a request can easily make that meaningless at best and a footgun at worst.
+
 <h3 id="error-types">Use Error or DOMException for errors</h3>
 
 Represent errors in web APIs as ECMAScript error objects (e.g., {{Error}})


cynthia:
I think this should be moved into this section: "Use milliseconds for time measurement" as a conditional for cases where the time crosses boundaries that require interacting with HTTP headers.

cynthia (changes requested):
This isn't layered in the right place.

martinthomson (commented):
All of this is true, but I question whether this rises to a level at which it is necessary to capture here.

The HTTP working group maintains guidance on the use of that protocol and this is not in that envelope.

The underlying fact here is that not everyone has a clock that has the same date and time.  What is documented here are some of the consequences of that, but there are many other consequences too.  In particular, this whole business about clock sync is too much for me.

Is there a context in which this sort of guidance would have helped? That is, relative to the application of a bit of experience or common sense (which is what I see in #344).

I might instead say something like:

> ### Do not assume that everyone has the same view of the time
>
> Consider the potential for different entities to have divergent views of what the current time is.
> If your design uses time, implement safeguards against disagreements about the current time.
> 
> This is not about time zones, 
> though that problem can often be addressed by using UTC 
> in machine-to-machine interfaces. 
> Time zones only apply at the point that there is a need to present localized information.
> 
> The web is comprised of many computers,
> most of which have clocks,
> but very few of which agree about the current time.
> Some computers will have clocks that disagree by a lot.
> Though clocks might disagree on the current time,
> they disagree less about the rate at which time advances.
>
> Using differences in time relative to a common reference point - rather than absolute times -
> can help avoid problems arising from disagreements.
> Even then, include allowances for clocks that drift,
> especially when time passes.
>
> An approximate agreement about the current time can be critical to security,
> such as in determining certificate validity periods when authenticating servers.
> In such cases, agreeing about the current time cannot be avoided;
> being tolerant of errors is instead recommended.

I didn't add anything about the fact that even a single computer can have multiple time systems.  For instance, media processing logic - audio in particular - can use a different oscillator from the main system clock.  That's probably a detail that is specialized enough that we can rely on the people who need to know already knowing.

LeaVerou:
> The HTTP working group maintains guidance on the use of that protocol and this is not in that envelope.

Generally, we *do* have principles that are concrete guidance about specific web technologies. If the HTTP WG maintains a separate principles document, we should link to it, could you please share it?

martinthomson:
https://httpwg.org/specs/rfc9205.html

torgo:
as discussed in today's f2f session, we agreed not to include this at this time.

PR #470, opened 2024-01-25
Add info to 9.1 ("Use care when exposing identifying information about devices") & change link

### LINKED ISSUE #398: Add info to device apis section about privacy
As discussed in the privacy TF - make it clear that the app should not be able to determine whether the user has rejected use of a sensor API vs sensor not existing...

torgo:
Closes #398

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/470.html" title="Last updated on Mar 6, 2024, 8:18 PM UTC (f6bebfa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/470/c8a5914...f6bebfa.html" title="Last updated on Mar 6, 2024, 8:18 PM UTC (f6bebfa)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2413,7 +2413,7 @@ each increase the risk of harming the user's privacy.
 The web app should not be able to distinguish between: the user rejecting
 permission to use a sensor/capability; and that sensor/capability not being present.
 
-Another risk is that as more specific information is shared,
+As more specific information is shared,
 the set of
 [fingerprinting data](https://www.w3.org/TR/fingerprinting-guidance/)
 available to sites gets larger.
--- a/index.bs
+++ b/index.bs
@@ -2395,7 +2395,7 @@ So, these are called wrapper APIs.
 
 This section contains principles for consideration when designing APIs for devices.
 
-<h3 id="device-ids">Use care when exposing identifying information about devices</h3>
+<h3 id="device-ids">Don't expose unnecessary information about devices</h3>
 
 In line with the [Data Minimization](#data-minimization) principle, if you need to give web sites access to information about a device, only expose the minimal amount of data necessary.
 
--- a/index.bs
+++ b/index.bs
@@ -2410,8 +2410,8 @@ additional information about a device,
 or device identifiers,
 each increase the risk of harming the user's privacy.
 
-The web app should not be able to distinguish between: the user rejecting
-permission to use a sensor/capability; and that sensor/capability not being present.
+The web app should not be able to distinguish between the user rejecting
+permission to use a sensor/capability, and the sensor/capability not being present.
 
 As more specific information is shared,
 the set of
--- a/index.bs
+++ b/index.bs
@@ -2409,7 +2409,7 @@ additional information about a device,
 or device identifiers,
 each increase the risk of harming the user's privacy.
 
-The web app should not be able to distinguish between the user rejecting
+A web app should not be able to distinguish between the user rejecting
 permission to use a sensor/capability, and the sensor/capability not being present.
 
 As more specific information is shared,
--- a/index.bs
+++ b/index.bs
@@ -2397,8 +2397,7 @@ This section contains principles for consideration when designing APIs for devic
 
 <h3 id="device-ids">Use care when exposing identifying information about devices</h3>
 
-If you need to give web sites access to information about a device,
-use the guidelines below to decide what information to expose.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2411,15 +2410,16 @@ additional information about a device,
 or device identifiers,
 each increase the risk of harming the user's privacy.
 
-One risk is that as more specific information is shared,
+The web app should not be able to distinguish between: the user rejecting
+permission to use a sensor/capability; and that sensor/capability not being present.
```
rhiaro, line 2414:
```suggestion
The web app should not be able to distinguish between the user rejecting
permission to use a sensor/capability, and the sensor/capability not being present.
```

index.bs
```diff
@@ -2411,15 +2410,16 @@ additional information about a device,
...
+
+Another risk is that as more specific information is shared,
```
rhiaro, line 2416:
```suggestion
As more specific information is shared,
```

index.bs
cynthia: I feel like **sensor/capability** should be generalized to **device** given the principle.

cynthia (dismissed):
Aside from terminology jumping away from principle subject would likely confuse the reader, this looks good.

index.bs
torgo: We considered that but what we're really talking about is a capability on a device… e.g. bluetooth or ambient light sensor...

martinthomson: Isn't this a larger principle?  Namely where designs that depend on something from a user, where the user might reasonably deny that something, those designs should avoid leaking information about denial such that sites might retaliate in some way.  Designs can help with that by making denial indistinguishable from other reasons  that the something might not be available (like in this case, where the something might not even exist).  If that is not possible, then it might be appropriate to manufacture some base rate of failure.

matatk: @martinthomson: yes, it does seem like a solid larger principle; it [came up in our breakout](https://github.com/w3ctag/meetings/blob/gh-pages/2024/01-london/minutes.md#add-info-to-device-apis-section-about-privacy), and we actually found that the Geolocation API returns a specific result that the request was denied - looks like we should ask if that could be changed.

index.bs
```diff
@@ -2397,7 +2397,6 @@ This section contains principles for consideration when designing APIs for devic
 
 <h3 id="device-ids">Use care when exposing identifying information about devices</h3>
 
-
 If you need to give web sites access to information about a device,
 use the guidelines below to decide what information to expose.
```
torgo, line 2401:
```suggestion
In line with the [Data Minimization](#data-minimization) principle, if you need to give web sites access to information about a device, only expose the minimal amount of data necessary. The guidelines below can help make this determination.```

**Comment by torgo** [comment]
@cynthia would the wording change I've just left as a suggested change address the issue you've highlighted?

File: index.bs [hunk 4]
```diff
@@ -2411,15 +2410,16 @@ additional information about a device,
 or device identifiers,
 each increase the risk of harming the user's privacy.
 
-One risk is that as more specific information is shared,
+The web app should not be able to distinguish between the user rejecting
```
martinthomson, line 2413:
```suggestion
A web app should not be able to distinguish between the user rejecting
```

Do we have guidance around the usage of terms like "web app" and "[web]site"?  I have something of a strong preference for the latter, but acknowledge that there might be some sort of significant difference.

index.bs
```diff
@@ -2411,15 +2410,16 @@ additional information about a device,
...
+permission to use a sensor/capability, and the sensor/capability not being present.
+
+As more specific information is shared,
 the set of
```
martinthomson, line 2417:
```suggestion
the
```

index.bs
martinthomson:
```suggestion
In line with the [Data Minimization](#data-minimization) principle, if you need to give web sites access to information about a device, only expose the minimal amount of data necessary.
```

That last sentence seems a shade redundant (all of these sections provide guidelines that help).

index.bs
```diff
@@ -2397,9 +2397,7 @@ This section contains principles for consideration when designing APIs for devic
 
 <h3 id="device-ids">Use care when exposing identifying information about devices</h3>
```
martinthomson, line 2398:
```suggestion
<h3 id="device-ids">Don't expose unnecessary information about devices</h3>
```

hober (approved):

martinthomson (approved):

PR #471, opened 2024-01-25
Add guidance for factory method naming

### LINKED ISSUE #378: New principle: Naming of factory methods
This was brought up in our discussion of `Response.json()` in https://github.com/w3ctag/design-reviews/issues/741

We should have guidelines for naming factory methods. Currently the web platform is pretty inconsistent:

- `document.createXXX()` for creating nodes, also `document.createRange()`, `document.createTreeWalker()` etc
- the deprecated `document.initXXXEvent()` factory methods
- The [Temporal API](https://tc39.es/proposal-temporal/docs/) uses `from()` methods
- [`createImageBitmap()`](https://developer.mozilla.org/en-US/docs/Web/API/createImageBitmap) which is oddly placed in the global scope (?!)
- `Object.create()`
- `CanvasRenderingContext2D#createXXX()`
- `Array.from()`
- `Object.fromEntries()`
- `String.fromCharCode()`
- `audioContext.createXYZ()`
- `Response.json()`
- `DOMQuad.fromRect()`

Perhaps we should have guidance that factory methods should start with `create` or `from` to make it obvious that they are factory methods? Not sure we should restrict it further, as these are not interchangeable.

Though there are also things like `Date.parse()`, which are fine as-is…

cynthia:
Resolves #378.

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/471.html" title="Last updated on Jan 25, 2024, 4:26 PM UTC (63d0cf4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/471/c8a5914...63d0cf4.html" title="Last updated on Jan 25, 2024, 4:26 PM UTC (63d0cf4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3443,8 +3443,8 @@ converted to a target object.
 For example, `Foo.fromBar()` would imply
 that a `Foo` object will be created using a `Bar` object.
 
-Inventing other prefixes
-and usage of legacy prefixes should be avoided
+Avoid inventing other prefixes
+and using of legacy prefixes
 unless there is a strong reason to do so.
 Consistency with existing factory methods under the same object,
 such as `document.initXXX()` would be an example of an exemption.
--- a/index.bs
+++ b/index.bs
@@ -3434,7 +3434,7 @@ prefix the method name with `from`.
 
 Factory methods <a href="#constructors">are an exception</a>, and not the norm.
 An example of valid usage of a factory method is
-when an object is being created it also requires association
+when an object is being created also requires association
 with the parent object.
 `document.createXXX()` is an example of this.
 Use the prefix `from`
--- a/index.bs
+++ b/index.bs
@@ -3432,7 +3432,7 @@ prefix the method name with `create`.
 However, if your factory method creates an object from existing data,
 prefix the method name with `from`.
 
-Factory methods <a href="#constructors">are an exception</a>, and not the norm.
+Factory methods should be an exception, not the norm, and only used <a href="#constructors">for valid reasons</a>.
 An example of valid usage of a factory method is
 when an object is being created also requires association
 with the parent object
--- a/index.bs
+++ b/index.bs
@@ -3437,6 +3437,7 @@ An example of valid usage of a factory method is
 when an object is being created also requires association
 with the parent object
 (e.g. `document.createXXX()`).
+
 Use the prefix `from`
 when there is a source object expected to be
 converted to a target object.
--- a/index.bs
+++ b/index.bs
@@ -3425,7 +3425,7 @@ initialisms, even if they are repeated.
 
 <h4 id="factory-naming" class="no-num no-toc">Start factory method names with `create` or `from`</h4>
 
-When defining a new factory method, the name should be or start with `create` or `from`.
+Factory method names should start with `create` or `from`, optionally followed by a more specific noun. 
 
 If a factory method constructs a new empty object,
 prefix the method name with `create`.
--- a/index.bs
+++ b/index.bs
@@ -3423,6 +3423,33 @@ examples that violate the above rules are {{XMLHttpRequest}} and
 initialisms, even if they are repeated.
 </div>
 
+<h4 id="factory-naming" class="no-num no-toc">Start factory method names with `create` or `from`</h4>
+
+When defining a new factory method, prefix its name with either `create` or `from`.
+
+If a factory method constructs a new empty object,
+prefix the method name with `create`.
+However, if your factory method creates an object from existing data,
+prefix the method name with `from`.
+
+Factory methods <a href="#constructors">are an exception</a>, and not the norm.
+An example of valid usage of a factory method is
+when an object is being created it also requires association
+with the parent object.
+`document.createXXX()` is an example of this.
+The prefix `from` is expected to be used
+when there is a source object expected to be
+converted to a target object.
+For example, `Foo.fromBar()` would imply
+that a `Foo` object will be created using a `Bar` object.
+
+Inventing other prefixes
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3423,6 +3423,33 @@ examples that violate the above rules are {{XMLHttpRequest}} and
...
+When defining a new factory method, prefix its name with either `create` or `from`.
+
+If a factory method constructs a new empty object,
+prefix the method name with `create`.
+However, if your factory method creates an object from existing data,
+prefix the method name with `from`.
+
+Factory methods <a href="#constructors">are an exception</a>, and not the norm.
+An example of valid usage of a factory method is
+when an object is being created it also requires association
```
cynthia, line 3437:
```suggestion
when an object is being created also requires association
```

index.bs
```diff
@@ -3423,6 +3423,33 @@ examples that violate the above rules are {{XMLHttpRequest}} and
...
+when there is a source object expected to be
+converted to a target object.
+For example, `Foo.fromBar()` would imply
+that a `Foo` object will be created using a `Bar` object.
+
+Inventing other prefixes
+and usage of legacy prefixes should be avoided
+unless there is a strong reason to do so.
+Consistency with existing factory methods under the same object,
+such as `document.initXXX()` would be an example of an exemption.
```
rhiaro, line 3450:
```suggestion
A reason to make an exception would be to maintain consistency with existing factory methods
under the same object, such as `document.initXXX()`.
```

index.bs
```diff
@@ -3423,6 +3423,33 @@ examples that violate the above rules are {{XMLHttpRequest}} and
...
+If a factory method constructs a new empty object,
+prefix the method name with `create`.
+However, if your factory method creates an object from existing data,
+prefix the method name with `from`.
+
+Factory methods <a href="#constructors">are an exception</a>, and not the norm.
+An example of valid usage of a factory method is
+when an object is being created it also requires association
+with the parent object.
+`document.createXXX()` is an example of this.
```
LeaVerou: I think this is actually an antipattern that we should actively discourage. This should have been `Element.create({ document })`, because:
1. It’s a pattern that does not scale: what happens when you need to associate with two objects?
2. It does not make it clear what object is created. With factory methods being static methods on a class, even when a subclass is returned, at least we know what the base class is, whereas something like `document.createXXX()` could return anything.
3. It introduces an undesirable coupling between two classes when all we want is a relationship. E.g. what if in the future we want elements that are not associated with a document? 
4. It’s violates most precedent. 

Furthermore, the way this is phrased is inaccurate: for DOM elements the parent object is `element.parentNode`, not `document`. Which highlights another problem with this pattern.

index.bs
```diff
@@ -3423,6 +3423,33 @@ examples that violate the above rules are {{XMLHttpRequest}} and
 initialisms, even if they are repeated.
 </div>
 
+<h4 id="factory-naming" class="no-num no-toc">Start factory method names with `create` or `from`</h4>
+
+When defining a new factory method, prefix its name with either `create` or `from`.
```
LeaVerou, line 3428: There are also many cases where `create()` or `from()` is not a prefix but the entire name of the function, e.g. `Object.create()`, `Temporal.from()` etc. We should probably acknowledge them.

index.bs
```diff
@@ -3423,6 +3423,34 @@ examples that violate the above rules are {{XMLHttpRequest}} and
...
```
cynthia, line 3428:
```suggestion
When defining a new factory method, the name should be or start with `create` or `from`.
```

index.bs
cynthia: See suggestion below.

index.bs
```diff
@@ -3423,6 +3423,34 @@ examples that violate the above rules are {{XMLHttpRequest}} and
 initialisms, even if they are repeated.
 </div>
 
+<h4 id="factory-naming" class="no-num no-toc">Start factory method names with `create` or `from`</h4>
+
+When defining a new factory method, the name should be or start with `create` or `from`.
```
LeaVerou, line 3428:
```suggestion
When defining a new factory method, the name should be prefixed with `create` or `from`. The suffix can be empty, i.e. it is common to name generic factory methods `create()` or `from()`.
```

index.bs
```diff
@@ -3423,6 +3423,34 @@ examples that violate the above rules are {{XMLHttpRequest}} and
 initialisms, even if they are repeated.
 </div>
 
+<h4 id="factory-naming" class="no-num no-toc">Start factory method names with `create` or `from`</h4>
```
LeaVerou, line 3426: Is it a good idea for this to live under naming? What if we want to add guidance about what object these methods should live on?

index.bs
cynthia: How about
> Factory method names should start with `create` or `from` (The suffix can be empty, e.g., create()).

index.bs
```diff
@@ -3423,6 +3423,34 @@ examples that violate the above rules are {{XMLHttpRequest}} and
...
+prefix the method name with `create`.
+However, if your factory method creates an object from existing data,
+prefix the method name with `from`.
+
+Factory methods should be an exception, not the norm, and only used <a href="#constructors">for valid reasons</a>.
+An example of valid usage of a factory method is
+when an object is being created also requires association
+with the parent object
+(e.g. `document.createXXX()`).
+Use the prefix `from`
```
LeaVerou, line 3440:
```suggestion

Use the prefix `from`
```

LeaVerou (changes requested):

LeaVerou (approved):

PR #476, opened 2024-03-06
Denying consent is better if undetectable as such

### LINKED ISSUE #475: New: Avoid revealing that consent was denied
> Isn't this a larger principle?  Namely where designs that depend on something from a user, where the user might reasonably deny that something, those designs should avoid leaking information about denial such that sites might retaliate in some way.  Designs can help with that by making denial indistinguishable from other reasons  that the something might not be available (like in this case, where the something might not even exist).  If that is not possible, then it might be appropriate to manufacture some base rate of failure.

_Originally posted by @martinthomson in https://github.com/w3ctag/design-principles/pull/470#discussion_r1467099291_

martinthomson:
This is not a full generalization of the concepts that discussed in #470 and #475, but I think that it suffices.

Closes #475.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -218,6 +218,11 @@ of a new feature must justify the additional burden on users
 to decide whether to grant permission for each feature
 whenever it's requested by a Web page.
 
+Refusal is most effective if the site cannot
+distinguish refusal from other, common situations.
+This can make it more difficult for a site to
+pressure users to grant consent.
+
 <p class="example">
 For example,
 the [Geolocation API](https://www.w3.org/TR/geolocation-API/)


index.bs
```diff
@@ -559,11 +564,19 @@ in the language where it's used
...
+for the feature.
+If there is a user need or design principle which would fail
 if feature detection were available for the feature,
 then you should not support feature detection.
 
+Detecting the availability of a feature does not imply
+detecting whether <a href="#consent">consent</a> to use the feature
+has been granted.
+Generally, detecting whether the feature is implemented
+is separable from determining whether use of the feature has been authorized.
```
rhiaro, line 576:
```suggestion
can be done separately from determining whether use of the feature has been authorized.
```

rhiaro (approved):
Small editorial suggestion for simpler language, but otherwise this is a good addition.

This does seem like something that should at least be able to reference the Privacy Principles, but I haven't found the appropriate section after a quick look just now.

plinss (approved):

rhiaro:
Related [in privacy principles](https://w3ctag.github.io/privacy-principles/#non-retaliation)

Issue #463, opened 2024-01-20
New principle: APIs which vend byte buffers should return a Uint8Array


bakkot:
[This already seems to be a principle](https://github.com/whatwg/fetch/issues/1724#issuecomment-1814626681). See some more discussion [here](https://matrixlogs.bakkot.com/WHATWG/2024-01-18#L12-L0). 

ArrayBuffer is almost never what you want because you can't read from one directly. You have to create a view such as a Uint8Array to actually read it.

Almost all APIs which accept an ArrayBuffer also accept a Uint8Array, and some APIs [require a Uint8Array](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder/encodeInto), so it's rare for a Uint8Array to be less convenient than an ArrayBuffer.

This is currently somewhat inconsistent: some APIs (like [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response/arrayBuffer)) provide an ArrayBuffer, while others (like [TextEncoder](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder/encode)) provide a Uint8Array. Having this written down as a principle could help with that.

I'm not sure what the process here is, but if this is accepted I'm happy to send a PR adding this.

plinss:
Discussed this in a breakout today with @hober and @rhiaro. We'd be grateful for you to write a PR.

One nuance we though of is that there may be times when an ArrayBuffer is actually appropriate. In the case where the returned data is really just bytes, then we agree a Uint8Array is the correct choice, but if the data can be reasonably interpreted as other types then an ArrayBuffer is likely more appropriate. @annevk thoughts?

We also note that the ergonomics of converting an ArrayBuffer to a Uint8Array is simpler than converting a Uint8Array to a Int32Array, for example.

bakkot:
> if the data can be reasonably interpreted as other types then an ArrayBuffer is likely more appropriate

If the method actually returns int32s, then using an Int32Array would be appropriate. But it's hard to think of a time when an _ArrayBuffer_ specifically would be appropriate: an ArrayBuffer includes no interpretation of the bytes. If there's some interpretation which is not one of the TypedArray types, then you're almost always going to need to read the bytes out individually so you can interpret them, and that's easiest with Uint8Array.

[...]

PR #480, opened 2024-03-11
New principle: vend byte arrays as Uint8Arrays

bakkot:
Fixes https://github.com/w3ctag/design-principles/issues/463.

***
<a href="https://pr-preview.s3.amazonaws.com/bakkot/design-principles/pull/480.html" title="Last updated on Apr 3, 2024, 1:35 AM UTC (ad59b8b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/480/222226d...bakkot:ad59b8b.html" title="Last updated on Apr 3, 2024, 1:35 AM UTC (ad59b8b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1856,6 +1856,19 @@ as the [Promise integration](https://github.com/WebAssembly/js-promise-integrati
 is still under development as of today.
 </div>
 
+<h3 id="uint8array">Vend byte arrays as Uint8Arrays</h3>
+
+If an API returns a byte array, make it a Uint8Array, not an ArrayBuffer.
+
+ArrayBuffers cannot be read from directly;
+the developer would have to create a view such as a Uint8Array
+to actually read data out an ArrayBuffer.
+So provide them with the Uint8Array in the first place.
+
+If the bytes in the buffer have a natural intepretation
+as one of the other TypedArray types, provide that instead.
+For example, if the bytes represent Float32 values, use a Float32Array.
+
 <h2 id="event-design">Event Design</h2>
 
 <h3 id="one-time-events">Use promises for one time events</h3>
--- a/index.bs
+++ b/index.bs
@@ -1858,16 +1858,16 @@ is still under development as of today.
 
 <h3 id="uint8array">Output an array of bytes with Uint8Array</h3>
 
-If an API returns a byte array, make it a Uint8Array, not an ArrayBuffer.
+If an API returns a byte array, make it a {{Uint8Array}}, not an {{ArrayBuffer}}.
 
-ArrayBuffers cannot be read from directly;
+`ArrayBuffer`s cannot be read from directly;
 the developer would have to create a view such as a Uint8Array
-to actually read data out an ArrayBuffer.
-So provide them with the Uint8Array in the first place.
+to read data.
+Providing a `Uint8Array` avoids that additional effort.
 
 If the bytes in the buffer have a natural intepretation
 as one of the other TypedArray types, provide that instead.
-For example, if the bytes represent Float32 values, use a Float32Array.
+For example, if the bytes represent Float32 values, use a {{Float32Array}}.
 
 <h2 id="event-design">Event Design</h2>
 
--- a/index.bs
+++ b/index.bs
@@ -1856,7 +1856,7 @@ as the [Promise integration](https://github.com/WebAssembly/js-promise-integrati
 is still under development as of today.
 </div>
 
-<h3 id="uint8array">Vend byte arrays as Uint8Arrays</h3>
+<h3 id="uint8array">Output an array of bytes with Uint8Array</h3>
 
 If an API returns a byte array, make it a Uint8Array, not an ArrayBuffer.
 


index.bs
```diff
@@ -1856,6 +1856,19 @@ as the [Promise integration](https://github.com/WebAssembly/js-promise-integrati
 is still under development as of today.
 </div>
 
+<h3 id="uint8array">Vend byte arrays as Uint8Arrays</h3>
```
martinthomson, line 1859: This is good advice, but I'd like to see the reciprocal as well, which is that APIs should take `ArrayBufferView` as input, not any specific type.

Nit:
```suggestion
<h3 id="uint8array">Output an array of bytes with Uint8Array</h3>
```

bakkot: > This is good advice, but I'd like to see the reciprocal as well, which is that APIs should take `ArrayBufferView` as input, not any specific type.

At least one API ([encodeInto](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder/encodeInto)) takes a Uint8Array specifically, and I don't know that I agree it would be better for it to take an `ArrayBufferView`. It seems like a confusion of types for a user to pass a `Uint16Array` there, for example. Maybe that only applies when taking an input as a write target rather than just a source.

Also, wouldn't [BufferSource](https://webidl.spec.whatwg.org/#BufferSource) be more appropriate? [It](https://dontcallmedom.github.io/webidlpedia/names/BufferSource.html) is more common than [ArrayBufferView](https://dontcallmedom.github.io/webidlpedia/names/ArrayBufferView.html), I'm pretty sure.

martinthomson: My mistake, BufferSource was the one I was looking for.

encodeInto has an in-out argument, so perhaps it is a poor example of an input argument.

I'm interested in whether you think that people should be permitted to supply data in a form that best suits them.  it seems like `ArrayBufferView` endianness issues might cause you to prefer something less likely to tempt that sort of concern, but I see that things like Web Crypto use it happily (noting that I'm quite receptive to arguments of the form "Web Crypto is not an API you should model behaviour on").

bakkot: Were I designing web APIs from scratch, I would take only the narrowest appropriate TypedArray type, and in the case of byte-taking APIs specifically I'd take Uint8Arrays and ArrayBuffers (and no other types), since those both unambiguously represent a sequence of bytes, whereas (e.g.) a Uint16Array does not unambiguously represent a sequence of bytes.

Even leaving aside the issue of endianness, one can store byte values in a Uint16Array, and passing such a thing to a byte-taking API is not going to do what the user presumably expected. Requiring the user to explicitly do either `new Uint8Array(uint16ArrayInstance)` or `new Uint8Array(uint16ArrayInstance.buffer, uint16ArrayInstance.byteOffset, uint16ArrayInstance.byteLength)` (depending on which of those two behaviors they actually wanted) would be better. More generally, I think we'd all be better off if web platform APIs were much stricter about their input types (and [I'm trying to move JS, at least, in that direction](https://docs.google.com/presentation/d/1AFzFeVtbUCpPcMXTER0Zzb5l5c5oPdXCF4Yi_9B1EEM/edit#slide=id.g2557c67b38a_0_39)).

But I'm not designing web APIs from scratch, and in the world we actually live in, BufferSource is pretty widely used. So I'm neutral on this. I don't really want to be the one to codify "accept BufferSource" as a principle, though.

annevk: I guess there is a question as to whether we want to tackle this incrementally or offer advice for a variety of things. Shared memory also comes to mind as something that might need some advice.

And yeah, I think accepting BufferSource is probably the way to go, unless there's a very compelling reason not to have parity with the vast majority of existing APIs.

bakkot: Resizable buffers too, come to that.

It's a bit hard to offer advice on those right now because it seems like web platform hasn't entirely figured out whether APIs should take BufferSources backed by shared / resizable buffers. Personally I think almost all places should take them, but `[AllowResizable]` and `[AllowShared]` are still default-off right now (IIUC just because that was the safe option) and few (if any?) places have been updated to support them.

I suppose starting with a principle that shared/resizable buffers should be accepted could be a way to start moving in that direction. But I think it would be best to do that in a different issue/PR, rather than holding up this one until that's done.

index.bs
```diff
@@ -1856,6 +1856,19 @@ as the [Promise integration](https://github.com/WebAssembly/js-promise-integrati
 is still under development as of today.
 </div>
 
+<h3 id="uint8array">Output an array of bytes with Uint8Array</h3>
+
+If an API returns a byte array, make it a Uint8Array, not an ArrayBuffer.
+
+ArrayBuffers cannot be read from directly;
+the developer would have to create a view such as a Uint8Array
+to actually read data out an ArrayBuffer.
```
martinthomson, line 1865:
```suggestion
to read data.
```

index.bs
```diff
@@ -1856,6 +1856,19 @@ as the [Promise integration](https://github.com/WebAssembly/js-promise-integrati
...
+So provide them with the Uint8Array in the first place.
```
martinthomson, line 1866:
```suggestion
Providing a Uint8Array avoids that additional effort.
```

index.bs
```diff
@@ -1856,6 +1856,19 @@ as the [Promise integration](https://github.com/WebAssembly/js-promise-integrati
 is still under development as of today.
 </div>
 
+<h3 id="uint8array">Output an array of bytes with Uint8Array</h3>
+
+If an API returns a byte array, make it a Uint8Array, not an ArrayBuffer.
```
martinthomson, line 1861:
```suggestion
If an API returns a byte array, make it a {{Uint8Array}}, not an {{ArrayBuffer}}.
```

index.bs
```diff
@@ -1856,6 +1856,19 @@ as the [Promise integration](https://github.com/WebAssembly/js-promise-integrati
...
+
+ArrayBuffers cannot be read from directly;
```
martinthomson, line 1863:
```suggestion
`ArrayBuffer`s cannot be read from directly;
```

index.bs
```diff
@@ -1856,6 +1856,19 @@ as the [Promise integration](https://github.com/WebAssembly/js-promise-integrati
...
+the developer would have to create a view such as a Uint8Array
+to actually read data out an ArrayBuffer.
+So provide them with the Uint8Array in the first place.
+
+If the bytes in the buffer have a natural intepretation
+as one of the other TypedArray types, provide that instead.
+For example, if the bytes represent Float32 values, use a Float32Array.
```
martinthomson, line 1870:
```suggestion
For example, if the bytes represent Float32 values, use a {{Float32Array}}.
```

torgo (approved):
lgtm

PR #487, opened 2024-04-04
Update principle on HTTP usage

martinthomson:
The existing one is dated.  Most of that advice is now in RFC 9205.

There is also more general advice in that document, so the guidance here gets more complex as a result.

***
<a href="https://pr-preview.s3.amazonaws.com/martinthomson/design-principles/pull/487.html" title="Last updated on Apr 5, 2024, 10:31 AM UTC (7fc9d5b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/487/12d02ab...martinthomson:7fc9d5b.html" title="Last updated on Apr 5, 2024, 10:31 AM UTC (7fc9d5b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1872,10 +1872,10 @@ rather than after.
 
 When a promise is resolved, a <a href=https://html.spec.whatwg.org/multipage/webappapis.html#microtask>microtask</a> is queued to run its reaction callbacks.
 Microtasks are processed when the JavaScript stack empties.
-<a href=https://dom.spec.whatwg.org/#dispatching-events>Dispatching an event</a> is synchronous, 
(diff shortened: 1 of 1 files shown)


annevk (commented):
I think we should continue having the advice that new headers should use structured fields.

martinthomson:
That's included by reference.  See https://datatracker.ietf.org/doc/html/rfc9205#section-4.7 I don't want to duplicate that advice here.

annevk:
I wouldn't mind redundancy on that particular requirement as for pretty much every new header that's come up in the last couple years folks had to be reminded, but okay.

LeaVerou (approved):

torgo (approved):
lgtm

PR #488, opened 2024-04-04
Editorial updates to media types section

martinthomson:
This section was looking a bit dated.

* We no longer use the term "MIME type" but prefer "media type"
* Some basic grammatical fixes
* Updating and adding links
* Explain what "MIME sniffing" does (because of the above change)
* Correct a typo in `Writ[e]ableStream`

***
<a href="https://pr-preview.s3.amazonaws.com/martinthomson/design-principles/pull/488.html" title="Last updated on Apr 4, 2024, 4:53 AM UTC (293083f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/488/ff46d8f...martinthomson:293083f.html" title="Last updated on Apr 4, 2024, 4:53 AM UTC (293083f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -925,7 +925,7 @@ the functionality you are adding is **not** similar to that of the existing attr
     The `type` attribute is used on the <{input}> and <{button}> elements
     to further specialize the element type,
     whereas on every other element (e.g. <{link}>, <{script}>, <{style}>)
-    it specifies MIME type.
+    it specifies a media type.
     This is an antipattern; one of these groups of attributes should have had a different name.
 </div>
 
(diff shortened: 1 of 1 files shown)


annevk:
This is wrong. The web platform uses the term MIME type. And in particular we use https://mimesniff.spec.whatwg.org/#understanding-mime-types.

LeaVerou (approved):

martinthomson:
It is equally wrong to use the wrong terminology.  The IETF uses "media type" for this and consistency with that is more important than avoiding a collision with a use of the same term in a different domain.  I don't expect that you will agree, but I think that the right thing to do is acknowledge the collision and rely on context to disambiguate (as we do for many things in language).

annevk:
It seems pretty bad for the TAG to contradict the entire set of web platform specifications. Decisions like that will certainly make me less inclined to refer to its documents in the future.

A reasonable discussion can be had and perhaps if there's a motivated volunteer a change could be made, but trying to force it through like this is rather meh and uninspiring.

PR #496, opened 2024-06-03
New principle: A Promise represents completion or a value, not a callback (#342)

### LINKED ISSUE #342: New principle: Support common promise practices
**TL;DR:** Avoid limiting or specializing APIs to _"same microtask"_.

This came up in https://github.com/w3c/mediacapture-screen-share/issues/190#issuecomment-929491274 in the context of trying to specify a new platform API that should _only_ work immediately following the resolution of a promise from another platform API. A solution was proposed to limit the new API to the same microtask as the promise resolution callback, but this would have broken some fairly common practices around asynchronous programming with promises in ways that might be surprising to most people.

It would be nice to write down these common practices, and say "don't break them":

 1. Encapsulation — wrapping an async platform call in an `async function`
 1. Error logging — of a final async step (s/`return promise`/`try { return await promise } catch(e) { log.add(e) }`/)
 1. Shimming — of an async platform API in a JS library
 1. Promise branching — Multiple `then`s on the same promise `p` queue a microtask each as I understand
 2. Timeouts — `Promise.race([foo(), new Promise(r => setTimeout(r, 10000)])`

These are helpful techniques that seem benign. What they have in common is that they may trigger a microtask checkpoint, which isn't terribly obvious. I think most people would expect these techniques to work across platform APIs, so it would be nice if we could preserve that expectation. cc @annevk

jan-ivar:
Fixes #342.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/design-principles/pull/496.html#allow-microtask-switches" title="Last updated on Dec 4, 2024, 11:17 AM UTC (8f15631)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/496/55a06e0...jan-ivar:8f15631.html" title="Last updated on Dec 4, 2024, 11:17 AM UTC (8f15631)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1884,8 +1884,8 @@ a promise resolution or rejection callback.
 
 Authors tend to expect that any `async` function can be wrapped
 in another that appends synchronous code at the end.
-This might be so that a return value can be examined, such as using try/catch.
-Using `await` or similar wrappers queues a microtask:
+This might be to examine a resolved value or rejection using try/catch.
+But using `await` or similar wrappers queue a microtask:
 <div class="example">
 ```js
 async function fooWrapper() {
@@ -1901,13 +1901,25 @@ async function fooWrapper() {
 ```
 </div>
 
-To support this pattern, a specification that wishes to restrict `bar()` to
-only succeed if called immediately following a promise resolving can
-minimally limit it to the same task, but not the same microtask.
+In general, the settled result from an asynchronous function
+should be usable at any time.
+
+However, there could be cases where APIs might need to restrict
+when a result can be acted on.
+
+In the above example, perhaps the `platform.foo()` function
+establishes state that has real-time dependencies such that calling
+`bar()` is not viable at any future time.
+
+Never limit the viability of acting on a settled result to a microtask,
+as this prevents most forms of code composition.
+
+If possible, set a short timeout interval. Failing that, viability can
+be limited to the same task, but not a single microtask.
 
 <div class="note">
 One case where this came up was the [captureController.setFocusBehavior()](https://w3c.github.io/mediacapture-screen-share/#dom-capturecontroller-setfocusbehavior)
-method, where timing concerns were solved by adding a timeout on top of
+method, where timing concerns were instead solved by adding a timeout on top of
 the requirement of it being called on the same task that resolves the
 `getDisplayMedia()` promise.
 </div>
--- a/index.bs
+++ b/index.bs
@@ -1922,7 +1922,7 @@ be limited to the same task, but not a single microtask.
 One case where this came up was the [captureController.setFocusBehavior()](https://w3c.github.io/mediacapture-screen-share/#dom-capturecontroller-setfocusbehavior)
 method which controls whether a window the user selects to screen-capture
 should immediately be pushed to the front or not. It can be called as late
-as right <em>after</em> the `getDisplayMedia()` promise. This allows
+as right <em>after</em> the resolution of the `getDisplayMedia()` promise. This allows
 applications to make a different decision based on the window the user chose.
 But the application must act right away or this becomes surprising to the user.
 
--- a/index.bs
+++ b/index.bs
@@ -1883,8 +1883,9 @@ Avoid requiring that something needs to happen synchronously within
 a promise resolution or rejection callback.
 
 Authors tend to expect that any `async` function can be wrapped
-in another that appends synchronous code at the end e.g. to inspect state
-using try/catch. But the `await` queues a microtask:
+in another that appends synchronous code at the end.
+This might be so that a return value can be examined, such as using try/catch.
+Using `await` or similar wrappers queues a microtask:
 <div class="example">
 ```js
 async function fooWrapper() {
--- a/index.bs
+++ b/index.bs
@@ -1882,7 +1882,7 @@ For example, if the bytes represent Float32 values, use a {{Float32Array}}.
 Avoid requiring that something needs to happen synchronously within
 a promise resolution or rejection callback.
 
-Authors take for granted that an `async` function can be wrapped
+Authors tend to expect that any `async` function can be wrapped
 in another that appends synchronous code at the end e.g. to inspect state
 using try/catch. But the `await` queues a microtask:
 <div class="example">
@@ -1900,12 +1900,15 @@ async function fooWrapper() {
 ```
 </div>
 
-To not interfere with this pattern, the lowest recommended restriction is "same task".
+To support this pattern, a specification that wishes to restrict `bar()` to
+only succeed if called immediately following a promise resolving can
+minimally limit it to the same task, but not the same microtask.
 
 <div class="note">
 One case where this came up was the [captureController.setFocusBehavior()](https://w3c.github.io/mediacapture-screen-share/#dom-capturecontroller-setfocusbehavior)
-method, where timing concerns were solved by adding an timeout on top of
-the same task requirement.
+method, where timing concerns were solved by adding a timeout on top of
+the requirement of it being called on the same task that resolves the
+`getDisplayMedia()` promise.
 </div>
 
 <h2 id="event-design">Event Design</h2>
--- a/index.bs
+++ b/index.bs
@@ -1877,11 +1877,12 @@ If the bytes in the buffer have a natural intepretation
 as one of the other TypedArray types, provide that instead.
 For example, if the bytes represent Float32 values, use a {{Float32Array}}.
 
-<h3 id="allow-microtask-switches">A Promise represents a value, not a callback</h3>
+<h3 id="allow-microtask-switches">A Promise represents completion or a value, not a callback</h3>
 
 Avoid requiring that something needs to happen synchronously within
 a promise resolution or rejection callback.
 
+This breaks basic assumptions around asynchronous APIs being wrappable.
 Authors tend to expect that any `async` function can be wrapped
 in another that appends synchronous code at the end.
 This might be to examine a resolved value or rejection using try/catch.
@@ -1904,8 +1905,8 @@ async function fooWrapper() {
 In general, the settled result from an asynchronous function
 should be usable at any time.
 
-However, there could be cases where APIs might need to restrict
-when a result can be acted on.
+However, if the result of your API is time sensitive, you might need
+to restrict when it can be acted on.
 
 In the above example, perhaps the `platform.foo()` function
 establishes state that has real-time dependencies such that calling
@@ -1919,9 +1920,14 @@ be limited to the same task, but not a single microtask.
 
 <div class="example">
 One case where this came up was the [captureController.setFocusBehavior()](https://w3c.github.io/mediacapture-screen-share/#dom-capturecontroller-setfocusbehavior)
-method, where timing concerns were instead solved by adding a timeout on top of
-the requirement of it being called on the same task that resolves the
-`getDisplayMedia()` promise.
+method which controls whether a window the user selects to screen-capture
+should immediately be pushed to the front or not. It can be called as late
+as right <em>after</em> the `getDisplayMedia()` promise. This allows
+applications to make a different decision based on the window the user chose.
+But the application must act right away or this becomes surprising to the user.
+
+This solution was to add a timeout on top of the requirement of it being
+called on the same task that resolves the `getDisplayMedia()` promise.
 </div>
 
 <h2 id="event-design">Event Design</h2>
--- a/index.bs
+++ b/index.bs
@@ -1930,6 +1930,9 @@ The solution was to add a timeout on top of the requirement of it being
 called on the same task that resolves the `getDisplayMedia()` promise.
 </div>
 
+If an API depends on setting up temporary conditions then invoking the caller,
+that is a good reason to use a callback rather than a promise.
+
 <h2 id="event-design">Event Design</h2>
 
 <h3 id="one-time-events">Use promises for one time events</h3>
--- a/index.bs
+++ b/index.bs
@@ -1926,7 +1926,7 @@ as right <em>after</em> the `getDisplayMedia()` promise. This allows
 applications to make a different decision based on the window the user chose.
 But the application must act right away or this becomes surprising to the user.
 
-This solution was to add a timeout on top of the requirement of it being
+The solution was to add a timeout on top of the requirement of it being
 called on the same task that resolves the `getDisplayMedia()` promise.
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -1917,7 +1917,7 @@ as this prevents most forms of code composition.
 If possible, set a short timeout interval. Failing that, viability can
 be limited to the same task, but not a single microtask.
 
-<div class="note">
+<div class="example">
 One case where this came up was the [captureController.setFocusBehavior()](https://w3c.github.io/mediacapture-screen-share/#dom-capturecontroller-setfocusbehavior)
 method, where timing concerns were instead solved by adding a timeout on top of
 the requirement of it being called on the same task that resolves the
--- a/index.bs
+++ b/index.bs
@@ -1877,6 +1877,37 @@ If the bytes in the buffer have a natural intepretation
 as one of the other TypedArray types, provide that instead.
 For example, if the bytes represent Float32 values, use a {{Float32Array}}.
 
+<h3 id="allow-microtask-switches">A Promise represents a value, not a callback</h3>
+
+Avoid requiring that something needs to happen synchronously within
+a promise resolution or rejection callback.
+
+Authors take for granted that an `async` function can be wrapped
+in another that appends synchronous code at the end e.g. to inspect state
+using try/catch. But the `await` queues a microtask:
+<div class="example">
+```js
+async function fooWrapper() {
+  const foo = await platform.foo();
+  console.log("success");
+  return foo;
+}
+
+(async () => {
+  const foo = await fooWrapper();
+  foo.bar(); // on the same task, but not the same microtask that logged "success"
+})();
+```
+</div>
+
+To not interfere with this pattern, the lowest recommended restriction is "same task".
+
+<div class="note">
+One case where this came up was the [captureController.setFocusBehavior()](https://w3c.github.io/mediacapture-screen-share/#dom-capturecontroller-setfocusbehavior)
+method, where timing concerns were solved by adding an timeout on top of
+the same task requirement.
+</div>
+
 <h2 id="event-design">Event Design</h2>
 
 <h3 id="one-time-events">Use promises for one time events</h3>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1877,6 +1877,37 @@ If the bytes in the buffer have a natural intepretation
...
+}
+
+(async () => {
+  const foo = await fooWrapper();
+  foo.bar(); // on the same task, but not the same microtask that logged "success"
+})();
+```
+</div>
+
+To not interfere with this pattern, the lowest recommended restriction is "same task".
```
alvestrand, line 1903:
```suggestion
To allow this pattern, the specification for bar() can say that it must be called in the same task as the one where the promise is returned, but cannot say that it must be called in the same microtask.
```

Trying to make text that I will understand the next time I read it.

index.bs
```diff
@@ -1877,6 +1877,37 @@ If the bytes in the buffer have a natural intepretation
...
+
+<div class="note">
+One case where this came up was the [captureController.setFocusBehavior()](https://w3c.github.io/mediacapture-screen-share/#dom-capturecontroller-setfocusbehavior)
+method, where timing concerns were solved by adding an timeout on top of
+the same task requirement.
```
alvestrand, line 1908: I don't see a same task requirement in https://w3c.github.io/mediacapture-screen-share/#dom-capturecontroller-setfocusbehavior - please clarify.
setFocusBehavior() as currently specified doesn't return a promise.

jan-ivar: setFocusBehavior() throws outside of the getDisplayMedia success task. See [wpt](https://searchfox.org/mozilla-central/rev/1d1ca77dd7ca10953c6f5f1b77b98e4eb528274e/testing/web-platform/tests/screen-capture/getdisplaymedia-capture-controller.https.window.js#150-151,153,156,158-160) and [fyi](https://wpt.fyi/results/screen-capture/getdisplaymedia-capture-controller.https.window.html?label=experimental&label=master&aligned).

The spec could use an example.

alvestrand: Aha. In order to find that, one has to find the link from "finalize focus decision algorithm" to "getDisplayMedia".
Agreed that the spec could be easier to comprehend if an example was added - or just text that states the purpose of the mechanism.

So setFocusBehavior() is an example of a function that must be called in the same task as getDisplayMedia's resolution, but the description never uses the words "same task".

jan-ivar: setFocusBehavior() queues tasks to implement this (details fixed in https://github.com/w3c/mediacapture-screen-share/issues/303).

Other specs [use](https://w3c.github.io/mediacapture-extensions/#dfn-expose-video-frame-counters-steps) _"Let task be a the current [task](https://html.spec.whatwg.org/multipage/webappapis.html#concept-task). If [[[LastTask]]](https://w3c.github.io/mediacapture-extensions/#dfn-lasttask-0) is equal to task..."_. They accomplish similar things.

index.bs
jan-ivar: I've incorporated this in a new version. PTAL

index.bs
```diff
@@ -1877,6 +1877,40 @@ If the bytes in the buffer have a natural intepretation
...
 For example, if the bytes represent Float32 values, use a {{Float32Array}}.
 
+<h3 id="allow-microtask-switches">A Promise represents a value, not a callback</h3>
+
+Avoid requiring that something needs to happen synchronously within
+a promise resolution or rejection callback.
+
+Authors tend to expect that any `async` function can be wrapped
+in another that appends synchronous code at the end e.g. to inspect state
+using try/catch. But the `await` queues a microtask:
```
martinthomson, line 1887:
```suggestion
in another that appends synchronous code at the end.
This might be so that a return value can be examined, such as using try/catch.
Using `await` or similar wrappers queues a microtask:
```

index.bs
```diff
@@ -1877,6 +1877,40 @@ If the bytes in the buffer have a natural intepretation
...
+(async () => {
+  const foo = await fooWrapper();
+  foo.bar(); // on the same task, but not the same microtask that logged "success"
+})();
+```
+</div>
+
+To support this pattern, a specification that wishes to restrict `bar()` to
+only succeed if called immediately following a promise resolving can
+minimally limit it to the same task, but not the same microtask.
```
martinthomson, line 1905: Ideally, we should discourage any such limitations first.  Then we talk about why a platform API might have cause to add a restriction like this.

I would say instead something like:

> In general, the return value from a function should be usable at any time.
>
> There could be cases where APIs might need to restrict the scope over which a value can be used.
> In the above example, perhaps the `platform.foo()` function establishes state that has real-time dependencies
> such that calling `bar()` is not viable at any future time.
>
> Never limit the validity of returned values to a microtask, as this prevents most forms of code composition.
> If possible, set a short timeout interval.
> Failing that, returned values can be limited to the same task, but not a single microtask.

jan-ivar: Totally agree. Just a bit concerned about the phrase "returned value", given the subtleties involved (technically the promise is the returned value).

Maybe s/returned value/resolved value/ ?

jan-ivar: Or maybe "settled result" since this is less about the value and more about the timing of the callback. E.g. some API devising some synchronous contract on rejection wouldn't be OK either. I'll wordsmith.

martinthomson: Oh, I really meant "returned value" to be expansive: the promise and the value it resolves to OR the value it might produce on rejection.  But your suggestion works.

martinthomson (approved):
This LGTM.  Time to ask the rest of the TAG to review.

jan-ivar:
Thanks @martinthomson!

I appear to be unable to add reviewers in this repo, so hopefully there's a process for that?

index.bs
```diff
@@ -1877,6 +1877,53 @@ If the bytes in the buffer have a natural intepretation
 as one of the other TypedArray types, provide that instead.
 For example, if the bytes represent Float32 values, use a {{Float32Array}}.
 
+<h3 id="allow-microtask-switches">A Promise represents a value, not a callback</h3>
```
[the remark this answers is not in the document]
LeaVerou: Agreed with Jeffrey.

index.bs
```diff
@@ -1877,6 +1877,53 @@ If the bytes in the buffer have a natural intepretation
...
+establishes state that has real-time dependencies such that calling
+`bar()` is not viable at any future time.
+
+Never limit the viability of acting on a settled result to a microtask,
+as this prevents most forms of code composition.
+
+If possible, set a short timeout interval. Failing that, viability can
+be limited to the same task, but not a single microtask.
+
+<div class="note">
```
LeaVerou, line 1920:
```suggestion
<div class="example">
```

index.bs
```diff
@@ -1877,6 +1877,53 @@ If the bytes in the buffer have a natural intepretation
...
+One case where this came up was the [captureController.setFocusBehavior()](https://w3c.github.io/mediacapture-screen-share/#dom-capturecontroller-setfocusbehavior)
+method, where timing concerns were instead solved by adding a timeout on top of
+the requirement of it being called on the same task that resolves the
+`getDisplayMedia()` promise.
```
LeaVerou, line 1924: This is not providing enough information for the reader to understand how the principle was violated.

LeaVerou (changes requested):
Overall this seems a bit esoteric, and I agree with @jyasskin that the title as stated is not always true. I think the examples need some work to establish what should or shouldn't be done, otherwise this principle only makes sense to those who already understand it.

index.bs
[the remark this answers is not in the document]
jan-ivar: Makes sense. 

> Maybe a title like "Promise resolution should not expire", although that could probably be improved.

Resolution does often expire though. The principle here is to avoid synchronous expiry as a design (a pattern with some callbacks or even event listeners).

How about "A Promise represents completion or a value, not a callback" ?

jan-ivar:
Feedback incorporated. Thanks!

index.bs
```diff
@@ -1877,6 +1877,59 @@ If the bytes in the buffer have a natural intepretation
...
+as this prevents most forms of code composition.
+
+If possible, set a short timeout interval. Failing that, viability can
+be limited to the same task, but not a single microtask.
+
+<div class="example">
+One case where this came up was the [captureController.setFocusBehavior()](https://w3c.github.io/mediacapture-screen-share/#dom-capturecontroller-setfocusbehavior)
+method which controls whether a window the user selects to screen-capture
+should immediately be pushed to the front or not. It can be called as late
+as right <em>after</em> the `getDisplayMedia()` promise. This allows
```
alvestrand, line 1925:
```suggestion
as right <em>after</em> the resolution of the `getDisplayMedia()` promise. This allows
```

martinthomson (commented):
I like this.  I think that the one thing that might be added

> If an API depends on setting up temporary conditions then invoking the caller, 
> that is a good reason to use a callback rather than a promise.

jan-ivar:
@LeaVerou does this look better?

index.bs
```diff
@@ -1877,6 +1877,62 @@ If the bytes in the buffer have a natural intepretation
...
+One case where this came up was the [captureController.setFocusBehavior()](https://w3c.github.io/mediacapture-screen-share/#dom-capturecontroller-setfocusbehavior)
+method which controls whether a window the user selects to screen-capture
+should immediately be pushed to the front or not. It can be called as late
+as right <em>after</em> the resolution of the `getDisplayMedia()` promise. This allows
+applications to make a different decision based on the window the user chose.
+But the application must act right away or this becomes surprising to the user.
+
+The solution was to add a timeout on top of the requirement of it being
+called on the same task that resolves the `getDisplayMedia()` promise.
+</div>
```
jyasskin, line 1931: Given my suggestion above, some of this text becomes redundant:

```suggestion
Occasionally, users need the state represented by a Promise to automatically move on.

<div class="example">
Developers need to be able to call
<code>captureController.{{CaptureController/setFocusBehavior()}}</code>
after the Promise returned by {{MediaDevices/getDisplayMedia()}}
so that they can use attributes of the window the user chose
when deciding how focus should change.

However, users will be surprised if
the focus change doesn't feel like an immediate effect from selecting a window,
so {{CaptureController/setFocusBehavior()}} needs to automatically stop working
shortly after that Promise fulfills.
</div>

A timeout can be a good trigger for these automatic state changes,
especially when they're happening to aid user perception.

If you need a faster automatic state change,
you can [=queue a task=] to make the change.

As mentioned above,
never change the state established by a Promise between microtasks.
```

index.bs
```diff
@@ -1877,6 +1877,62 @@ If the bytes in the buffer have a natural intepretation
...
+
+If an API depends on setting up temporary conditions then invoking the caller,
+that is a good reason to use a callback rather than a promise.
```
jyasskin, line 1934: Maybe:

```suggestion
If a feature requires the developer to only rely on some state between precise points within a task,
that can be a good reason to use a callback rather than a promise.
```

jyasskin (commented):
Sorry for taking a long time to review this. I ... rewrote a big chunk of it .... What do you think?

index.bs
```diff
@@ -1877,6 +1877,62 @@ If the bytes in the buffer have a natural intepretation
 as one of the other TypedArray types, provide that instead.
 For example, if the bytes represent Float32 values, use a {{Float32Array}}.
 
+<h3 id="allow-microtask-switches">A Promise represents completion or a value, not a callback</h3>
+
+Avoid requiring that something needs to happen synchronously within
+a promise resolution or rejection callback.
```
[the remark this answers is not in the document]
martinthomson: You seem to have hit on something useful that is at the core of this:

* The state of the execution environment, as presented to the developer, cannot change except when executing a new task.

It's OK for the information presented to be a view of the state of things that might subsequently (or even immediately) be invalidated.  If the API relates to something that is acting outside of the sandbox, like a camera or another application, then there is no way to be certain that information presented about that thing is correct.  The best you can do is send it messages and see what happens.  But things that comprise the environment are there to serve the developer's needs.

The whole Promise resolution thing is secondary to all of that.  Maybe there is a different framing we could use that highlights that.

index.bs
jan-ivar: > If you need a faster automatic state change, you can [=queue a task=] to make the change.

Faster in the non-degenerate case. To clarify the example, setFocusBehavior() does _both_: It queues _"a task to run the [finalize focus decision algorithm](https://w3c.github.io/mediacapture-screen-share/#dfn-finalize-focus-decision-algorithm)"_ AND applies a deadline for that task.

IOW, the timer is only there to catch busy-waiting exploits in the callback.

index.bs
```diff
@@ -1877,6 +1877,62 @@ If the bytes in the buffer have a natural intepretation
 as one of the other TypedArray types, provide that instead.
 For example, if the bytes represent Float32 values, use a {{Float32Array}}.
 
+<h3 id="allow-microtask-switches">A Promise represents completion or a value, not a callback</h3>
```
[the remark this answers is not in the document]
jan-ivar: Won't that renumber § 6.11 through § 6.14?

index.bs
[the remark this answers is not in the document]
jan-ivar: Thanks for the review and text!

> The state of the execution environment, as presented to the developer, cannot change except when executing a new task.

I'm not sure we can say the environment is immutable. E.g. JS can call track.[stop()](https://w3c.github.io/mediacapture-main/getusermedia.html#dom-mediastreamtrack-stop) and synchronously observe track.readyState changed from `"live"` to `"ended"`. Only in-parallel changes seem problematic. 

Also, apart from the busy-wait timer (which is more like ~1 second and a mitigation really), I don't think it's accurate to describe an overly strict and synchronous API contract as a in-parallel change to the environment.

The platform already seems to contain examples of synchronous API contracts, like [preventDefault()](https://dom.spec.whatwg.org/#ref-for-dom-event-preventdefault%E2%91%A2) that don't seem to violate [§ 5.2. Preserve run-to-completion semantics](https://pr-preview.s3.amazonaws.com/jan-ivar/design-principles/pull/496.html#js-rtc).

I think the anti-pattern is applying such synchronous API contracts to a promise callback.

jyasskin (changes requested):

index.bs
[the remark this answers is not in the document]
jan-ivar: A singular underlying principle is encouraging, suggesting we're on the right track, or at least being consistent.

But practically speaking, might it still deserve two guidelines if the patterns to avoid are distinct enough?

Anti-pattern 1:
- rationale: app must do something right away or not at all; a microtask window seems shorter than a task 
- symptom: breaks the async-functions-are-wrappable invariant

Anti-pattern 2:
- rationale: mitigate busy-waiting exploits; a 1 second timer deadline won't bother well-behaved apps
- symptom: breaks malicious or already broken apps as a secondary action at a distance symptom

Reasons to treat them separately:
- the first one seems more serious than the second
- the second one seems in the long-running scripts mitigation category; exempt from guidelines?
- `setFocusBehavior()` went from violating only the first to violating only the second; an improvement?

I'm concerned these distinctions might be lost if we merge/generalize the language too much.

jan-ivar:
Note I replied a few days ago in two threads above, which github isn't that great at highlighting. I'm a bit worried going too general here risks losing the advice.

jyasskin:
Ok, #536, https://w3ctag.github.io/design-principles/#js-rtc, is merged and covers most of the content we wanted to include in this principle. @jan-ivar, it'll probably be least confusing if you open a new PR or issue for whatever you think is left, so the discussion doesn't get mixed up. I'll leave this open as a reminder to check on that.

index.bs
martinthomson: I don't understand this statement:

> Occasionally, users need the state represented by a Promise to automatically move on.

jyasskin: I think this PR should probably be entirely rewritten or abandoned, so we don't need to wordsmith all the pieces of it, but the sense there was that Promise resolution implies some state of the world, like "the user wants to present X window". Normally, that state should be stable until the user or program explicitly changes it. But in the case of `getDisplayMedia()` the state also includes "and the user might want to automatically focus the window", which the user wants to stop being true after they've stopped feeling like they just picked it. So it needs to "automatically move on" after some small amount of time.

jyasskin:
@jan-ivar, in https://github.com/w3ctag/design-principles/pull/536#pullrequestreview-2502562802 you asked that we keep this open for a little bit. Is there anything more to do with this, or shall we close it now?

jan-ivar:
It's taken me this long to put my finger on where I think [#js-rtc](https://w3ctag.github.io/design-principles/#js-rtc) fails to fully subsume [#allow-microtask-switches](https://pr-preview.s3.amazonaws.com/jan-ivar/design-principles/pull/496.html#allow-microtask-switches) (this PR) as guidance.

The former preserves single-thread semantics by queueing in-parallel (surprise) state changes to avoid them mid-task, while the latter explicitly warns against requiring _synchronous usage_ of a Promise result. 

Examples of _synchronous contracts_ exist in callbacks tday:
```js
window.onbeforeunload = event => event.preventDefault(); // prompts the user "Leave site?" 
```
I.e. I believe this isn't supposed to work _(ignore that it seems racy in Chrome and Firefox, at least from web console!)_:
```js
window.onbeforeunload = async event => { await undefined; event.preventDefault(); } // ignore!?
```
It might be useful to pause here and rationalize why the above doesn't violate #js-rtc as now written.

Now, hypothetically, say someone designed this today, and followed [#one-time-events](https://w3ctag.github.io/design-principles/#one-time-events) (I know it can fire again... it's just an example):
```js
window.beforeUnload.then(result => result.preventDefault());
```
Now should the following work?
```js
window.beforeUnload.then().then(result => result.preventDefault());
```

In this hypothetical, losing preventDefault() after a microtask break is really a design issue: the API gave you a Promise that must be used synchronously—exactly what #allow-microtask-switches says “don’t do.”

- #allow-microtask-switches is violated because it breaks basic assumptions around asynchronous function composition.
- #js-rtc primarily forbids the browser from injecting mid-task changes outside the developer’s control. But here, the dev is faced with a stated API contract, so it’s not an invisible or external mid-task update—it’s an API design that’s poorly suited specifically for Promise callbacks (but not regular callbacks).

jan-ivar:
@jyasskin ^

jyasskin:
Sorry @jan-ivar for taking a long time to get back to this. I agree that we should say something about why an API with a synchronous or 1-microtask `.preventDefault()` doesn't make sense as a `Promise` API. Might this fit inside https://w3ctag.github.io/design-principles/#promises, by adding a definition of asynchronous APIs and saying that Promises should _not_ be used for non-asynchronous APIs?

I suspect it'll be more productive to draft a new PR in that direction, rather than continuing this one, but perhaps we should discuss more in #342 before starting that PR? But if you or someone else wants to start that PR now, I'm happy to discuss in the PR instead.

Issue #289, opened 2021-02-24
Guidance about which state properties should be reflected as HTML attributes


LeaVerou:
This came up while reviewing `<popup>` in https://github.com/w3ctag/design-reviews/issues/599#issuecomment-781386845, and more specifically about whether it should have an `open` attribute like `<dialog>` and `<details>`.

There is [another issue about keeping IDL attributes and HTML attributes in sync](https://github.com/w3ctag/design-principles/issues/279) when both exist, but we don't have anything about *which* IDL attributes should also exist as attributes. 

My personal opinion is to err on the side of including attributes for main properties. Providing declarative HTML-based ways to interact with UI controls has several benefits. Off the top of my head:

- It facilitates experimentation and iteration (e.g. much quicker to type `open` when iterating than to write script)
- It makes it more learnable: there is a baseline uniform API (DOM) that novices can use to interact with the element via script, even before learning the more specialized API that might come with each element
- It means behaviors can be expressed reactively in libraries supporting such syntax (e.g. `open={{expression}}`) and be handled by any library handling HTML generically.
- It means UI states can be addressed via established and widely understood attribute selectors, instead of new pseudo-classes
- It makes state inspectable via dev tools which facilitates debugging

On the other hand, there have been some arguments against this, I believe primarily centered around performance. In the case of `<dialog open>`, things got messy I believe due to some inconsistency with the `open` and `close` events, I'm not sure if there's anything more general to learn from that.

Anyhow, I think this is something we should discuss, resolve on so that we're all on the same page, and provide guidance about in the form of a design principle.

annevk:
I think there is an unresolved issue around `dialog` elements when the events fire. I think if they fire using custom-element-reaction time™ and only in documents with a browsing context it ought to be fine, but that's not the case right now.

The `dialog` element is also somewhat precedent breaking in that for other form controls, user input never modified the tree. And it's also not clear how to do certain features with that model. E.g., for `<textarea>` its contents represent the default value. How would you serialize user input?

LeaVerou:
> The `dialog` element is also somewhat precedent breaking in that for other form controls, user input never modified the tree. And it's also not clear how to do certain features with that model. E.g., for `<textarea>` its contents represent the default value. How would you serialize user input?

[...]

PR #501, opened 2024-07-15
Guidance about reflecting state in HTML attributes

LeaVerou:
***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/501.html" title="Last updated on Dec 3, 2024, 4:27 PM UTC (039c7ef)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/501/55a06e0...039c7ef.html" title="Last updated on Dec 3, 2024, 4:27 PM UTC (039c7ef)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1035,7 +1035,7 @@ See also [[#worker-only]].
 <h3 id="html-attribute-property-reflection">Expose content attributes for writable IDL attributes</h3>
 
 When designing a new *writable* IDL attribute to specify element state,
-err on the side of exposing a corresponding content attribute.
+expose a corresponding content attribute unless you have a compelling reason not to.
 
 Providing declarative ways to manipulate state has several benefits for authors:
 - It takes advantage of HTML reactivity, and alleviates the need to apply changes at a specific moment in time.
--- a/index.bs
+++ b/index.bs
@@ -1032,6 +1032,25 @@ The alternative user experience is a flash of unstyled content, which is undesir
 
 See also [[#worker-only]].
 
+<h3 id="html-attribute-property-reflection">Expose content attributes for writable IDL attributes</h3>
+
+When designing a new *writable* IDL attribute to specify element state,
+err on the side of exposing a corresponding content attribute.
+
+Providing declarative ways to manipulate state has several benefits for authors:
+- It takes advantage of HTML reactivity, and alleviates the need to apply changes at a specific moment in time.
+- It integrates with existing DOM methods, improving learnability, whereas an API specific to the feature needs to be learned separately.
+- Behaviors can be expressed reactively in libraries supporting HTML-based syntax and be handled by any library handling HTML generically.
+- UI states can be addressed in CSS via simple attribute selectors rather than new pseudo-classes.
+- State can be inspected and manipulated via the browser's developer tools with no need for implementing a new UI.
+
+<div class="example">
+A counterpattern to this guidance can be found in
+<{input}>'s  {{HTMLInputElement/indeterminate}} IDL property
+which was never exposed as a content attribute making it impossible to fully update checkbox state
+by modifying HTML.
+</div>
+
 <h3 id="html-idl-must-by-synced">Keep attributes in sync</h3>
 
 New content attributes
--- a/index.bs
+++ b/index.bs
@@ -1038,7 +1038,7 @@ When designing a new *writable* IDL attribute to specify element state,
 expose a corresponding content attribute unless you have a compelling reason not to.
 
 Providing declarative ways to manipulate state has several benefits for authors:
-- It takes advantage of HTML reactivity, and alleviates the need to apply changes at a specific moment in time.
+- It takes advantage of HTML’s [reactivity](https://en.wikipedia.org/wiki/Reactive_programming), and alleviates the need to apply changes at a specific moment in time.
 - It integrates with existing DOM methods, improving learnability, whereas an API specific to the feature needs to be learned separately.
 - Behaviors can be expressed reactively in libraries supporting HTML-based syntax and be handled by any library handling HTML generically.
 - UI states can be addressed in CSS via simple attribute selectors rather than new pseudo-classes.


index.bs
```diff
@@ -1032,6 +1032,25 @@ The alternative user experience is a flash of unstyled content, which is undesir
...
+<h3 id="html-attribute-property-reflection">Expose content attributes for writable IDL attributes</h3>
+
+When designing a new *writable* IDL attribute to specify element state,
+err on the side of exposing a corresponding content attribute.
+
+Providing declarative ways to manipulate state has several benefits for authors:
+- It takes advantage of HTML reactivity, and alleviates the need to apply changes at a specific moment in time.
+- It integrates with existing DOM methods, improving learnability, whereas an API specific to the feature needs to be learned separately.
+- Behaviors can be expressed reactively in libraries supporting HTML-based syntax and be handled by any library handling HTML generically.
+- UI states can be addressed in CSS via simple attribute selectors rather than new pseudo-classes.
```
hober, line 1044: I wonder if this is a good thing in all cases. For instance, `[lang=foo]` v. `[lang|=foo]` v. `:lang(foo)`.

index.bs
```diff
@@ -1032,6 +1032,25 @@ The alternative user experience is a flash of unstyled content, which is undesir
...
+- State can be inspected and manipulated via the browser's developer tools with no need for implementing a new UI.
+
+<div class="example">
+A counterpattern to this guidance can be found in
+<{input}>'s  {{HTMLInputElement/indeterminate}} IDL property
+which was never exposed as a content attribute making it impossible to fully update checkbox state
+by modifying HTML.
+</div>
```
hober, line 1052: On the other hand, `indeterminate` is a terrible feature that should never have been added to the platform in the first place. Maybe we should try to find a more positive example?

hober (approved):
Looks good to me.

index.bs
```diff
@@ -1032,6 +1032,25 @@ The alternative user experience is a flash of unstyled content, which is undesir
 
 See also [[#worker-only]].
 
+<h3 id="html-attribute-property-reflection">Expose content attributes for writable IDL attributes</h3>
+
+When designing a new *writable* IDL attribute to specify element state,
+err on the side of exposing a corresponding content attribute.
+
+Providing declarative ways to manipulate state has several benefits for authors:
+- It takes advantage of HTML reactivity, and alleviates the need to apply changes at a specific moment in time.
```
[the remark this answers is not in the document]
LeaVerou: [Reactive Programming](https://en.wikipedia.org/wiki/Reactive_programming) certainly is and it falls out of the definition that HTML is reactive, but I’m not aware of a source establishing that HTML is reactive.

index.bs
LeaVerou: This seems orthogonal. In that case the pseudo-class serves a distinct purpose by targeting the actual computed language (which may be specified on an ancestor). I’d argue there may even be value in a generic attribute selector for these inherited attributes.

index.bs
LeaVerou: Could you elaborate? Is it the UI of an indeterminate checkbox that you think is a mistake or the way it was added to HTML?

index.bs
```diff
@@ -1032,6 +1032,25 @@ The alternative user experience is a flash of unstyled content, which is undesir
 
 See also [[#worker-only]].
 
+<h3 id="html-attribute-property-reflection">Expose content attributes for writable IDL attributes</h3>
+
+When designing a new *writable* IDL attribute to specify element state,
```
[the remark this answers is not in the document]
LeaVerou: I disagree, I think "writable" is essential to the principle and should be emphasized.

index.bs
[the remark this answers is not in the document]
LeaVerou:
```suggestion
- It takes advantage of HTML’s [reactivity](https://en.wikipedia.org/wiki/Reactive_programming), and alleviates the need to apply changes at a specific moment in time.
```

index.bs
LeaVerou: There’s a separate, existing, principle about this: https://www.w3.org/TR/design-principles/#html-idl-must-by-synced

`indeterminate` *in addition* to that, also has no content attribute counterpart so there is no HTML way to specify it at all (current OR initial status).

LeaVerou: @hober Interesting. Could you elaborate on the UI part?

annevk: That principle is about content attributes needing to be reflected right? But `indeterminate` is not a content attribute.

index.bs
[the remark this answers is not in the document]
martinthomson: I would suggest a rephrase then.

> When designing a new IDL attribute that reflects the state of an element,
> expose a corresponding content attribute,
> unless the attribute is immutable or
> there are other strong reasons not to.

martinthomson (commented):
I made this comment a VERY long time ago.  I have not re-examined it for whether I still believe it.

index.bs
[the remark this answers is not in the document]
xiaochengh: I'm also confused why "writable" matters here. Is there a reason immutable IDL attributes are less suitable to be reflected?

xiaochengh:
Just a note that the original example `<popup>` in #289 is no longer there, and got replaced by the `popover` attribute. And  popover visibility state is not exposed as content or IDL attributes, but only via CSS `:popover-open` selector.

annevk:
To be clear, https://github.com/w3ctag/design-principles/pull/501/files#r1680595686 is still applicable. This should not be merged with the `indeterminate` example.

martinthomson:
I'm not sure that I understand the concern about `indeterminate`.  It seems like this could be made a content attribute (though what `indeterminate` + `checked` means is interesting to contemplate).  The `popover` example might be similarly enlightening.

Basically, I think that our examples need to show where values are reflected as attributes in a good way (maybe; that's most of the platform, so it doesn't make sense to cherry pick one example) and where values were not reflected as attributes, to make the reasons for why that might be a good choice clear.

annevk:
The concern is that the `indeterminate` IDL attribute is akin to the `value` IDL attribute, which doesn't have a content attribute counterpart. There's a `value` content attribute, but its IDL counterpart is `defaultValue`. So this suggestion makes it seem as if the TAG doesn't understand how HTML forms work. Which is reasonable as it's rather involved and essentially everybody gets confused about this, but adding to this confusion seems ill-advised.

martinthomson:
To be clear, \*I\* don't understand how forms work; the use of pronoun should have made that clear.  I'm just trying to shepherd this to a constructive conclusion.

@annevk, perhaps I can request that you offer an example that would help illustrate this principle.  Or, maybe you would prefer that we not pursue this further.  Explaining why would be helpful.

annevk:
Can you clarify what part about my `value` example is unclear?

martinthomson:
Ah, I hadn't interpreted that as a suggestion for a better example.  Indeed, I would be OK switching to `value` and `defaultValue` in the example.

annevk:
Oh my bad, I had in fact misunderstood your question.

I'm actually not sure the HTML community is fully aligned on this principle. While some state certainly deserves to be reflected through a content attribute, that does not go for state that is the result of end user input.

The `open` content attribute of the `dialog` and `details` elements is designed to always reflect the current state of the control, but this has resulted in a number of subtle issues and also seems to go against expectations. Web developers don't expect node tree mutations as the result of end users manipulating a control. We also have to be very careful to not re-introduce mutation events if we want to make these state changes observable to web developers.

cc @smaug---- @lukewarlow

lukewarlow:
I do agree that IDL attributes should generally reflect to HTML content attributes. And that the names should match up (value attribute should be defaultvalue imo).

Being able to represent initial state via HTML is good. Indeterminate is an example where we can't currently do that.

The specific point about targeting attributes rather than pseudo-classes seems wrong though. The open attribute is a design mistake and even that we now have the `:open` pseudo class to target instead of using `[open]`. Custom elements can similarly follow this pattern using CSS custom states.

Also suggesting that reflecting IDL to HTML content attributes allows for this isn't quite right either, because internal element state that's changed via interaction (other than for open) doesn't reflect to the attribute and so you'd still need a CSS way to access that state.

This note seems to be over reaching slightly.

The note about reactive html also seems potentially wrong depending on if I'm reading it right. The open attribute is reactive and even that doesn't work if your library is setup to act on the attribute (or property, tbf so perhaps not a great example?)

martinthomson:
Discussed today in a breakout.  Given the lack of input here, we're going to close this specific PR.  Good discussion here though, which I've referenced in the issue (#289).

PR #502, opened 2024-07-15
Establish a rule about element overloading

### LINKED ISSUE #370: New principle: Discourage overloading (in HTML)
This was brought up by @hober in a breakout today.

We should have a principle about avoiding things like `<input type>` in the future.

martinthomson:
I'm sure that we could say more, but I think that this is enough.

Closes #370.

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/502.html" title="Last updated on Dec 4, 2024, 4:10 PM UTC (c0567ab)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/502/efad741...c0567ab.html" title="Last updated on Dec 4, 2024, 4:10 PM UTC (c0567ab)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1174,6 +1174,43 @@ If the attribute identifies a URL that is auxilliary to the element's purpose, l
 
 Note: remember that attributes containing URLs should be represented in IDL as {{USVString}}; see [[#idl-string-types]].
 
+<h3 id="html-overloading">Give each HTML element a single purpose</h3>
+
+Each HTML element should have a clear purpose.
+Using HTML attributes can modify the semantics of an element,
+but attributes should not fundamentally alter that purpose.
+
+Rather than defining elements that can operate in different modes,
+having a separate element for each mode is preferable.
+This simplifies usage for authors.
+
+<div class=example>
+The <{input}> element uses a <{input/type}> attribute to switch between
+very different modes.
+Though each of these modes share a common goal of taking user input,
+that is too broad a purpose.
+The semantics of these modes could have been more clearly expressed
+through the use of distinct element names.
+</div>
+
+Overloading of semantics on the same element
+through the use of attributes could be used to select
+between different behavior.
+This only makes sense where the purpose remains the same
+for any value of the attributes.
+
+<div class=example>
+The <{textarea}> element and an <{input}> element with a <{input/type}> of `"text"`
+could have used the same element as the differences between these is not material.
+</div>
+
+Potential counterexamples include overspecialization and elements that rely on context to establish their semantics.
+The now-deprecated <{acronym}> element is duplicative of the <{abbr}> element
+and therefore overspecialization.
+The <{source}> element can have distinct purposes depending on context.
+A <{source}> element is not standalone; it's semantics are determined by its parent.
+
+
 <h2 id="css">Cascading Style Sheets (CSS)</h2>
 
 This section details design principles for features which are exposed via CSS.
--- a/index.bs
+++ b/index.bs
@@ -1204,11 +1204,18 @@ The <{textarea}> element and an <{input}> element with a <{input/type}> of `"tex
 could have used the same element as the differences between these is not material.
 </div>
 
-Potential counterexamples include overspecialization and elements that rely on context to establish their semantics.
+Potential counterexamples include overspecialization
+and elements that rely on context to establish their semantics.
+
+<div class=example>
 The now-deprecated <{acronym}> element is duplicative of the <{abbr}> element
 and therefore overspecialization.
+</div>
+
+<div class=example>
 The <{source}> element can have distinct purposes depending on context.
 A <{source}> element is not standalone; it's semantics are determined by its parent.
+</div>
 
 
 <h2 id="css">Cascading Style Sheets (CSS)</h2>


index.bs
```diff
@@ -1066,6 +1066,43 @@ If the attribute identifies a URL that is auxilliary to the element's purpose, l
...
 Note: remember that attributes containing URLs should be represented in IDL as {{USVString}}; see [[#idl-string-types]].
 
+<h3 id="html-overloading">Give each HTML element a single purpose</h3>
+
+Each HTML element should have a clear purpose.
+Using HTML attributes can modify the semantics of an element,
+but attributes should not fundamentally alter that purpose.
+
+Rather than defining elements that can operate in different modes,
+having a separate element for each mode is preferable.
```
hober, line 1076: Counterexample: `<legend>`, `<caption>`, and `<figcaption>` probably should have all been the same element (but can't be for compat reasons).

index.bs
```diff
@@ -1066,6 +1066,43 @@ If the attribute identifies a URL that is auxilliary to the element's purpose, l
...
+<div class=example>
+The <{textarea}> element and an <{input}> element with a <{input/type}> of `"text"`
+could have used the same element as the differences between these is not material.
+</div>
+
+Potential counterexamples include overspecialization and elements that rely on context to establish their semantics.
+The now-deprecated <{acronym}> element is duplicative of the <{abbr}> element
+and therefore overspecialization.
+The <{source}> element can have distinct purposes depending on context.
+A <{source}> element is not standalone; it's semantics are determined by its parent.
```
hober, line 1103:
```suggestion
A <{source}> element is not standalone; its semantics are determined by its parent.
```

hober (approved):
Looks good to me.

hober:
It would be nice for a non-native English speaker to take an editorial pass on this.

index.bs
```diff
@@ -1066,6 +1066,43 @@ If the attribute identifies a URL that is auxilliary to the element's purpose, l
...
+Each HTML element should have a clear purpose.
+Using HTML attributes can modify the semantics of an element,
+but attributes should not fundamentally alter that purpose.
+
+Rather than defining elements that can operate in different modes,
+having a separate element for each mode is preferable.
+This simplifies usage for authors.
+
+<div class=example>
+The <{input}> element uses a <{input/type}> attribute to switch between
```
LeaVerou, line 1080: `<link>` is another good example to include here.

index.bs
```diff
@@ -1066,6 +1066,43 @@ If the attribute identifies a URL that is auxilliary to the element's purpose, l
...
+
+Overloading of semantics on the same element
+through the use of attributes could be used to select
+between different behavior.
+This only makes sense where the purpose remains the same
+for any value of the attributes.
+
+<div class=example>
+The <{textarea}> element and an <{input}> element with a <{input/type}> of `"text"`
+could have used the same element as the differences between these is not material.
```
LeaVerou, line 1096:
```suggestion
could have used the same element as the differences between these are not material.
```

index.bs
```diff
@@ -1066,6 +1066,43 @@ If the attribute identifies a URL that is auxilliary to the element's purpose, l
...
+</div>
+
+Potential counterexamples include overspecialization and elements that rely on context to establish their semantics.
+The now-deprecated <{acronym}> element is duplicative of the <{abbr}> element
```
LeaVerou, line 1100: This should be in a `<div class=example>`

index.bs
```diff
@@ -1066,6 +1066,43 @@ If the attribute identifies a URL that is auxilliary to the element's purpose, l
...
+and therefore overspecialization.
+The <{source}> element can have distinct purposes depending on context.
```
LeaVerou, line 1102: This should also be in a (different) `<div class=example>`.

I disagree that this is an antipattern. The purpose of `<source>` is clear, the details change based on its context which is fine — having distinct elements would have been an antipattern just like `<caption>`/`<figcaption>`/`<legend>` that @hober mentioned.

index.bs
```diff
@@ -1066,6 +1066,43 @@ If the attribute identifies a URL that is auxilliary to the element's purpose, l
...
+that is too broad a purpose.
+The semantics of these modes could have been more clearly expressed
+through the use of distinct element names.
+</div>
+
+Overloading of semantics on the same element
+through the use of attributes could be used to select
+between different behavior.
+This only makes sense where the purpose remains the same
+for any value of the attributes.
```
LeaVerou, line 1092: This reads a little awkward. Maybe something like

```suggestion
through the use of attributes is valid when used to differentiate
behavior within the same purpose.
```

?

LeaVerou (changes requested):
See comments.

I wonder if it might make sense to mention distinct elements sharing the same interface as a halfway solution, with the example of `<video>` and `<audio>` as instances of `HTMLMediaElement`. `<h1>` to `<h6>` could also be a counterexample of insufficient overloading.

index.bs
hober: I think Martin agrees with you. This suggests this needs to be reworded so that people don't misread him.

LeaVerou:
> It would be nice for a non-native English speaker to take an editorial pass on this.

🙋🏽‍♀️

index.bs
```diff
@@ -1066,6 +1066,43 @@ If the attribute identifies a URL that is auxilliary to the element's purpose, l
...
+having a separate element for each mode is preferable.
+This simplifies usage for authors.
+
+<div class=example>
+The <{input}> element uses a <{input/type}> attribute to switch between
+very different modes.
+Though each of these modes share a common goal of taking user input,
+that is too broad a purpose.
+The semantics of these modes could have been more clearly expressed
+through the use of distinct element names.
```
annevk, line 1085: While this is true, I wouldn't want people to walk away with the idea that we should completely abandon `input`. Making minor changes to `input` seems clearly preferable to creating a whole new element.

hober: Yeah, I'll try to come up with an additional sentence to carry this nuance in a follow up.

LeaVerou (approved):

PR #504, opened 2024-07-15
Prefer composition

### LINKED ISSUE #298: Guidance for when to use inheritance vs composition
This was brought up in our discussion of https://github.com/w3ctag/design-reviews/issues/587.

In the example spec, there's a question if the color select event should inherit from a PointerEvent, or should be a new event type that contains an optional reference to the pointer event.

This raises the general question, of when should we recommend composition vs inheritance. We should do some spelunking and look for existing patterns.

martinthomson:
WIP

Closes #298.

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/504.html" title="Last updated on Aug 21, 2024, 8:01 AM UTC (9620dfc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/504/55a06e0...9620dfc.html" title="Last updated on Aug 21, 2024, 8:01 AM UTC (9620dfc)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -900,12 +900,6 @@ something that handles the supertype might not work properly.
 In the theory of type systems, this is known as the
 [Liskov Substitution Principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle).
 
-A simpler approach is often to avoid inheritance
-and reuse existing capabilities
-by composition.
-New items can define properties that hold
-any existing components that are needed.
-
 <div class=example>
 The {{Event}} type is a super-type of {{KeyboardEvent}} and {{PointerEvent}}.
 Events always have the same basic set of properties and actions that apply,
@@ -913,6 +907,12 @@ like whether the event {{Event/bubbles}}.
 The subtypes add new properties and actions,
 but instances of those subtypes still act in every way as an {{Event}}.
 </div>
+
+A simpler approach is often to avoid inheritance
+and reuse existing capabilities
+by composition.
+New items can define properties that hold
+any existing components that are needed.
 
 <div class=example>
 Form-associated custom elements allow custom elements to interact with form submission.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -783,7 +783,7 @@ These APIs should be designed to require some indication of user intention (such
 This indicates that the user is intentionally interacting with the web page in question.
 
 User activation is defined in detail
-[in the HTML standard](https://html.spec.whatwg.org/#tracking-user-activation).
+[in the HTM sLtandard](https://html.spec.whatwg.org/#tracking-user-activation).
```
hober, line 786: I doubt you meant to make this change.

hober (approved):
Looks good to me, modulo the one obvious typo.

index.bs
martinthomson:
```suggestion
[in the HTML standard](https://html.spec.whatwg.org/#tracking-user-activation).
```

index.bs
```diff
@@ -883,6 +883,48 @@ that authors can use to extend the language
...
+New items can define properties that hold
+any existing components that are needed.
+
+<div class=example>
+The {{Event}} type is a super-type of {{KeyboardEvent}} and {{PointerEvent}}.
+Events always have the same basic set of properties and actions that apply,
+like whether the event {{Event/bubbles}}.
+The subtypes add new properties and actions,
+but instances of those subtypes still act in every way as an {{Event}}.
+</div>
```
annevk, line 915: Event uses inheritance though. And for EventTarget we encourage it too although maybe there is a case where EventTarget is mixed in instead.

martinthomson: This is intended to show a good example of inheritance.  Maybe this makes sense above the "A simpler approach" paragraph, to avoid the implication that this is talking about composition.  Would that be better?

annevk: Yes!

Issue #473, opened 2024-02-10
New principle: Incremental user effort should result in incremental value (or more)


LeaVerou:
This point is alluded to in [2.2. Consider tradeoffs between high level and low level APIs](https://w3ctag.github.io/design-principles/#high-level-low-level):

> A well-layered solution should ensure continuity of the ease-of-use vs power tradeoff curve and avoid sharp cliffs where a small amount of incremental use case complexity results in a large increase of code complexity.

However I think it’s important enough to promote to a separate principle. It’s not just about making simple things easy and complex things possible, it’s also about everything in between, and I see Web platform APIs (and UIs in general) with sharp cliffs all the time I'd say it's one of the most pervasive API design problems.

martinthomson:
WebRTC has a similarly awkward complexity cliffs.

PR #505, opened 2024-07-16
Text formats are for people

### LINKED ISSUE #453: New principle: Text-based syntaxes should be designed to be usable by humans
This came up in my discussion with @eemeli about MessageFormat 2, whose syntax is designed under the assumption that it will primarily be generated by tooling.

IMO any text-based format *will* inevitably be written and/or edited by humans, and this should always take the experience of being written by humans into account in the design of the syntax. We've failed at this before (e.g. SVG path syntax), and the result has been humans editing the format and hating every minute of it. We already have formats that are machine-friendly: binary formats. The primary reason to define text-based formats is to make them editable by humans, so we may as well go all the way.

The tradeoff of course is how to balance goals like parsing efficiency and filesize efficiency, both of which are inversely correlated with human usability. What strategy would be best? Two syntaxes, one machine-friendly and one human-friendly? But then you end up with the same problem, as humans may still need to *edit* the machine-friendly format generated by a tool. Should we just recommend that cases where efficiency is important should be binary formats? That also seems suboptimal. Maybe we should have a more vague recommendation like "try to balance these concerns with human usability"? Let's discuss.

martinthomson:
Draft

Closes #453.

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/505.html" title="Last updated on Dec 4, 2024, 3:55 PM UTC (f1cfef8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/505/55a06e0...f1cfef8.html" title="Last updated on Dec 4, 2024, 3:55 PM UTC (f1cfef8)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -587,7 +587,7 @@ See also:
 * [[#secure-context]]
 * [[#consent]]
 
-<h3 id=text-formats>Design textual formats for people</h3>
+<h3 id=text-formats>Design textual formats for humans</h3>
 
 Design textual formats that can be easily produced and consumed by people.
 
--- a/index.bs
+++ b/index.bs
@@ -587,6 +587,44 @@ See also:
 * [[#secure-context]]
 * [[#consent]]
 
+<h3 id=text-formats>Make textual formats for people</h3>
+
+Design textual formats to be both produced and consumed by people.
+
+People who are presented with a textual format will
+expect to be able to use a text editor
+to produce or modify content.
+People who edit text will introduce a range of errors, but
+could struggle to identify and fix those errors.
+
+People expect some amount of flexibility in terms of how their edits are processed.
+Allowing flexibility in whitespace, quoting, delimiters, and other syntactic elements ensures
+that content is still comprehensible.
+
+<div class=example>
+The JSON format is an example of a format that appears to be flexible,
+but it lacks many of these basic usability amenities.
+In JSON there is:
+no commenting capability,
+objects and arrays cannot have trailing commas, and
+quotes are mandatory for object keys.
+This makes JSON prone to syntactic errors as a result of manual edits.
+</div>
+
+Containing the scope of a format that is affected by a syntax error
+could improve robustness and human usability.
+This requires that specifications fully define processing and
+error handling rules so that errors are handled in the same way.
+
+<div class=example>
+Typos in CSS properties only cause that property to be ignored.
+Errors in properties rarely cause an entire rule to be lost.
+</div>
+
+If your format is intended to be used only by machines,
+a binary format is likely to be more efficient,
+in addition to discouraging people from authoring or editing content.
+
 <h3 id="secure-context">Consider limiting new features to secure contexts</h3>
 
 Always limit your feature to secure contexts
--- a/index.bs
+++ b/index.bs
@@ -601,7 +601,7 @@ perhaps a textual format is not appropriate.
 SVG path syntax was designed to be compact, 
 with single-letter commands and unlabeled series of coordinates after them.
 At the time, file size was a primary concern, 
-but while the importance of this has changed over time,
+but while the importance of size efficiency has declined over time,
 the cost to human readability has remained the same.
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -589,11 +589,24 @@ See also:
 
 <h3 id=text-formats>Make textual formats for people</h3>
 
-Design textual formats to be both produced and consumed by people.
+Design textual formats that can be easily produced and consumed by people.
 
-People who are presented with a textual format will
-expect to be able to use a text editor
-to produce or modify content.
+Favor readability over compactness. 
+File size can be optimized by tooling,
+and tends to become a lower priority over time. 
+When file size is a significantly higher priority, 
+perhaps a textual format is not appropriate.
+
+<div class=example>
+SVG path syntax was designed to be compact, 
+with single-letter commands and unlabeled series of coordinates after them.
+At the time, file size was a primary concern, 
+but while the importance of this has changed over time,
+the cost to human readability has remained the same.
+</div>
+People who are presented with a textual format
+should be able to use a text editor
+to easily produce or modify content.
 People who edit text will introduce a range of errors, but
 could struggle to identify and fix those errors.
 
--- a/index.bs
+++ b/index.bs
@@ -604,6 +604,7 @@ At the time, file size was a primary concern,
 but while the importance of this has changed over time,
 the cost to human readability has remained the same.
 </div>
+
 People who are presented with a textual format
 should be able to use a text editor
 to easily produce or modify content.
--- a/index.bs
+++ b/index.bs
@@ -628,7 +628,7 @@ This makes JSON prone to syntactic errors as a result of manual edits.
 Containing the scope of a format that is affected by a syntax error
 could improve robustness and human usability.
 This requires that specifications fully define processing and
-error handling rules so that errors are handled in the same way.
+error handling so that all inputs result in consistent outcomes.
 
 <div class=example>
 Typos in CSS properties only cause that property to be ignored.
--- a/index.bs
+++ b/index.bs
@@ -613,8 +613,9 @@ People who edit text will introduce a range of errors, but
 could struggle to identify and fix those errors.
 
 People expect some amount of flexibility in terms of how their edits are processed.
-Allowing flexibility in whitespace, quoting, delimiters, and other syntactic elements ensures
-that content is still comprehensible.
+[[#avoid-ambiguity|Clearly defining]] syntactic flexibility--
+such as in white space, quoting, or delimiters--
+could ensure that content is both easy to edit and produces consistent results.
 
 <div class=example>
 The JSON format is an example of a format that appears to be flexible,
--- a/index.bs
+++ b/index.bs
@@ -639,7 +639,7 @@ Errors in properties rarely cause an entire rule to be lost.
 
 If your format is intended to be used only by machines,
 a binary format is likely to be more efficient,
-in addition to discouraging people from authoring or editing content.
+in addition to discouraging people from authoring or editing content directly.
 
 <h3 id="secure-context">Consider limiting new features to secure contexts</h3>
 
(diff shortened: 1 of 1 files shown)


hober (approved):
Looks good to me.

index.bs
```diff
@@ -587,6 +587,44 @@ See also:
...
+no commenting capability,
+objects and arrays cannot have trailing commas, and
+quotes are mandatory for object keys.
+This makes JSON prone to syntactic errors as a result of manual edits.
+</div>
+
+Containing the scope of a format that is affected by a syntax error
+could improve robustness and human usability.
+This requires that specifications fully define processing and
+error handling rules so that errors are handled in the same way.
```
plinss, line 617:
```suggestion
This requires that specifications define the data model first, and then fully define the processing and
error handling rules so that any arbitrary input byte stream will result in a consistent representation of the data.
```

index.bs
```diff
@@ -587,6 +587,44 @@ See also:
 * [[#secure-context]]
 * [[#consent]]
 
+<h3 id=text-formats>Make textual formats for people</h3>
```
LeaVerou, line 590: I think "make" reads a bit weird in this context. Maybe something like
```suggestion
<h3 id=text-formats>Textual formats should be usable by people</h3>
```
?

index.bs
```diff
@@ -587,6 +587,44 @@ See also:
...
+
+Design textual formats to be both produced and consumed by people.
```
LeaVerou, line 592:
```suggestion
Design textual formats that can be easily produced and consumed by people.
```

index.bs
```diff
@@ -587,6 +587,44 @@ See also:
...
+
+People who are presented with a textual format will
+expect to be able to use a text editor
+to produce or modify content.
+People who edit text will introduce a range of errors, but
+could struggle to identify and fix those errors.
+
+People expect some amount of flexibility in terms of how their edits are processed.
+Allowing flexibility in whitespace, quoting, delimiters, and other syntactic elements ensures
+that content is still comprehensible.
```
LeaVerou, line 602: This is not about comprehensibility, but ease of editing / error prevention.

Might be good to link to the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle) somewhere.

LeaVerou, line 602:
```suggestion
Syntactic [robustness](https://en.wikipedia.org/wiki/Robustness_principle) 
typically improves the usability of textual formats.
If user intent is unambiguous, flexibility in whitespace, quoting, delimiters, and other syntactic elements 
makes the syntax less error-prone and creates a more pleasant editing experience.
```

index.bs
```diff
@@ -587,6 +587,44 @@ See also:
 * [[#secure-context]]
 * [[#consent]]
 
+<h3 id=text-formats>Make textual formats for people</h3>
+
+Design textual formats to be both produced and consumed by people.
+
```
LeaVerou, line 593:
```suggestion

Favor readability over compactness. 
File size can be optimized by tooling,
and tends to become a lower priority over time. 
When file size is a significantly higher priority, 
perhaps a textual format is not appropriate.

<div class=example>
SVG path syntax was designed to be compact, 
with single-letter commands and unlabeled series of coordinates after them.
At the time, file size was a primary concern, 
but while the importance of this has changed over time,
the cost to human readability has remained the same.
</div>
```

index.bs
```diff
@@ -587,6 +587,44 @@ See also:
...
+People who are presented with a textual format will
+expect to be able to use a text editor
```
LeaVerou, line 595:
```suggestion
People who are presented with a textual format
should be able to use a text editor
```

It’s not about expectation. People often _expect_ terrible user experiences — that doesn't make them okay.

index.bs
```diff
@@ -587,6 +587,44 @@ See also:
...
+to produce or modify content.
```
LeaVerou, line 596:
```suggestion
to easily produce or modify content.
```

hober:
I think it might be worth adding a sentence linking this principle to [EWP §2.10 The web is transparent](https://www.w3.org/TR/ethical-web-principles/#transparent). Something along the lines of [this old blog post from Edd Dumbill](https://web.archive.org/web/20170113214807/http://times.usefulinc.com/2006/01/20-apple-markup):

> This is one reason that people prefer RELAX NG over XML Schema, for instance. Where markup is concerned, it turns out that the excuse "only computers will read it, and we'll provide tools to generate it" doesn't cut it. The web's had a view-source mentality since it started, and the aesthetics of markup matter a great deal.

LeaVerou:
I was just reminded of this PR after reading [this comment](https://github.com/w3c/csswg-drafts/issues/9889#issuecomment-2264151794) in the csswg repo.

I think we should actually _remove_ the specific guidance this currently has like:

> Allowing flexibility in whitespace, quoting, delimiters, and other syntactic elements ensures that content is still comprehensible.

Adding specific guidance in there muddles the core point of the principle, which is to **consider** human users and not design syntaxes that you expect to be consumed entirely by tooling. WRT what to _do_ to make the syntax easier for humans, that is literally what the whole Design Principles document is for!

Specific guidance implies scope, it implies that if you do these things mentioned here, you've complied with the principle. This is a higher level principle.

That said, we could perhaps list syntax flexibility as an example. But it's far from the only one, or even a core one. E.g. the SVG path syntax is actually quite flexible syntactically, but is *dreadful* for humans.

index.bs
martinthomson: s/Make/Design

```suggestion
<h3 id=text-formats>Design textual formats for people</h3>
```

We're using action statements throughout, and this is the best I had.  Yours ends up being a bit softer and not as imperative.  Don't we want them all to be "Do this", "Do that"?

index.bs
martinthomson: I literally wrote an RFC that says that the robustness principle harms interoperability: https://www.rfc-editor.org/rfc/rfc9413.html

I have friends who like to troll me with the phrase "Postel was wrong", but I'm guessing that didn't know :)

Either way, I think that this would be a mistake for the reasons stated in that document (see [this bit](https://www.rfc-editor.org/rfc/rfc9413#section-2.3-3)).  I think that your point about this being an example is a good one though.  I've suggested something below.

index.bs
martinthomson:
```suggestion
This requires that specifications fully define processing and
error handling so that all inputs result in consistent outcomes.
```

That's a lot of words.  Defining a data model is secondary so we can probably leave that out.  The key here is just that every implementation will produce the same output when given the same input.

It's not the case that arbitrary inputs always produce a good outcome, but that they are handled the same way by every implementation.

index.bs
```diff
@@ -587,6 +587,57 @@ See also:
...
+File size can be optimized by tooling,
+and tends to become a lower priority over time. 
+When file size is a significantly higher priority, 
+perhaps a textual format is not appropriate.
+
+<div class=example>
+SVG path syntax was designed to be compact, 
+with single-letter commands and unlabeled series of coordinates after them.
+At the time, file size was a primary concern, 
+but while the importance of this has changed over time,
```
martinthomson, line 604:
```suggestion
but while the importance of size efficiency has declined over time,
```

index.bs
```diff
@@ -587,6 +587,57 @@ See also:
...
+the cost to human readability has remained the same.
+</div>
+People who are presented with a textual format
```
martinthomson, line 607:
```suggestion

People who are presented with a textual format
```

index.bs
martinthomson:
```suggestion
People expect some amount of flexibility in terms of how their edits are processed.
Clearly defining syntactic flexibility--
such as in white space, quoting, or delimiters--
could ensure that content is both easy to edit and produces consistent results.
```

index.bs
```diff
@@ -587,6 +587,57 @@ See also:
 * [[#secure-context]]
 * [[#consent]]
 
+<h3 id=text-formats>Make textual formats for people</h3>
+
+Design textual formats that can be easily produced and consumed by people.
+
+Favor readability over compactness.
```
martinthomson, line 594:
```suggestion
Favor readability over compactness. 
Textual formats improve [transparency](https://www.w3.org/TR/ethical-web-principles/#transparent).
```

index.bs
```diff
@@ -587,6 +587,58 @@ See also:
 * [[#secure-context]]
 * [[#consent]]
 
+<h3 id=text-formats>Design textual formats for people</h3>
```
LeaVerou, line 590:
```suggestion
<h3 id=text-formats>Design textual formats for humans</h3>
```

index.bs
LeaVerou: This sentence reads a bit out of place here, though I have no concrete suggestion for where to move it :/

index.bs
LeaVerou: Much better!

index.bs
martinthomson: Yeah, I struggled with that too.  Let me try again.

index.bs
```diff
@@ -587,6 +587,58 @@ See also:
...
+
+Design textual formats that can be easily produced and consumed by people.
+
```
martinthomson, line 593:
```suggestion
Textual formats also improve [transparency](https://www.w3.org/TR/ethical-web-principles/#transparent).

```

hober (approved):
👍

hober:
> 👍

whoops, this was just a review of a one-line change mt proposed.

martinthomson:
Just noting here that we might want to have a followup for DRY and other sorts of ease of authoring affordances.  (#473 

We might also mention priority of constituencies to prioritize ease of authoring over ease of parsing. (new)

Semantic flexibility was something that we might want to break out into something more complete and nuanced. (new)

index.bs
```diff
@@ -587,6 +587,60 @@ See also:
...
+<div class=example>
+The JSON format is an example of a format that appears to be flexible,
+but it lacks many of these basic usability amenities.
+In JSON there is:
+no commenting capability,
+objects and arrays cannot have trailing commas, and
+quotes are mandatory for object keys.
+This makes JSON prone to syntactic errors as a result of manual edits.
+</div>
```
annevk, line 628: Despite this JSON will likely be the correct choice in quite a few scenarios. I worry a bit that this advice might lead people to design new formats which come with their own cost.

Issue #509, opened 2024-08-22
New Principle: Use `[Exposed=*]` to make only purely computational APIs available


ptomato:
### Problem statement

In [webidl#526](https://github.com/whatwg/webidl/pull/526) a new `Exposed` annotation, `[Exposed=*]`, was introduced. It denotes a fundamental set of interfaces that are intended to be exposed in Window, all Workers, all Worklets, ShadowRealm, as well as any future global scopes.

We've heard concerns (e.g., [tc39/proposal-shadowrealm#401](https://github.com/tc39/proposal-shadowrealm/issues/401)) that it's not clear when an interface should belong to this fundamental set and when it shouldn't. We'd like to issue guidance for this in a design principle.

I'm interested in feedback on the proposed guidance below, and am willing to write up the conclusions in a PR if it is generally positive.

### Proposed guidance

`[Exposed=*]` should be applied only to **purely computational interfaces**. That is, they do not perform I/O and do not affect the state of the user agent or the user's device.

Anything annotated with `[SecureContext]` should not be exposed everywhere; not all global scopes are secure contexts.

Anything relying on an event loop should not be exposed everywhere; not all global scopes have an event loop.

The `[Exposed=*]` annotation should also be applied conservatively. If an interface is not that useful without other interfaces that are not exposed everywhere, default to not exposing that interface as well.

### Further reading

Directly related, but predates `[Exposed=*]`: [#35](https://github.com/w3ctag/design-principles/issues/35)

Discussion resulting in the addition of `[Exposed=*]`: [webidl#468](https://github.com/whatwg/webidl/issues/468)

Other relevant reading: [#325](https://github.com/w3ctag/design-principles/issues/325), [#360](https://github.com/w3ctag/design-principles/issues/360), [#448](https://github.com/w3ctag/design-principles/issues/448), [tc39/ecma262#1120](https://github.com/tc39/ecma262/issues/1120), [WebAudio/web-audio-api#2499](https://github.com/WebAudio/web-audio-api/issues/2499), [tc39/proposal-shadowrealm#398](https://github.com/tc39/proposal-shadowrealm/issues/398)

zcorpan:
cc @smaug---- @mgaudet

annevk:
This sounds reasonable. I'm not sure if it needs to be a principle or could just be advice in the Web IDL standard. It might also be helpful to have some examples so people can evaluate whether they understand the rules correctly.

ptomato:
[...]

PR #510, opened 2024-09-12
New text for guidance on exposing interfaces everywhere

ptomato:
Here's an initial draft for text on exposing interfaces everywhere. It's similar to the text I wrote in #509 and the examples I gave there.

Closes: #509

***
<a href="https://pr-preview.s3.amazonaws.com/ptomato/w3ctag-design-principles/pull/510.html" title="Last updated on Oct 4, 2024, 1:02 AM UTC (0777bc0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/510/55a06e0...ptomato:0777bc0.html" title="Last updated on Oct 4, 2024, 1:02 AM UTC (0777bc0)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2756,6 +2756,68 @@ because use of {{ScriptProcessorNode}} from the main thread
 frequently resulted in a poor user experience. [[WebAudio]]
 </p>
 
+## Only purely computational features should be exposed everywhere ## {#expose-everywhere}
+<!--
+    https://github.com/w3ctag/design-principles/issues/509
+    https://github.com/whatwg/webidl/issues/468
+    https://github.com/w3ctag/design-principles/issues/35
+-->
+
+When exposing a feature,
+please consider whether it makes sense
+to expose the feature to all possible environments
+(via the {{Exposed|[Exposed=*]}} annotation
+or including it on all global scope interfaces).
+
+Only purely computational features should be exposed everywhere.
+That is,
+they do not perform I/O
+and do not affect the state of the user agent or the user’s device.
+
+<p class="example">
+The {{TextEncoder}} interface converts a string to UTF-8 encoded bytes.
+This is a purely computational interface,
+generally useful as a JavaScript language facility,
+so it should be exposed everywhere.
+</p>
+
+<p class="example">
+<a attribute spec=html>localStorage</a> affects the state of the user agent, so it should not be exposed everywhere.
+</p>
+
+<p class="example">
+Technically, {{console}} could affect the state of the user agent
+(by causing log messages to appear in the developer tools)
+or the user’s device
+(by writing to a log file.)
+But these things are not observable from the running code,
+and the practicality of having {{console}} everywhere outweighs the disadvantages.
+</p>
+
+Additionally, anything relying on an event loop
+should not be exposed everywhere.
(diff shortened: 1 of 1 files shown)


caridy:
this is great, thanks @ptomato for putting the time on this

index.bs
```diff
@@ -2756,6 +2756,70 @@ because use of {{ScriptProcessorNode}} from the main thread
...
+(by writing to a log file.)
+But these things are not observable from the running code,
+and the practicality of having {{console}} everywhere outweighs the disadvantages.
+</p>
+
+Additionally, anything annotated with {{SecureContext}},
+or relying on an event loop,
+should not be exposed everywhere.
+Not all global scopes are secure contexts
+or have an event loop.
```
jyasskin, line 2801: I'm not sure about excluding [`[SecureContext]`](https://webidl.spec.whatwg.org/#SecureContext) things from `[Exposed=*]`. Won't `[SecureContext]` itself exclude those items from global scopes where they'd be inappropriate?

```suggestion
Additionally, anything relying on an event loop
should not be exposed everywhere.
Not all global scopes have an event loop.
```

jyasskin (approved):
We should think some about the [SecureContext] question, but otherwise this looks ready to go in.

jyasskin:
@zcorpan Intuitively I disagree with your argument in https://github.com/whatwg/webidl/issues/1119#issuecomment-2034009609, at least given the guidance in this PR. The downside of requiring review for every new kind of realm is that when we add a new realm type, there's going to be a long string of paper-cuts as developers reach for various utility functions and notice that they haven't yet been added to this particular realm.

index.bs
ptomato: Thanks for the review!

My motivation for this was to get exactly the same set of exposed APIs in ShadowRealm regardless of which global scope created the ShadowRealm. I could go either way on this; we could remove this sentence from the guidance and specify in https://github.com/whatwg/html/pull/9893 that ShadowRealm scopes are never secure contexts. That would have the same result, I think.

bakkot: Since browsers are going to differ in which APIs they implement anyway, I don't think there's that much reason to worry about getting exactly the same set of exposed APIs in ShadowRealm. By contrast, I do think it's important that pure-computation APIs which happen to be restricted to secure contexts, like [`crypto.randomUUID()`](https://github.com/WICG/uuid/issues/23), should be available in ShadowRealms when they are themselves running in secure contexts.

The stated intention of browsers in restricting features to secure-context-only is to encourage the use of secure contexts for all new applications, with this being assumed (IMO incorrectly) not to be a very large cost to the usability of the feature because applications should be in secure contexts anyway so developers following best practices (i.e. secure contexts) shouldn't even notice. That will not be true if we have this restriction here.

ptomato: I'm not sure I agree, but I will at least remove `[SecureContext]` from here so that discussion becomes irrelevant to this PR and we can continue talking about whether a ShadowRealm is a secure context in the HTML PR.

ptomato:
What should be the next steps here? As far as I'm concerned, the secure context question no longer needs to be part of this PR.

martinthomson (approved):

PR #516, opened 2024-11-28
Remove extensible web text

martinthomson:
This looks to be a stub or early draft of https://extensiblewebmanifesto.org/

It's old and probably doesn't need to be kept here.


LeaVerou (approved):

torgo (approved):
yes this is fine

PR #536, opened 2024-12-06
Improve the run-to-completion principle.

### LINKED ISSUE #456: "while a JavaScript event loop is running." is probably not intended
https://w3ctag.github.io/design-principles/#js-rtc

> Don’t modify data accessed via JavaScript APIs while a JavaScript [event loop](https://html.spec.whatwg.org/multipage/webappapis.html#event-loop) is running.

An event loop is running from the creation of an agent, and it doesn't stop until the agent is destroyed. By this definition, data can never really update.

This should probably be reworded in terms of [running scripts](https://html.spec.whatwg.org/multipage/webappapis.html#calling-scripts) and [callback invoking](https://webidl.spec.whatwg.org/#invoke-a-callback-function).

It might be worth calling out that `new Date()`, `Date.now()`, `performance.now()` are accepted breakages of this rule.

In fact, it's interesting to compare `performance.now()` to `document.timeline.currentTime`, as the former breaks run-to-completion, whereas the latter is stable for the whole frame (beyond a single callback).

jyasskin:
* "While an event loop is running" wasn't the right period to limit changes.
* We have a list of exceptions to this principle.

This is meant to fix #456; @jakearchibald, can you check it? Thank you for the text!

I think this also supersedes #496. @jan-ivar, have I caught everything you meant to say there?

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/design-principles/pull/536.html#js-rtc" title="Last updated on Dec 17, 2024, 10:26 PM UTC (21262c9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/536/6c1568b...jyasskin:21262c9.html" title="Last updated on Dec 17, 2024, 10:26 PM UTC (21262c9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1478,7 +1478,10 @@ Data can update synchronously from the result of developer action.
 and is immediately observable.
 </div>
 
-A few kinds of situations justify violating this rule:
+A few kinds of situations justify violating this rule.
+In these cases,
+be sure to [[#attributes-like-data|expose values as methods rather than attributes
+if they might change in the middle of a JS statement]].
 
 * Observing the current time,
     as in {{Date/now|Date.now()}} and {{Performance/now|performance.now()}},
@@ -1488,6 +1491,8 @@ A few kinds of situations justify violating this rule:
     as in the case of the proposed {{isInputPending()}}.
 * States meant to protect users from surprising UI changes,
     like [=transient activation=].
+    Note that {{UserActivation/isActive|navigator.userActivation.isActive}}
+    violates [[#attributes-like-data]] and should not.
 
 <h3 id="js-gc">Don't expose garbage collection</h3>
 
--- a/index.bs
+++ b/index.bs
@@ -47,6 +47,7 @@ spec:webidl
     type:idl; text:long
     type:idl; text:short
     type:interface; text:double
+spec:web-locks; type:interface; text:LockManager
 </pre>
 <pre class='ignored-specs'>
 spec: css21
--- a/index.bs
+++ b/index.bs
@@ -1461,7 +1461,7 @@ During a `while (true)`,
 and after `await`ing an already-resolved `Promise`,
 you *shouldn't* expect things like:
 
-* The DOM to update as a result of the HTML parser loading network content
+* The DOM to update as a result of the HTML parser loading new content from the network
 * {{HTMLImageElement/width|img.width}} to change as a result of loading image data from the network
 * Buttons of a {{Gamepad}} to change state
 * {{Element/scrollTop}} to change, even if scrolling can visually occur
@@ -1492,7 +1492,7 @@ if they might change in the middle of a JS statement]].
 * States meant to protect users from surprising UI changes,
     like [=transient activation=].
     Note that {{UserActivation/isActive|navigator.userActivation.isActive}}
-    violates [[#attributes-like-data]] and should not.
+    violates [[#attributes-like-data|the guidance that recommends a method for this case]].
 
 <h3 id="js-gc">Don't expose garbage collection</h3>
 
--- a/index.bs
+++ b/index.bs
@@ -1437,9 +1437,6 @@ to have bindings in other programming languages.
 
 <h3 id="js-rtc">Preserve run-to-completion semantics</h3>
 
-Don't modify data accessed via JavaScript APIs
-while a JavaScript <a>event loop</a> is running.
-
 A JavaScript Web API is generally a wrapper around
 a feature implemented in a lower-level language,
 such as C++ or Rust.
@@ -1452,18 +1449,45 @@ Because of that, JavaScript authors take for granted
 that the data available to a function won’t change unexpectedly
 while the function is running.
 
-So if a JavaScript Web API exposes some piece of data,
-such as an object property,
-the user agent must not update that data
-while a JavaScript task is running.
-Instead, if the underlying data changes,
-<a>queue a task</a> to modify the exposed version of the data.
+Changes that are not the result of developer action
+and changes that are asynchronously delivered
+should not happen in the middle of other JavaScript,
+including between [=microtasks=].
+Instead, a task should be [[html#queuing-tasks|queued]],
+or they should update as part of [=update the rendering=].
 
 <div class="example">
-If a JavaScript task has accessed the {{NavigatorOnline/onLine|navigator.onLine}} property,
-and browser's online status changes,
-the property won't be updated until the next task runs.
+During a `while (true)`,
+and after `await`ing an already-resolved `Promise`,
+you *shouldn't* expect things like:
+
+* The DOM to update as a result of the HTML parser loading network content
+* {{HTMLImageElement/width|img.width}} to change as a result of loading image data from the network
+* Buttons of a {{Gamepad}} to change state
+* {{Element/scrollTop}} to change, even if scrolling can visually occur
+
+These things aren't updated by the currently running script,
+so they shouldn't change during the current task.
+
 </div>
+
+Data can update synchronously from the result of developer action.
+
+<div class="example">
+{{ChildNode/remove()|node.remove()}} changes the DOM synchronously
+and is immediately observable.
+</div>
+
+A few kinds of situations justify violating this rule:
+
+* Observing the current time,
+    as in {{Date/now|Date.now()}} and {{Performance/now|performance.now()}},
+    although note that it's also useful to present a consistent task-wide time
+    as in {{AnimationTimeline/currentTime|document.timeline.currentTime}}.
+* Functions meant to help developers interrupt synchronous work,
+    as in the case of the proposed {{isInputPending()}}.
+* States meant to protect users from surprising UI changes,
+    like [=transient activation=].
 
 <h3 id="js-gc">Don't expose garbage collection</h3>
 
--- a/index.bs
+++ b/index.bs
@@ -1442,14 +1442,10 @@ propagate that change to JavaScript between tasks,
 for example by [[html#queuing-tasks|queuing a task]],
 or as part of [=update the rendering=].
 
-A JavaScript Web API is generally a wrapper around
-a feature implemented in a lower-level language,
-such as C++ or Rust.
-Unlike those languages,
-when using JavaScript developers can expect
-that once a piece of code begins executing,
-it will continue executing until it has completed.
-
+Unlike lower-level languages
+such as C++ or Rust,
+JavaScript has historically acted as if
+only one piece of code can execute at once.
 Because of that, JavaScript authors take for granted
 that the data available to a function won’t change unexpectedly
 while the function is running.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1444,19 +1441,46 @@ Because of that, JavaScript authors take for granted
...
 </div>
 
+A few kinds of situations justify violating this rule:
+
+* Observing the current time,
+    as in {{Date/now|Date.now()}} and {{Performance/now|performance.now()}},
+    although note that it's also useful to present a consistent task-wide time
+    as in {{AnimationTimeline/currentTime|document.timeline.currentTime}}.
+* Functions meant to help developers interrupt synchronous work,
+    as in the case of the proposed {{isInputPending()}}.
```
annevk, line 1480: This feature is somewhat controversial because it violates run-to-completion.

index.bs
```diff
@@ -1444,19 +1441,46 @@ Because of that, JavaScript authors take for granted
...
+* States meant to protect users from surprising UI changes,
+    like [=transient activation=].
```
annevk, line 1482: I don't think you can get in a state there where `x.isActive !== x.isActive`?

index.bs
```diff
@@ -1444,19 +1441,46 @@ Because of that, JavaScript authors take for granted
...
+{{ChildNode/remove()|node.remove()}} changes the DOM synchronously
+and is immediately observable.
 </div>
 
+A few kinds of situations justify violating this rule:
+
+* Observing the current time,
+    as in {{Date/now|Date.now()}} and {{Performance/now|performance.now()}},
+    although note that it's also useful to present a consistent task-wide time
+    as in {{AnimationTimeline/currentTime|document.timeline.currentTime}}.
```
jakearchibald, line 1478: I think this is wider than a task. I think it's stable for a frame. But I guess that also means it's stable for a task.

index.bs
jakearchibald: Feels reasonable to drop this reference. The other examples are enough.

jakearchibald:
Generally LGTM!

index.bs
jyasskin: If I'm reading it correctly, Chromium's `isActive` bottoms out at https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/common/frame/user_activation_state.cc;l=81;drc=616d60fca655937c2b730db94fd32d37ddff3bb5

```c++
bool UserActivationState::IsActiveInternal() const {
  return base::TimeTicks::Now() <= transient_state_expiry_time_;
}
```

I didn't notice any caching, which means that can return false immediately after it returned true, within the same line of JS code. @mustaqahmed, can you confirm whether I'm reading that correctly? 

This intra-task change also seems to be the semantics that @jan-ivar is saying should be acceptable in https://github.com/w3ctag/design-principles/pull/496/files#r1820794873.

FWIW, putting users ahead of authors in this way seems to be correct to me.

index.bs
jyasskin: The TAG is on record supporting this exception in https://github.com/w3ctag/design-reviews/issues/415#issuecomment-527655955 and https://github.com/w3ctag/design-reviews/issues/475#issuecomment-621438616 (cc @dbaron for any historical context). I can break this into a separate PR if we need to discuss it more, but this function seems to be a case of putting authors ahead of technical purity. There might also be a better phrasing to explain what makes this exception acceptable?

index.bs
jyasskin: Yeah ... I think this phrasing is probably ok, but I could word it more precisely if you think it's likely to confuse people?

jyasskin (commented):
Thanks for the reviews!

index.bs
annevk: I guess you're right. It seems really bad that this is a getter then and not a method.

jyasskin: I've pointed that out in fac9f74fecf71a3469fc05ba6951111ab4f22be6, although it's possible that's too much emphasis.

index.bs
```diff
@@ -1444,19 +1441,51 @@ Because of that, JavaScript authors take for granted
...
 </div>
 
+Data can update synchronously from the result of developer action.
+
+<div class="example">
+{{ChildNode/remove()|node.remove()}} changes the DOM synchronously
+and is immediately observable.
+</div>
+
+A few kinds of situations justify violating this rule.
```
dbaron, line 1473: FYI, the recent edit here makes it a little less clear that the list following is a list of examples.

dbaron (commented):
For what it's worth, I'm generally supportive of these changes.  I think the core of this principle is about preventing surprising behavior and preventing broken code that results from that surprise.  In that vein, the clearer it is that the functions have this sort of behavior, the better.  (For example, it's reasonably clear that something like `Date.now()` might change as slow code executes.)

For things that do violate this, we want it to be pretty clear that code (conceptually, maybe not syntactically) like this is bad:

```js
if (window.isPurple) {
  // (1)
} else {
  // (2)
}

// ...

if (window.isPurple) {
  // assume path (1) above executed
}
```

index.bs
```diff
@@ -1444,19 +1441,51 @@ Because of that, JavaScript authors take for granted
...
 
 <div class="example">
-If a JavaScript task has accessed the {{NavigatorOnline/onLine|navigator.onLine}} property,
-and browser's online status changes,
-the property won't be updated until the next task runs.
+During a `while (true)`,
+and after `await`ing an already-resolved `Promise`,
+you *shouldn't* expect things like:
+
+* The DOM to update as a result of the HTML parser loading network content
```
martinthomson, line 1456:
```suggestion
* The DOM to update as a result of the HTML parser loading new content from the network

```

index.bs
```diff
@@ -1444,19 +1441,51 @@ Because of that, JavaScript authors take for granted
...
+* Observing the current time,
+    as in {{Date/now|Date.now()}} and {{Performance/now|performance.now()}},
+    although note that it's also useful to present a consistent task-wide time
+    as in {{AnimationTimeline/currentTime|document.timeline.currentTime}}.
+* Functions meant to help developers interrupt synchronous work,
+    as in the case of the proposed {{isInputPending()}}.
+* States meant to protect users from surprising UI changes,
+    like [=transient activation=].
+    Note that {{UserActivation/isActive|navigator.userActivation.isActive}}
+    violates [[#attributes-like-data]] and should not.
```
martinthomson, line 1487:
```suggestion
    violates [[#attributes-like-data|the guidance that recommends a method for this case]].
```

index.bs
```diff
@@ -1444,19 +1441,51 @@ Because of that, JavaScript authors take for granted
...
+
+<div class="example">
+{{ChildNode/remove()|node.remove()}} changes the DOM synchronously
+and is immediately observable.
+</div>
+
+A few kinds of situations justify violating this rule.
+In these cases,
+be sure to [[#attributes-like-data|expose values as methods rather than attributes
+if they might change in the middle of a JS statement]].
```
jyasskin, line 1476: Tentatively, to restore the clarity that the list contains the acceptable violations:

```suggestion
A few kinds of situations justify violating this rule:
```

The note next to 'transient activation' below is probably enough.

index.bs
jyasskin: I've undone that edit.

index.bs
```diff
@@ -1444,19 +1441,48 @@ Because of that, JavaScript authors take for granted
...
+should not happen in the middle of other JavaScript,
+including between [=microtasks=].
+Instead, a task should be [[html#queuing-tasks|queued]],
+or they should update as part of [=update the rendering=].
 
 <div class="example">
-If a JavaScript task has accessed the {{NavigatorOnline/onLine|navigator.onLine}} property,
-and browser's online status changes,
-the property won't be updated until the next task runs.
+During a `while (true)`,
```
martinthomson, line 1452:
```suggestion
During synchronous execution (such as a `while` loop),
```

index.bs
```diff
@@ -1429,9 +1429,6 @@ to have bindings in other programming languages.
 
 <h3 id="js-rtc">Preserve run-to-completion semantics</h3>
 
-Don't modify data accessed via JavaScript APIs
-while a JavaScript <a>event loop</a> is running.
-
```
martinthomson, line 1434: I think that we need a positive statement for this principle.

Maybe

> Propagate changes to state
> that originate outside of JavaScript execution context
> between tasks, not within them.

martinthomson (approved):

index.bs
jyasskin: Yeah, or maybe by moving up the "a task should be queued" suggestion below...

> If a change to state originates outside of the JavaScript execution context,
> propagate that change to JavaScript between tasks,
> for example by [[html#queuing-tasks|queuing a task]],
> or as part of [=update the rendering=].

jan-ivar (commented):
Thanks, this looks like a great improvement to me!

I'd like to continue to discuss https://github.com/w3ctag/design-principles/pull/496 separately for a little bit if that's OK, but that doesn't need to block this. It may be redundant, but I'd like to double-check there isn't a principle there worth describing specific to designing promise-returning APIs that come with synchronous requirements.

index.bs
```diff
@@ -1444,19 +1441,48 @@ Because of that, JavaScript authors take for granted
...
-the property won't be updated until the next task runs.
+During synchronous execution (such as a `while` loop),
+and after `await`ing an already-resolved `Promise`,
+you *shouldn't* expect things like:
+
+* The DOM to update as a result of the HTML parser loading new content from the network
+* {{HTMLImageElement/width|img.width}} to change as a result of loading image data from the network
+* Buttons of a {{Gamepad}} to change state
+* {{Element/scrollTop}} to change, even if scrolling can visually occur
+
```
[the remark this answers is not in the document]
jyasskin: Yes, good point that we should specifically include changes in method behavior. I think the wording you've suggested implies that folks might be tempted to identify microtasks, which they generally aren't, so I'd like to call out a particular function. I had trouble finding a synchronous method that does actually change behavior depending on queued changes, and ones that return Promises can always return in a new task, so I think I have to use a counterfactual:

```suggestion
* A synchronous method to act differently depending on asynchronous state changes. For example, if {{LockManager}} had synchronous methods, their behavior would depend on concurrent calls in other windows.

```

martinthomson (approved):

index.bs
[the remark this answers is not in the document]
jakearchibald: > A method acting differently depending on which [=microtask=] it is called on.

fwiw, a close example of this is `requestAnimationFrame`. It makes sense that it sometimes runs the callback before the next frame, and sometimes after, but it's annoying that you don't know which of these will happen. https://github.com/whatwg/html/issues/10113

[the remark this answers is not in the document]
jakearchibald: Yeah, honestly, I think my example shows that:

> A method acting differently depending on which [=microtask=] it is called on.

…isn't the right terminology.

I think `requestAnimationFrame` is ok, it's just missing that bit that tells you in advance if it'll run before or after the next frame.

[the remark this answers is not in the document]
jyasskin: Sounds good. I think that means https://w3ctag.github.io/design-principles/#js-rtc is basically correct, and there's nothing more for me to do on this PR. Let me know or file an issue if I have that wrong. :)

PR #542, opened 2024-12-13
Add a section about looking for guidance in other specs.

### LINKED ISSUE #303: New principle: Patterns for URLs in APIs
This was spun out of the discussion here: https://github.com/w3ctag/design-principles/issues/278

URLs are used all over the web platform. We unfortunately don't have any guidance on it aside from the occasional one-off comment when we see an antipattern in a review.

1. Best practices for APIs which accept or emit URLs
2. Best practices for URLs as part of state
3. Guidance on what to do when one needs to implement a URL-like object (e.g. https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidate)

jyasskin:
Fixes #303.

I'm happy to add more links to useful guidance either in this PR or in future ones.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/design-principles/pull/542.html#specs-include-guidance" title="Last updated on Dec 17, 2024, 7:22 PM UTC (f980d22)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/542/6c1568b...jyasskin:f980d22.html" title="Last updated on Dec 17, 2024, 7:22 PM UTC (f980d22)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -544,6 +544,20 @@ i.e. consistency within CSS.
 
 There is also a separate section on <a href="#naming-consistency">naming consistency</a>.
 
+<h3 id="specs-include-guidance">Follow guidance from feature specifications</h3>
+
+The specifications for individual features
+should include guidance for using those features
+consistently across other specifications.
+When using a feature,
+look for that guidance and follow it.
+An incomplete list of such guidance follows:
+
+* [[encoding#specification-hooks|Encoding]]
+* [[fetch#fetch-elsewhere|Fetch]]
+* [[hr-time#sec-tools|Time]]
+* [[url#url-apis-elsewhere|URLs]]
+
 <h3 id="feature-detect">New features should be detectable</h3>
 
 Provide a way for authors to programmatically detect
--- a/index.bs
+++ b/index.bs
@@ -562,6 +562,7 @@ An incomplete list of such guidance follows:
 * [[fetch#fetch-elsewhere|Fetch]]
 * [[hr-time#sec-tools|Time]]
 * [[url#url-apis-elsewhere|URLs]]
+*  [[#using-http]]
 
 <h3 id="feature-detect">New features should be detectable</h3>
 
--- a/index.bs
+++ b/index.bs
@@ -562,7 +562,13 @@ An incomplete list of such guidance follows:
 * [[fetch#fetch-elsewhere|Fetch]]
 * [[hr-time#sec-tools|Time]]
 * [[url#url-apis-elsewhere|URLs]]
-*  [[#using-http]]
+* [[#using-http]]
+
+Consult with the relevant community
+when using their specification.
+This informs that community about new usage
+and creates opportunities for improvements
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -536,6 +536,20 @@ i.e. consistency within CSS.
...
 There is also a separate section on <a href="#naming-consistency">naming consistency</a>.
 
+<h3 id="specs-include-guidance">Follow guidance from feature specifications</h3>
+
+The specifications for individual features
+should include guidance for using those features
+consistently across other specifications.
+When using a feature,
+look for that guidance and follow it.
+An incomplete list of such guidance follows:
```
martinthomson, line 546:
```suggestion
When using a feature,
follow guidance in the specification for that feature.

Specifications for features
that will be used by other specifications
should include guidance for using those features.
This ensures that the feature is used correctly
and consistently
throughout the platform.

An incomplete list of such guidance follows:
```

martinthomson (commented):
Seems reasonable to say, though I think we need to change the formulation to match house style.

index.bs
cynthia: I strongly prefer this version over the original PR.

index.bs
```diff
@@ -536,6 +536,25 @@ i.e. consistency within CSS.
...
+and consistently
+throughout the platform.
+
+An incomplete list of such guidance follows:
+
+* [[encoding#specification-hooks|Encoding]]
+* [[fetch#fetch-elsewhere|Fetch]]
+* [[hr-time#sec-tools|Time]]
+* [[url#url-apis-elsewhere|URLs]]
+
```
annevk, line 557: Maybe add something along these lines:

> Also strongly consider asking feedback from the relevant specification community to better inform your own work and ensure it's aligned with the latest best practices. Often this as easy as filing a GitHub issue.

annevk (commented):
Nice!

index.bs
martinthomson, line 557:
```suggestion
*  [[#using-http]]

```

martinthomson: Or

> Consult with the relevant community
> when using their specification.
> This informs that community about new usage
> and creates opportunities for improvements
> to both the using and used specifications.

This avoids indirect acknowledgment that the documentation for usage might not be tracking best practices (when does it ever?).  I'm going for even more indirect, though I think that the goal of mutual benefit is a better angle.

The GitHub issue thing is a detail that is relevant only in some communities.

jyasskin: Done.

jyasskin: I've added this, but I wonder if, in the interest of conciseness, we should put
> [Building Protocols with HTTP](https://httpwg.org/specs/rfc9205.html) and [other HTTP RFCs](https://datatracker.ietf.org/doc/search?name=HTTP&sort=&rfcs=on&by=group&group=httpbis)

here and remove the #using-http section.

martinthomson (approved):

PR #543, opened 2024-12-13
Fix the bikeshed warnings.

jyasskin:
***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/design-principles/pull/543.html" title="Last updated on Dec 13, 2024, 5:58 AM UTC (c69cd61)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/543/1a74419...jyasskin:c69cd61.html" title="Last updated on Dec 13, 2024, 5:58 AM UTC (c69cd61)">Diff</a>

PR #544, opened 2024-12-13
Switch the GH Action to spec-prod.

jyasskin:


.github/workflows/deploy.yml
```diff
@@ -1,21 +1,18 @@
...
       with:
-        python-version: '3.10' 
-    - run: pip install bikeshed
-    - run: bikeshed update --skip-manifest
-    - run: make
-    - uses: peaceiris/actions-gh-pages@v3
-      with:
-        github_token: ${{ secrets.GITHUB_TOKEN }}
-        publish_dir: build
+        BUILD_FAIL_ON: fatal
```
martinthomson, line 17: I would say that this is probably better as 

```suggestion
        BUILD_FAIL_ON: warning
```

Based on #543.

jyasskin: Yep, if #543 goes in first, we should make this change here, and otherwise we should make it in #543.

jyasskin: Ok, done.

plinss (approved):

PR #555, opened 2025-02-19
Clarify the strings for enums/constants principle

### LINKED ISSUE #427: Clarify which enum the principle 6.11 is talking about
https://www.w3.org/TR/design-principles/#string-constants

> 6.11. Use strings for constants and enums
> 
> If your API needs a constant, or a set of enumerated values, use string values.
> 
> Strings are easier for developers to inspect, and in JavaScript engines there is no performance benefit from using integers instead of strings.
> 
> If you need to express a state which is a combination of properties, which might be expressed as a bitmask in another language, use a dictionary object instead. This object can be passed around as easily as a single bitmask value.

This sounds much like it's talking about C++ enums, right? Since there are also Web IDL enums that are strings, can this be clarified? It may also sound like discouraging against IDL enums as in https://github.com/WICG/compression/issues/50#issuecomment-1479097825.

martinthomson:
Adds examples and tightens things up.

Adds a mention that WebIDL enums only take string values.

One thing that jumped out here was that WebIDL permits the use of numeric/integer constants.  I searched all of the Firefox codebase and I didn't find a single useful instance where a numeric constant made sense.

My operating theory before starting was that there might be some case where developers might be exposed to a number in contexts outside of the web, which might mean that using a numeric constant in the platform made sense.

The one example of that we have is KeyboardEvent.keyCode, which might have made sense if it exposed the Unicode codepoint of a key.  That, however, is solidly deprecated in favor of a much better API.

Every other instance was a bad example.

Closes #427.

***
<a href="https://pr-preview.s3.amazonaws.com/martinthomson/design-principles/pull/555.html" title="Last updated on Mar 4, 2025, 3:37 PM UTC (4d89cad)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/555/af2965b...martinthomson:4d89cad.html" title="Last updated on Mar 4, 2025, 3:37 PM UTC (4d89cad)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1998,12 +1998,12 @@ is a request, rather than a guarantee.
 
 Use strings as the values for constants or sets of enumerated values.
 
-Strings are easier for developers to inspect values
+Strings make it easier for developers to inspect values
 and read code that uses those values.
 In JavaScript engines there is no performance benefit
 to using integers instead of strings.
-The [WebIDL enum type](https://webidl.spec.whatwg.org/#idl-enums)
-only permits the use of string values.
+The values of a [WebIDL enum type](https://webidl.spec.whatwg.org/#idl-enums)
+are strings in order to follow this principle.
 
 If you need to express a state which is a combination of properties,
 which might be expressed as a bitmask in another language,
--- a/index.bs
+++ b/index.bs
@@ -1996,17 +1996,31 @@ is a request, rather than a guarantee.
 
 <h3 id="string-constants">Use strings for constants and enums</h3>
 
-If your API needs a constant, or a set of enumerated values,
-use string values.
+Use strings as the values for constants or sets of enumerated values.
 
-Strings are easier for developers to inspect,
-and in JavaScript engines there is no performance benefit
-from using integers instead of strings.
+Strings are easier for developers to inspect values
+and read code that uses those values.
+In JavaScript engines there is no performance benefit
+to using integers instead of strings.
+The [WebIDL enum type](https://webidl.spec.whatwg.org/#idl-enums)
+only permits the use of string values.
 
 If you need to express a state which is a combination of properties,
 which might be expressed as a bitmask in another language,
 use a dictionary object instead.
 This object can be passed around as easily as a single bitmask value.
+
+<div class=example>
+The type of a dedicated Worker is set using an enumerated {{WorkerType}} argument.
+This has two values ("classic" and "module") and might have used a Boolean type instead.
+The use of strings costs nothing and makes code clearer.
+</div>
+
+<div class=example>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2043,18 +2044,32 @@ is a request, rather than a guarantee.
...
 <h3 id="string-constants">Use strings for constants and enums</h3>
 
-If your API needs a constant, or a set of enumerated values,
-use string values.
+Use strings as the values for constants or sets of enumerated values.
 
-Strings are easier for developers to inspect,
-and in JavaScript engines there is no performance benefit
-from using integers instead of strings.
+Strings are easier for developers to inspect values
```
jyasskin, line 2049:
```suggestion
Strings make it easier for developers to inspect values
```

index.bs
```diff
@@ -2043,18 +2044,32 @@ is a request, rather than a guarantee.
...
+to using integers instead of strings.
+The [WebIDL enum type](https://webidl.spec.whatwg.org/#idl-enums)
+only permits the use of string values.
 
 If you need to express a state which is a combination of properties,
 which might be expressed as a bitmask in another language,
 use a dictionary object instead.
 This object can be passed around as easily as a single bitmask value.
 
+<div class=example>
```
jyasskin, line 2061: Add `id`s to the examples, to make deep links more stable.

index.bs
```diff
@@ -2043,18 +2044,32 @@ is a request, rather than a guarantee.
...
 
-Strings are easier for developers to inspect,
-and in JavaScript engines there is no performance benefit
-from using integers instead of strings.
+Strings are easier for developers to inspect values
+and read code that uses those values.
+In JavaScript engines there is no performance benefit
+to using integers instead of strings.
+The [WebIDL enum type](https://webidl.spec.whatwg.org/#idl-enums)
+only permits the use of string values.
```
jyasskin, line 2054: This could be more attached to the rest of the paragraph ... maybe

```suggestion
The values of a [WebIDL enum type](https://webidl.spec.whatwg.org/#idl-enums)
are strings in order to follow this principle.
```

index.bs
martinthomson: I'm going to suggest that we do that systemically, rather than start here.  (I am not sure that examples need that sort of link stability, but if we want that sort of stability, then we should do it properly.)

jyasskin (approved):

annevk:
> One thing that jumped out here was that WebIDL permits the use of numeric/integer constants.

That's because Web IDL has to account for legacy APIs, not because we thought it was a good idea. However, the graphics world does make heavy use of these and we could not really convince them otherwise. https://github.com/whatwg/webidl/issues/717 captures one such discussion.

torgo (approved):

PR #562, opened 2025-03-07
Principle: Write only one algorithm to accomplish a task.

### LINKED ISSUE #239: New principle: Discourage polyglot formats
This is a generalization of @dbaron's concern in #128.

Polyglot formats tend to lead to interoperability problems, so we should discourage defining them.

Polyglot formats are formats which are defined such that they can be processed by two or more different kinds of processors with roughly equivalent results. For instance, it's possible to write a computer program that is simultaneously valid C and valid C++. Another example is [Polyglot Markup](https://www.w3.org/TR/html-polyglot/), an abandoned attempt to define a markup syntax that was simultaneously valid HTML and XHTML, and whose documents would produce roughly equivalent DOM trees when parsed with an HTML parser or an XML parser.

Authors tend to test their document with only one kind of processor, so they inadvertently introduce errors which would only be caught by the other kind of processor. In the case of Polyglot Markup, this happened when authors introduced XML errors into their document but only tested with an HTML parser. Consumers using an XML parser would, instead of seeing the document, see an XML parser error screen.

If the polyglot format contains fields that are only used by one kind of processor, such fields are likely to experience bit rot problems when authors only routinely test their documents with the other kind of processor. (For instance, if authors routinely use JSON parsers to test their JSON-LD, the `@context` section is likely to experience bit rot. Downstream consumers of the document who use a JSON-LD processor will start encountering bugs. They'll report it upstream, and that person will say "it works for me, it must be a bug in your software".)

jyasskin:
This explains why and when "polyglot" formats are a bad idea.

Fixes #239.

There's some overlap between this and the preceding section, [Resolving tension between interoperability and implementability](https://www.w3.org/TR/design-principles/#implementability). Do y'all think it's ok, or are there bits we could refactor together?

I'd also like to give an example of parsing divergence yielding security bugs, but I didn't have any readily available. Ideas?

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -24,6 +24,23 @@ Required IDs: using-http
 
 Link Defaults: html (dfn) queue a task/in parallel/reflect
 </pre>
+<pre class="biblio">
+{
+    "draft-ietf-oauth-sd-jwt-vc": {
+        "href": "https://datatracker.ietf.org/doc/html/draft-ietf-oauth-sd-jwt-vc",
+        "title": "SD-JWT-based Verifiable Credentials (SD-JWT VC)",
+        "status": "Internet-Draft",
+        "publisher": "IETF"
+    },
+    "vc-data-model-1.1": {
+        "href": "https://www.w3.org/TR/vc-data-model-1.1/",
+        "title": "Verifiable Credentials Data Model v1.1",
+        "status": "REC",
+        "publisher": "W3C",
+        "deliveredBy": ["https://www.w3.org/groups/wg/vc"]
+    }
+}
+</pre>
 <pre class="link-defaults">
 spec:css-cascade-5; type:dfn; text:inherit
 spec:css2; type: property; text: line-height
@@ -3487,6 +3504,52 @@ You may also find that none of these options is acceptable to your group.
 While the best path forward may be to choose not to specify the feature,
 there is the risk that some implementations
 may ship the feature as a nonstandard API.
+
+<h3 id="multiple-algorithms">Write only one algorithm to accomplish a task</h3>
+
+When specifying how to accomplish a task, write a single algorithm to do it,
+instead of letting implementers pick between multiple algorithms.
+It is very difficult to ensure that
+two different algorithms produce the same results in all cases,
+and doing so is rarely worth the cost.
+
+Multiple algorithms seem particularly tempting when defining
+how to parse file formats,
+but they are an equally bad idea in that context.
+
+<div class="example" id="html-polyglot">
+
+[[html-polyglot inline]] was an attempt to define a common subset of both XHTML and HTML
+that could be parsed into equivalent DOM trees
+using either the [[HTML#the-xhtml-syntax|XHTML parsing]]
+or [[HTML#syntax|HTML parsing]] algorithm.
+Authors who tried to use this syntax tended to produce documents
+that actually only worked with one of the two parsers.
+
+</div>
+
+<div class="example" id="json-ld-polyglot">
+
+The [[vc-data-model-1.1 inline obsolete]] and [[cid inline]] both endorse implementations
+that use either JSON or JSON-LD to parse bytes into their data models.
+Because JSON-LD provides many more ways
+to assign properties to particular objects than JSON does,
+these specifications had to add extra rules to both kinds of parsers
+in order to ensure that each input document had exactly one possible interpretation.
+[[vc-data-model-2.0 inline]] and [[draft-ietf-oauth-sd-jwt-vc inline]] fixed the problem
+by defining different media types for JSON-LD vs JSON,
+which means implementations are instructed to use exactly one of the algorithms.
+The authors of [[cid inline]] have decided the extra specification work and testing
+are acceptable prices to be able to allow both implementation strategies,
+and they have partially mitigated the risks by explicitly stating that
+if a single document can conformantly have multiple interpretations,
+it's a specification bug.
+
+</div>
+
+Note: While [[rfc6838#section-6|structured suffixes]] define that
+a document can be parsed in two different ways,
+they do not violate this rule because the results have different data models.
 
 <h3 id="monkey-patching">Avoid monkey patching</h3>
 A <dfn export>monkey patch</dfn> layers new functionality on top of an existing specification in a way that extends, overrides, or otherwise modifies the existing specification's behavior.


index.bs
```diff
@@ -3488,6 +3505,52 @@ While the best path forward may be to choose not to specify the feature,
 there is the risk that some implementations
 may ship the feature as a nonstandard API.
 
+<h3 id="multiple-algorithms">Write only one algorithm to accomplish a task</h3>
```
annevk, line 3508: Maybe "goal" instead of "task"? This immediately made me think of the event loop.

index.bs
```diff
@@ -3488,6 +3505,52 @@ While the best path forward may be to choose not to specify the feature,
...
+are acceptable prices to be able to allow both implementation strategies,
+and they have partially mitigated the risks by explicitly stating that
+if a single document can conformantly have multiple interpretations,
+it's a specification bug.
+
+</div>
+
+Note: While [[rfc6838#section-6|structured suffixes]] define that
+a document can be parsed in two different ways,
+they do not violate this rule because the results have different data models.
```
martinthomson, line 3552: Isn’t the real difference here that the suffix parsing produces an intermediate result?

i suspect that this is insufficient still, because it doesn’t really get at why suffix parsers exist. That is still somewhat contested, but my view is that intermediate results are rarely able to be processed meaningfully, so they are limited to use in diagnostic tools and the like.

msporny (commented):
Hmm, the principle seems too blunt to be useful. Some high level thoughts to start, I'm still trying to think about what text would be useful:

* Polyglot, as a term, is wrong -- this is about interpreting the same data serialization using different algorithms, not about a single system interpreting that serialization using different algorithms. It's more about an ecosystem interpreting that same data serialization using different algorithms.
* Yes, there are cases where this resulted in bad outcomes -- XHTML/HTML is a good example.
* The comparison between VCDM and SD-JWT-VC is totally wrong, they're two totally different data models, using two totally different serializations, using two totally different algorithms -- and there are a number of us that think that whole thing is a massive standardization failure, so using that as an example of the right way to do something is not what we want to do.
* The multiple suffixes thing is also contested -- we couldn't find broad-scale usage of suffix-based processing, what @martinthomson is saying is important here. I'll add that the suffix-based processing is also not a clear example of why this principle is good or bad.

Fundamentally, the principle seems misguided. Yes, at some level one data format and one algorithm is a good thing. However, what a traditional web crawler gets out of a web page is different from what a browser parsing the web page works with is different from what a frontier AI model gets out of a web page. The algorithms that each uses are quite different and *useful* and this principle seems to be arguing against that.

I think the only solid ground here is the XHTML/HTML example. You're going to get push back on the other items being mentioned if they continue to be mentioned in the way the current PR is written up.

I'll try to think of some constructive text, but wanted to get some preliminary thoughts down in an effort to help shape the PR into something more easily defensible.

filip26:
**Algorithms + Data Structures = Programs (Niklaus Wirth).** 

It’s rational to avoid having two algorithms performing the same function, especially when considering costs like time and memory complexity, and to recommend the one that best fits the criteria. However, if this change is based on the assumption:

> use either JSON or JSON-LD to parse bytes into their data models.

then there is a misunderstanding of the algorithmization basics. JSON and JSON-LD have different data models, as noted. They involve different data structures in the equation at the top, which means different algorithms are needed because they operate on different data.

From this perspective, calling for one algorithm to operate on different data models does not make sense.

My recommendation would be to use a different argument when advocating for a single algorithm - considering factors such as time complexity, space complexity, etc.

martinthomson:
I don't agree with Manu about this being misguided.  The point here is that the same HTML document is not seeking to express multiple distinct sets of semantics depending on how it is processed, there is just one HTML with one interpretation, and one data model that both the producer of the content and the consumer of content can agree on.  If they disagree, that is likely due to one or other being wrong.

This is because there is just a single specification for HTML and a single way to interpret HTML content according to that content.

Obviously, what someone does once they have received HTML might differ, but those differences do not relate to how the HTML itself was interpreted, but how the content at the next layer (that is, the words and images and stuff like that) is interpreted.  Sure, a human and an AI model will seek to do different things with the information they are presented with, but the interpretation is singular.

Where CID struggles a little is that there are two paths to the same interpretation.  It manages that by giving implementations a choice and *promising* that the outcome will be the same either way.  It's bad, because now there is a third place where a bug can result in a different interpretation (producer, consumer, and now spec), but it's not fundamentally a polyglot in the sense that there are multiple *divergent* interpretations possible.

The core message is that having divergent paths is undesirable.  And yes, that means saying that seeking to have a pure JSON vs a JSON-LD interpretation of the same content is a bad idea.  Because divergence in data models means that there is no single interpretation of the content upon which all potential recipients might agree upon.

gkellogg:
This issue really gets at the heart of a basic divide at W3C: one that is browser-centric, vs. one which is data-centric. In fact, JSON-LD does parse JSON (and YAML and CBOR) into a common INFRA-based data structure (called the [Internal Representation](https://www.w3.org/TR/json-ld11-api/#dfn-internal-representation)) which various algorithms operate over to perform different transformations, including to interpret as RDF. This is the core reason behind JSON-LD, which has become extremely widely used on the Web (in large part, due to schema.org).

HTML is also often processed differently, typically by interpreting the resulting DOM. This might be done to extract Microdata/RDFa, interpret the contents of `script` elements, or to perform extensive re-formatting through ReSpec or Bikeshed. Search engines interpret the DOM for their own uses, so a general principle would seem to settle on a data representation which different applications can use to suit their different use cases.

In the case of Verifiable Credentials, the basic failure would seem to be a lack of agreement on how to work with the data that is represented in the JSON. This is an area the TAG can help with for future specs, rather than getting into a reductionist view that Polyglot formats are fundamentally a bad idea.

index.bs
```diff
@@ -3488,6 +3505,52 @@ While the best path forward may be to choose not to specify the feature,
...
+<div class="example" id="json-ld-polyglot">
+
+The [[vc-data-model-1.1 inline obsolete]] and [[cid inline]] both endorse implementations
+that use either JSON or JSON-LD to parse bytes into their data models.
+Because JSON-LD provides many more ways
+to assign properties to particular objects than JSON does,
+these specifications had to add extra rules to both kinds of parsers
+in order to ensure that each input document had exactly one possible interpretation.
+[[vc-data-model-2.0 inline]] and [[draft-ietf-oauth-sd-jwt-vc inline]] fixed the problem
+by defining different media types for JSON-LD vs JSON,
```
BigBlueHat, line 3540: This statement is incorrect. The VCDMv2 only has a single media type: [`application/vc`](https://www.w3.org/TR/vc-data-model-2.0/#media-types).

Likewise, SD-JWT-VC only has one: `application/dc+sd-jwt`. However, SD-JWT is not JSON parse-able base format.

Both specifications have parsing algorithms unique to their media types--and specific to their tasks.

It remains unclear how these examples are "polyglot".

hober (approved):
This looks really good to me on a first-pass look. I'll try to find the time to give it a closer read some other time, but please don't block on waiting for me to do so. :)

pchampin:
> It's bad, because now there is a third place where a bug can result in a different interpretation (producer, consumer, and now spec).

I disagree: bugs in specs are always possible, regardless of the fact that the spec acknowledges it or not.

Also note that an algorithm is not an implementation, and no algorithm is entirely neutral: Javascript developers do not write algorithms the same way Rust (or Java, or Scala...) developers do. As a consequence, every implementer of a spec has to adapt the algorithms. The differences are not limited to programming languages: whether you are using a relational, key-value, document, or graph database, you will encode and handle a given data model differently. Ultimately, every implementation defines a different interpretation path, whether we like it or not.

In some ecosystems (browser APIs come to mind), this heterogeneity may be limited, and therefore the "only one algorithm" principle is probably a good enough way to ensure interoperability. In other ecosystems, where the heterogeneity is higher, it is probably better to acknowledge it and provide guidance to the different kinds of implementers. This is the strategy taken by the editors of CID and VCDM, and should not IMO be flagged as bad practice™.

index.bs
```diff
@@ -3488,6 +3505,49 @@ While the best path forward may be to choose not to specify the feature,
...
 may ship the feature as a nonstandard API.
 
+<h3 id="multiple-algorithms">Write only one algorithm to accomplish a goal</h3>
+
+When specifying how to accomplish a goal, write a single algorithm to do it,
+instead of letting implementers pick between multiple algorithms.
+
+It is very difficult to ensure that
+two different algorithms produce the same results in all cases,
+and doing so is rarely worth the cost.
```
csarven, line 3515:
```suggestion
<h3 id="single-algorithms">Write only one algorithm to accomplish a goal</h3>

When defining how to achieve a feature, it's better to specify a single
approach rather than offering multiple options.

If multiple methods are allowed, they must be equivalent in conformance to
avoid unnecessary complexity, inconsistency, and security risks.
```

csarven (changes requested):
tl;dr: The concept of "polyglot format" lacks a clear definition. It doesn't serve as a helpful lens for this discussion. The current examples don't adequately support the argument. Either different and better examples are needed to distil a principle, or a different approach should be considered. So, I've opted for the latter and made a change suggestion that aims to provide a more agreeable principle.

---

An authoritative definition of "polyglot format" with explicit criteria would help clarify whether a format, profile, or data model qualifies as such. If such a definition exists and aligns with [open standards principles](https://open-stand.org/about-us/principles/), a reference would be helpful.

---

SVG and MathML can be written to be valid and processed in both standalone XML and inline HTML contexts.

Since `draft-ietf-oauth-sd-jwt-vc` and `vc-data-model-1.1` are separate specifications, the example does not fully illustrate the intended point.

Other examples:

* HTTP status codes: Servers use either `403` or `404` to prevent resource discovery.
* Click event handling: `addEventListener('click')` and `onclick` both achieve the same goal but through different mechanisms.

I don't see the significance of noting structured suffixes any more than the relationship between the 'text' top-level type and its subtypes. `text/html` is specific, and interpreting it as plain text is merely a useful step in the process of parsing it as intended (as pointed out by @martinthomson [here](https://github.com/w3ctag/design-principles/pull/562#discussion_r1984694552), but not an *alternative* interpretation path. The same applies to `application/ld+json`, with the goal to interpret it as JSON-LD, not merely/only JSON. Treating JSON-LD purely as JSON is analogous to treating CSV or HTML as plain text - although a useful step in the whole process, as pointed by @gkellogg [here](https://github.com/w3ctag/design-principles/pull/562#issuecomment-2711747542), it will ignore part of the intended structure, semantics, and functionality. Similarly, `application/json` is intended to represent JSON, not the raw series of strings that make it up.

CID's [context injection](https://www.w3.org/TR/cid/#context-injection) states "[a]ny differences in semantics between documents processed in either mode are either implementation or specification bugs".

As I understand it, this is analogous to different representations of an HTTP resource being deemed **equivalent** in meaning, e.g., when `/dog` depicts a dog with content negotiated as `image/jpeg`, it should also depict a dog, not a cat, when content negotiated as `image/png`. Stating that a resource can have multiple equivalent representations is not deemed to be a bug in neither HTTP or Web Architecture, and that obviously implies different algorithms to make sense of these different representations.

Sometimes reality is a bit more nuanced than "good" or "bad" =)

So, if a principle is to be stated beyond the obvious, it should encourage simplicity, clarity, and security, while accounting for the complexity and interoperability of different implementations.

TallTed:
> This explains why and when "polyglot" formats are a bad idea.

It doesn't appear to do anything of the sort. I don't see a clear definition of what such a _"polyglot" format_ is, nor how one might be worked with, nor how working with one might go bad (nor go well, so there's that).

I do see a number of apparent misunderstandings by the author which others have pointed out directly, so I won't go into those myself.

> Principle: Write only one algorithm to accomplish a task.

This doesn't appear to be the focus of this writing. Perhaps it shouldn't be the title of the PR, either?

martinthomson:
The term "polyglot" only appears as an anchor, not in text, so I'm not sure that the criticism of that term is relevant.

My view here is somewhat more nuanced, yet I reach the same conclusions as are laid out in the changeset.

Let's take JSON-LD vs. JSON, since I think that is at the heart of the concerns here.  You might consider a processing pipeline that goes something like this:

1. Text/Unicode decoding (from UTF-8)
2. JSON decode into a generic JSON-compatible information model
3. Take JSON-LD attributes and produce a JSON-LD-aware information model
4. Maybe some generic processing based on an understanding of the JSON-LD-relevant information
5. Application-specific processing

There's an alternative approach.  I've heard many people say that JSON-LD is just JSON if you don't understand JSON-LD.  So then you have this instead:

1. Text/Unicode decoding (from UTF-8)
2. JSON decode into a generic JSON-compatible information model
3. Application-specific processing

That's two paths to the same outcome.  You therefore need to be careful to ensure that the three final steps of the JSON-LD-aware path are identical to the application-specific JSON processing in the latter path.  That is, the inferences about semantics that an application that ignores JSON-LD decorations need to be identical to the same inferences that are made when the semantics are derived from an understanding of the decorations.  

What if the JSON-LD attributes indicated a different interpretation?  For instance, what if `@context` differs?  A JSON-LD pipeline might follow the interpretation that is indicated by processing a schema, producing a different outcome.  (The CID specification would seem to avoid this by specifying a SHA-256 hash of the only `@context` permissible in that case, but that would seem to lose much of the benefit that might come from JSON-LD processing as a result.)

That puts a lot of weight on the details of the application-specific processing in each of the alternatives.  Get it wrong and the same document will produce two different outcomes.  To use the CID spec further as an example, what happens if `@context` is something other than "https://www.w3.org/ns/cid/v1"?  (I couldn't work out the answer to this question; I'd infer that CID is incompletely specified, but there are conventions in the spec that I'm not familiar with.)

That's all that this proposed principle intends to say: that two paths to the same goal is harder to get right.

---

The example of SVG/MathML as HTML are not relevant as there is just one way to interpret those, even if (perhaps) the SVG or MathML is first interpreted as a bunch of generic HTML elements on its way to SVG parsing.  There is no other path defined to getting a valid SVG, so the outcome is never ambiguous.

`onclick` handlers are perhaps a better example of where divergence is justified.  The difference between declarative and imperative APIs on the platform that have a defined equivalency do introduce two ways to reach the same end state, but you will observe that they follow a singular process for most of their processing.  The necessary points of divergence are parallel in that there are clearly-defined alternatives with clear demarcation points for where they diverge and converge.  That is, the entry point might differ, but the result of processing in either case is defined by the same processing model, with the same inputs.

Consider also that JS strings can be delineated with single- or double-quotes [^1].  This is complexity that aids ease of use for JS authors.  That comes with a non-zero cost in complexity, but it is justified on the basis that it makes it easier to author JS, and there are a lot more JS authors than there are the specifications and implementations that bear the complexity cost.

Both of those examples (onclick, JS strings) are very minor divergences.  Most platforms grow alternative paths to the same outcome over time, as new APIs are added that supersede the old APIs, but old APIs stick around.  XmlHTTPRequest and Fetch are even better examples of non-trivial divergences.

I don't think that the content negotiation example is relevant either.  That `/dog` is a dog in one format and a cat in another could be a deliberate property of the resource.  The point of content negotiation is to allow resources to present different aspects based on the needs of clients.  That might lead to odd outcomes, but that's just how that is.  Some people continue to use examples like this as cause to consider conneg as a misfeature of HTTP; I'm partly convinced by those arguments, but it has seen a lot of application as a way to manage version migration, so I tend toward accepting the good along with the bad.

[^1]: and backticks, which have even more different rules

pchampin:
> That's all that this proposed principle intends to say: that two paths to the same goal is harder to get right.

I would agree with that, but in its current state, the PR is actually much less nuanced: "write a single algorithm" (imperative form), "ensur[ing] that two different algorithms produce the same results [...] is rarely worth the cost".

@csarven suggestion [above](https://github.com/w3ctag/design-principles/pull/562#discussion_r1998597693) seems more balanced.

Issue #514, opened 2024-11-20
New principle: only extend problematic and non-consensus features if the extension acts to fix the problems


jyasskin:
This was brought up in our discussion of [Protected Audiences Bidding and Auction Services API](https://github.com/w3ctag/design-reviews/issues/1009#issuecomment-2489696097).

This might fit into https://w3ctag.github.io/design-principles/#leave-the-web-better.

The idea is that improving a problematic feature will tend to increase use of that feature, which increases any problems caused by use of the feature, so the improvements should compensate for the increase in use with a decrease in the problems-per-use.

Extensions shouldn't have to completely fix the problems in order to be a good idea, just have some indication that the total amount of problems will go down.

martinthomson:
So the concern from our discussion was that an extension to something that has some sort of bad effect might make it more attractive to use the thing and thus have the bad effect occur more often.  The suggestion was that maybe someone building an extension has an obligation to make the situation better (perhaps by working to mitigate the problem) in addition to building their extension.  There isn't an expectation that someone completely solve a problem (that might be impossible), but incremental improvements might be OK.  Circumstances will differ and we'll have to make a judgment call.

Protected Audience is not obviously a good example of this because the proponents of Protected Audience do not always agree that these are negative effects.  A better example might be enhancements to `window.alert()`, which is acknowledged to be bad as a stop-the-world API.  Those improvements wouldn't be acceptable without doing something to make alerting less bad.

PR #563, opened 2025-03-17
Draft text for issue #514

lolaodelola:
This is the [discussed draft text](https://github.com/w3ctag/design-principles/issues/514) for issue #514.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -322,21 +322,12 @@ they are also defined by how content uses those features.
 
 <h3 id="leave-the-web-better">Leave the web better than you found it</h3>
 
-As you add new capabilities to the web platform,
-do so in a way that improves the overall platform,
-for example its security, privacy or accessibility characteristics.
+As you add new capabilities to the web platform, do so in a way that improves the overall platform, for example its security, privacy or accessibility characteristics.
+The existence of a defect in one part of the platform must not be used to excuse an addition or extension to the defect, which would further decrease the overall platform quality. Consistency is not a good reason to propagate problems. Where possible, build new web capabilities that improve the overall platform quality by mitigating existing defects. Do not degrade existing capabilities without good reason.
 
-The existence of a defect in one part of the platform
-must not be used to excuse an addition or extension to the defect,
-which would further decrease the overall platform quality.
-Where possible, build new web capabilities that improve the overall platform quality
-by mitigating existing defects.
-Do not degrade existing capabilities without good reason.
+Do not extend a platform feature that has known defects, without considering how to address those defects. Extensions could create new ways in which the defect could cause problems, such as by encouraging more use of the defective feature. It is not necessary to completely fix all aspects of a defect, it is more important to consider mitigations or fixes when developing such an extension.
 
-Parts of the web platform evolve independently.
-Issues that are present with a certain web technology now may be fixed in a subsequent iteration.
-Duplicating these issues makes fixing them more difficult.
-By adhering to this principle we can make sure overall platform quality improves over time.
+Parts of the web platform evolve independently. Issues that are present with a certain web technology now may be fixed in a subsequent iteration. Duplicating these issues makes fixing them more difficult. By adhering to this principle we can make sure overall platform quality improves over time.
 
 <h3 id="data-minimization">Minimize user data</h3>
 


index.bs
```diff
@@ -322,21 +322,12 @@ they are also defined by how content uses those features.
...
+The existence of a defect in one part of the platform must not be used to excuse an addition or extension to the defect, which would further decrease the overall platform quality. Consistency is not a good reason to propagate problems. Where possible, build new web capabilities that improve the overall platform quality by mitigating existing defects. Do not degrade existing capabilities without good reason.
+
+Do not extend a platform feature that has known defects, without considering how to address those defects. Extensions could create new ways in which the defect could cause problems, such as by encouraging more use of the defective feature. It is not necessary to completely fix all aspects of a defect, it is more important to consider mitigations or fixes when developing such an extension.
+
+Parts of the web platform evolve independently. Issues that are present with a certain web technology now may be fixed in a subsequent iteration. Duplicating these issues makes fixing them more difficult. By adhering to this principle we can make sure overall platform quality improves over time.
```
jyasskin, line 330: This document tends to [use semantic line breaks](https://rhodesmill.org/brandon/2012/one-sentence-per-line/). I'm inclined to say that all patches should do that, but even if we decide not to, they should definitely wrap at 80-100 columns.

Also please avoid re-wrapping lines you're not otherwise touching, to make the diff easier to read.

index.bs
```diff
@@ -322,21 +322,12 @@ they are also defined by how content uses those features.
...
-Where possible, build new web capabilities that improve the overall platform quality
-by mitigating existing defects.
-Do not degrade existing capabilities without good reason.
-
-Parts of the web platform evolve independently.
-Issues that are present with a certain web technology now may be fixed in a subsequent iteration.
-Duplicating these issues makes fixing them more difficult.
-By adhering to this principle we can make sure overall platform quality improves over time.
+As you add new capabilities to the web platform, do so in a way that improves the overall platform, for example its security, privacy or accessibility characteristics.
+The existence of a defect in one part of the platform must not be used to excuse an addition or extension to the defect, which would further decrease the overall platform quality. Consistency is not a good reason to propagate problems. Where possible, build new web capabilities that improve the overall platform quality by mitigating existing defects. Do not degrade existing capabilities without good reason.
```
jyasskin, line 326: I think the only new thing in this paragraph is the sentence about consistency. I think we need to be more nuanced about that, since https://w3ctag.github.io/design-principles/#consistency says consistency _is_ a good reason to propagate certain problems. I'm not sure exactly what to suggest, but I'll keep thinking about it.

index.bs
```diff
@@ -322,21 +322,12 @@ they are also defined by how content uses those features.
...
+
+Do not extend a platform feature that has known defects, without considering how to address those defects. Extensions could create new ways in which the defect could cause problems, such as by encouraging more use of the defective feature. It is not necessary to completely fix all aspects of a defect, it is more important to consider mitigations or fixes when developing such an extension.
```
jyasskin, line 328: We should probably say something stronger than "consider". Maybe use "should" in a more direct sentence about partially fixing the defects, since that [allows exceptions when justified](https://datatracker.ietf.org/doc/html/rfc2119#section-3)?

jyasskin (commented):
Thanks for the PR! I have several stylistic requests here, which is to be expected in your first several PRs.

Please write PR titles (and commit messages) to directly explain what's in the PR, since issue references in titles generally don't get linked to the issue, and reading a summary is easier than following a link. The issue link belongs in the body of the PR. I've updated this one, but feel free to change it again if you don't think it's an accurate summary of the PR.

index.bs
lolaodelola: It looks like my IDE hard wrap setting didn't take effect. Thanks for catching this! I'll add the line breaks and push it back up.

index.bs
```diff
@@ -322,21 +322,23 @@ they are also defined by how content uses those features.
...
-Parts of the web platform evolve independently.
-Issues that are present with a certain web technology now may be fixed in a subsequent iteration.
-Duplicating these issues makes fixing them more difficult.
-By adhering to this principle we can make sure overall platform quality improves over time.
+As you add new capabilities to the web platform, do so in a way that improves the overall platform,
+for example its security, privacy or accessibility characteristics. The existence of a defect in one
+part of the platform must not be used to excuse an addition or extension to the defect, which would
+further decrease the overall platform quality. Consistency is not a good reason to propagate
+problems. Where possible, build new web capabilities that improve the overall platform quality by
+mitigating existing defects. Do not degrade existing capabilities without good reason.
```
martinthomson, line 330: The amount of added text here causes this to lose its punchiness.  I would prefer to see something sharper.

(Also, the new wrapping here makes this hard to review.  Not a lot of text has changed, but this PR looks like a bigger change due to that wrapping.)

There is a high-level point (the first sentence):

> As you add new capabilities to the web platform,
> do so in a way that improves the overall platform.

Then there is the part about excuses, which in the original is pretty good as a second paragraph.  I would leave that alone, though splitting it in two to talk about consistency as a single paragraph (with a one-line insertion of "Consistency is not a good reason to propagate problems") and the new text about fixing defects in the latter.

In particular, the "where possible, build new capabilities [...] mitigating existing defects" is the sentiment of the newly added paragraph.  The point about degrading is probably not needed and could be dropped.

index.bs
```diff
@@ -322,21 +322,23 @@ they are also defined by how content uses those features.
...
+
+Do not extend a platform feature that has known defects, without considering how to address those
+defects. Extensions could create new ways in which the defect could cause problems, such as by
+encouraging more use of the defective feature.
+It is not necessary to completely fix all aspects of a defect, it is more important to consider
+mitigations or fixes when developing such an extension.
```
martinthomson, line 336:
```suggestion
Do not extend a platform feature that has known defects
without considering how to address those defects.
It is not necessary to completely fix existing problems,
but some attempt to address flaws is expected
when extending a feature with known issues.
This way, new or increased usage does not result in
increasing the harm that the defect might cause.
```

index.bs
```diff
@@ -322,21 +322,23 @@ they are also defined by how content uses those features.
...
+
+Parts of the web platform evolve independently. Issues that are present with a certain web
+technology now may be fixed in a subsequent iteration. Duplicating these issues makes fixing them
+more difficult. By adhering to this principle we can make sure overall platform quality improves
+over time.
```
martinthomson, line 341: I don't understand why the "independence" point was added.  I read it as weakening the message, rather than strengthening it.

martinthomson (changes requested):

index.bs
lolaodelola: I copied & pasted the text from the minutes taken in Paris so don't remember who suggested it specifically. Maybe it'd be helpful to bring to the group on Slack/breakout?

index.bs
martinthomson: I think that you missed the intent of the comment.

When using semantic breaks, you attempt to keep one concept per line.  That means that lines end on periods and semi-colons always, and at the end of obvious clauses.  (This is not a good idea with GitHub markdown, which retains line breaks, BTW.)

That means not doing a hard wrap at any particular column (like your 100ch), but choosing where to break deliberately.

index.bs
```diff
@@ -323,15 +323,23 @@ they are also defined by how content uses those features.
...
+do so in a way that improves the overall platform.
 
 The existence of a defect in one part of the platform
 must not be used to excuse an addition or extension to the defect,
 which would further decrease the overall platform quality.
-Where possible, build new web capabilities that improve the overall platform quality
-by mitigating existing defects.
-Do not degrade existing capabilities without good reason.
+
+Consistency is not a good reason to propagate problems.
```
jyasskin, line 332: We have to explain how to satisfy both this and https://www.w3.org/TR/design-principles/#consistency, which says consistency _is_ a good reason to propagate certain problems.

index.bs
```diff
@@ -323,15 +323,23 @@ they are also defined by how content uses those features.
...
+Where possible, build new web capabilities that improve the overall platform
+quality by mitigating existing defects.
```
jyasskin, line 334: These are the same as the pre-change lines, so make sure they're also wrapped the same:

```suggestion
Where possible, build new web capabilities that improve the overall platform quality
by mitigating existing defects.
```

index.bs
martinthomson: Is this a case where we can simply forward-reference?  (I think that the text on consistency could be improved, mostly by referring to the priority of constituencies, but that's orthogonal.)

I would maybe suggestion:

```suggestion
Consistency is not often a good reason to propagate problems; see [[#consistency]].
```

jyasskin: I think forward-referencing helps here, but folks are still going to be confused when the referenced section says things like "In some cases it makes sense to break consistency to improve usability, but the improvement should be very significant to justify this." I'm inclined to just undo this sentence, and let the new paragraph carry the meaning. Generally, the kinds of fixes it suggests don't actually break consistency.

martinthomson: WFM

```suggestion
```

martinthomson (approved):
@jyasskin we've mostly resolved things here.  Want to choose how to resolve the issues  and merge? (@marcoscaceres has an idea for our consistency question that might come in the next few hours.)

martinthomson:
Thanks for shepherding this @lolaodelola.  It was not a trivial one in the end.

PR #570, opened 2025-04-26
Don't require action within a Promise reaction microtask.

### LINKED ISSUE #342: New principle: Support common promise practices

jan-ivar:
Fixes #342.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/design-principles/pull/570.html" title="Last updated on May 19, 2025, 5:31 AM UTC (9935a7f)">Preview</a> (<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/design-principles/pull/570.html#sync-callbacks" title="#sync-callbacks">#sync-callbacks</a>) (<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/design-principles/pull/570.html#promises-and-reaction" title="#promises-and-reaction">#promises-and-reacti…</a>) | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/570/e10b073...jan-ivar:9935a7f.html" title="Last updated on May 19, 2025, 5:31 AM UTC (9935a7f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -324,15 +324,25 @@ they are also defined by how content uses those features.
 <h3 id="leave-the-web-better">Leave the web better than you found it</h3>
 
 As you add new capabilities to the web platform,
-do so in a way that improves the overall platform,
-for example its security, privacy or accessibility characteristics.
+do so in a way that improves the overall platform.
 
 The existence of a defect in one part of the platform
 must not be used to excuse an addition or extension to the defect,
 which would further decrease the overall platform quality.
+
+Consistency with existing problematic design patterns
+is not a good reason to propagate those patterns
+in newly specified Web APIs or platform features.
 Where possible, build new web capabilities that improve the overall platform quality
 by mitigating existing defects.
-Do not degrade existing capabilities without good reason.
+
+Do not extend a platform feature that has known defects,
+without considering how to address those defects.
+It is not necessary to completely fix existing problems,
+but some attempt to address flaws is expected
+when extending a feature with known issues.
+This way, new or increased usage does not result in
+increasing the harm that the defect might cause.
 
 Parts of the web platform evolve independently.
 Issues that are present with a certain web technology now may be fixed in a subsequent iteration.


index.bs
```diff
@@ -1965,6 +1965,38 @@ See also:
...
 
+<h4 id="sync-callbacks">When to use callback functions instead</h4>
+
+<p>If an algorithm requires that callback functions respond
+synchronously (e.g. to take or veto some action), then that is a
+synchronous API, and should not be modelled with a promise. Instead,
+use an explicit callback or event.</p>
+
+<p>In Promise reaction callbacks, you may not require action within
+the same microtask.</p>
```
jyasskin, line 1976: I think it might be
```suggestion
a fixed number of microtasks.</p>
```

because promise reactions always run with a delay of 1 microtask from when the promise was fulfilled or rejected.

index.bs
```diff
@@ -1965,6 +1965,38 @@ See also:
...
+
+<p>This protects promise composition: patterns like <code>await</code>,
+wrapper helpers, logging, <code>Promise.race()</code>, or multiple
+<code>.then()</code> branches all queue an extra microtask and must
+not cause action at a distance.</p>
```
jyasskin, line 1981: I'm not sure this would be "action at a distance". Maybe

```suggestion
not significantly change the behavior of the code they wrap.</p>
```

index.bs
```diff
@@ -1965,6 +1965,38 @@ See also:
...
+<p>If an algorithm requires that callback functions respond
+synchronously (e.g. to take or veto some action), then that is a
+synchronous API, and should not be modelled with a promise. Instead,
+use an explicit callback or event.</p>
+
+<p>In Promise reaction callbacks, you may not require action within
+the same microtask.</p>
+
+<p>This protects promise composition: patterns like <code>await</code>,
+wrapper helpers, logging, <code>Promise.race()</code>, or multiple
```
jyasskin, line 1979: Maybe

```suggestion
wrapping helper functions, logging, <code>Promise.race()</code>, or multiple
```

index.bs
```diff
@@ -1965,6 +1965,38 @@ See also:
...
 * <a href="https://www.w3.org/2001/tag/doc/promises-guide">Writing Promise-Using Specifications</a>
 
+<h4 id="sync-callbacks">When to use callback functions instead</h4>
+
+<p>If an algorithm requires that callback functions respond
+synchronously (e.g. to take or veto some action), then that is a
+synchronous API, and should not be modelled with a promise. Instead,
+use an explicit callback or event.</p>
+
+<p>In Promise reaction callbacks, you may not require action within
```
jyasskin, line 1975: It'd be nice to link this so people can get a firm definition of what it is. Perhaps:

```suggestion
<p>In [Promise reaction callbacks](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-promisereaction-records),
you may not require action within
```

index.bs
```diff
@@ -1965,6 +1965,38 @@ See also:
...
+
+<p>This protects promise composition: patterns like <code>await</code>,
+wrapper helpers, logging, <code>Promise.race()</code>, or multiple
+<code>.then()</code> branches all queue an extra microtask and must
+not cause action at a distance.</p>
+
+<div class=example>
+Calling <code>event.preventDefault()</code> in an event handler,
+or choosing a response object inside a
+<code>fetchEvent.respondWith()</code> callback.
```
jyasskin, line 1986: This should be a sentence. How about:

```suggestion
If developer-supplied code needs to be able to call <code>event.{{Event/preventDefault()}}</code>
or choose a response object with
<code>fetchEvent.{{FetchEvent/respondWith()}}</code>,
it needs to be a callback rather than a reaction passed to <code>Promise.{{Promise/then()}}</code>.
```

index.bs
```diff
@@ -1965,6 +1965,38 @@ See also:
...
+</div>
+<div class="example">
+A non-event example is the [captureController.setFocusBehavior()](https://w3c.github.io/mediacapture-screen-share/#dom-capturecontroller-setfocusbehavior)
```
jyasskin, line 1989:
```suggestion
A non-event example is the <code>captureController.{{CaptureController/setFocusBehavior()}}</code>
```

jyasskin (commented):
Thanks @jan-ivar. I've left a bunch of suggestions for @martinthomson to review, and you're welcome to apply any that you agree with or make other changes in response, but we can also take it from here.

index.bs
```diff
@@ -1966,6 +1966,40 @@ See also:
 
 * <a href="https://www.w3.org/2001/tag/doc/promises-guide">Writing Promise-Using Specifications</a>
 
+<h4 id="sync-callbacks">When to use callback functions instead</h4>
+
+<p>If an algorithm requires that callback functions respond
+synchronously (e.g. to take or veto some action), then that is a
+synchronous API, and should not be modelled with a promise. Instead,
```
martinthomson, line 1973:
```suggestion
synchronous API, and should not be modeled with a Promise. Instead,
```

index.bs
```diff
@@ -1966,6 +1966,40 @@ See also:
...
+use an explicit callback or event.</p>
+
+<p>In [Promise reaction callbacks](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-promisereaction-records),
+you may not require action within
+a fixed number of microtasks.</p>
+
+<p>This protects promise composition: patterns like <code>await</code>,
```
martinthomson, line 1980:
```suggestion
<p>This protects Promise composition: patterns like <code>await</code>,
```

index.bs
```diff
@@ -1966,6 +1966,40 @@ See also:
...
+wrapping helper functions, logging, <code>Promise.race()</code>, or multiple
+<code>.then()</code> branches all queue an extra microtask and must
+not significantly change the behavior of the code they wrap.</p>
+
+<div class=example>
+If developer-supplied code needs to be able to call <code>event.{{Event/preventDefault()}}</code>
+or choose a response object with
+<code>fetchEvent.{{FetchEvent/respondWith()}}</code>,
+it needs to be a callback rather than a reaction passed to <code>Promise.{{Promise/then()}}</code>.
```
martinthomson, line 1989: Is this a good example or a bad one?

index.bs
```diff
@@ -1966,6 +1966,40 @@ See also:
...
+</div>
+<div class="example">
+A non-event example is the <code>captureController.{{CaptureController/setFocusBehavior()}}</code>
+method to push a window the user has chosen to screen-capture, to
+the front. Applications can decide this based on the window chosen.
```
martinthomson, line 1994:
```suggestion
A non-event example is the <code>captureController.{{CaptureController/setFocusBehavior()}}</code>
method, which pushes a window the user has chosen to screen-capture to
the front. Applications can decide this based on the window chosen.
```

index.bs
```diff
@@ -1966,6 +1966,40 @@ See also:
...
+But any delay risks click-jacking.
```
martinthomson, line 1995:
```suggestion
The challenge is that allowing delay creates an opportunity for click-jacking attacks from sites.
```

A link defining click-jacking would be helpful.

index.bs
```diff
@@ -1966,6 +1966,40 @@ See also:
...
+
+Instead of requiring the method be called synchronously on the
+reaction microtask of the promise from `getDisplayMedia()`, the
```
martinthomson, line 1998:
```suggestion
reaction microtask of the promise from `getDisplayMedia()`, the
```
A link to `getDisplayMedia()` would also be helpful.

index.bs
```diff
@@ -1966,6 +1966,40 @@ See also:
 
 * <a href="https://www.w3.org/2001/tag/doc/promises-guide">Writing Promise-Using Specifications</a>
 
+<h4 id="sync-callbacks">When to use callback functions instead</h4>
+
+<p>If an algorithm requires that callback functions respond
+synchronously (e.g. to take or veto some action), then that is a
```
martinthomson, line 1972:
```suggestion
synchronously (e.g., to take or veto some action), then that is a
```

I'm not sure about our conventions here; this is the one that I've been taught to use. 

... checks ...

damn.  Well.

index.bs
```diff
@@ -1966,6 +1966,40 @@ See also:
 
 * <a href="https://www.w3.org/2001/tag/doc/promises-guide">Writing Promise-Using Specifications</a>
 
+<h4 id="sync-callbacks">When to use callback functions instead</h4>
+
+<p>If an algorithm requires that callback functions respond
+synchronously (e.g., to take or veto some action), then that is a
+synchronous API, and should not be modeled with a Promise. Instead,
+use an explicit callback or event.</p>
```
martinthomson, line 1974:
```suggestion
Use synchronous callbacks only when an immediate response is needed.

Synchronous actions are sometimes needed in APIs,
particularly where there are latency-sensitive actions involved.
For example, to cancel the default action a browser takes in response to an event
or to provide input on an active synchronous process like media processing.
Synchronous actions cannot be modeled with a Promise.
Instead, use an explicit callback or event.
```

index.bs
```diff
@@ -1966,6 +1966,40 @@ See also:
 
 * <a href="https://www.w3.org/2001/tag/doc/promises-guide">Writing Promise-Using Specifications</a>
 
+<h4 id="sync-callbacks">When to use callback functions instead</h4>
```
martinthomson, line 1969:
```suggestion
<h4 id="sync-callbacks" class="no-num no-toc">When to use callback functions</h4>
```

index.bs
```diff
@@ -1966,6 +1966,40 @@ See also:
...
+<p>In [Promise reaction callbacks](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-promisereaction-records),
+you may not require action within
+a fixed number of microtasks.</p>
+
+<p>This protects Promise composition: patterns like <code>await</code>,
+wrapping helper functions, logging, <code>Promise.race()</code>, or multiple
+<code>.then()</code> branches all queue an extra microtask and must
+not significantly change the behavior of the code they wrap.</p>
+
+<div class=example>
```
martinthomson, line 1985:
```suggestion
<div class=example>
Event handlers are invoked and cancelled synchronously.
```

index.bs
martinthomson: https://html.spec.whatwg.org/#writing-secure-applications-with-html is a section reference with a definition, but the spec doesn't have an anchor on the definition itself.

index.bs
jan-ivar: It's probably ingrained at this point. Whether it's a good pattern or not might be debatable. The last place I personally encountered it was https://w3c.github.io/webrtc-extensions/#dfn-icecandidatepairremove (though our motivation there was finding a manner of interfering with the ICE agent that the most people could stomach).

jan-ivar:
Thanks for all the feedback, this is looking much better!

index.bs
```diff
@@ -1966,6 +1966,49 @@ See also:
...
 * <a href="https://www.w3.org/2001/tag/doc/promises-guide">Writing Promise-Using Specifications</a>
 
+<h4 id="sync-callbacks" class="no-num no-toc">When to use callback functions</h4>
+
+Use synchronous callbacks only when an immediate response is needed.
+
+Synchronous actions are sometimes needed in APIs,
+particularly where there are latency-sensitive actions involved.
+For example, to cancel the default action a browser takes in response to an event
+or to provide input on an active synchronous process like media processing.
```
jyasskin, line 1976: This isn't quite a sentence. Perhaps:

```suggestion
For example, browsers need to know how to proceed immediately after an event finishes bubbling,
and media processing can't always wait for the next task boundary.
```

jyasskin (approved):

PR #584, opened 2025-09-11
Replace asking for users for consent with designing for user intent

mharbach:
After an [initial discussion](https://github.com/w3c/webappsec/blob/main/meetings/2025/2025-07-16-minutes.md#discouraging-permission-prompts) around [existing Chromium guidance to discourage the use of permission prompts](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/docs/security/no-prompts-please.md) in WebAppSec WG, I was encouraged to propose a change to TAG's design principles doc. This is that change.

My intention is to provide guidance around designing features and APIs in ways that limit the need for asking users questions. When there is a strong and justified need to ask such questions, I want to offer some principles on what needs to be true for users to be able to make good decisions.

H/t to @heisenburger, @mikewest, and @jyasskin for their input on this initial draft.

***
<a href="https://pr-preview.s3.amazonaws.com/mharbach/design-principles/pull/584.html" title="Last updated on Oct 21, 2025, 5:03 AM UTC (5ecc9c0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/584/444a329...mharbach:5ecc9c0.html" title="Last updated on Oct 21, 2025, 5:03 AM UTC (5ecc9c0)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -160,7 +160,7 @@ If this feature was proposed today, it would probably not proceed.
 <h3 id="user-intent" oldids="consent">Design for user intent</h3>
 
 Use your API design to encourage websites to
-meet user expectations when accessing exceptional capabilities.
+meet user expectations when accessing potentially unsafe capabilities.
 
 In the context of fulfilling a user need,
 a web page may want to make use of a feature


index.bs
```diff
@@ -158,57 +158,75 @@ This is often used to attempt to trick users into visiting scam websites.
...
+(e.g., personal information used, state changed).
+
+Design your API so its behavior matches what people intend and expect.
+Then, in the best case, people don't need to get involved in allowing API access
+and thus can't make decisions they regret later.
+Only ask the user for confirmation in exceptional circumstances.
+
+<p class="example">
+For example, web pages don’t need to ask for permission to play sounds, 
+as they are only able to so when there is enough engagement
```
hober: This is only true in browsers that have some kind of engagement heuristic. Ideally, this example would be reworked in a browser-agnostic way.

index.bs
```diff
@@ -158,57 +158,75 @@ This is often used to attempt to trick users into visiting scam websites.
...
+and it is therefore very likely that the user expects this.
+Similarly, PWAs wanting to use [Window Control Overlay](https://wicg.github.io/window-controls-overlay/) don’t need to ask for permission,
+as the user simply toggles this mode on when they need it.
+</p>
+
+<h4 id="user-decisions">Help users make good decisions</h4>
+
+Asking for user approval has been a common way to handle remaining risk.
```
hober: The first sentence after a heading ought to concisely state the conclusion of the following text. Historically we haven't been entirely consistent in that regard, but that's the goal to strive for.

index.bs
```diff
@@ -158,57 +158,75 @@ This is often used to attempt to trick users into visiting scam websites.
...
+annoyance from frequent interruptions, and regrettable decisions.
+
+When user involvement is unavoidable,
+empower users to make good decisions.
+Use these principles:
+
+* **Limit the consequences from a bad decision by addressing risk at the API level first.**
+    Try not to ask users questions where a "bad" decision can lead to severe consequences,
+    like a compromised device or serious personal data disclosure.
+    If you have to, make it harder for abusive sites to trick users into seeing the question.
```
hober: "If you have to" is probably having the opposite semantic effect from what you intended. This sentence reads like "making it harder for abusive sites to trick users" is optional.

index.bs
```diff
@@ -158,57 +158,75 @@ This is often used to attempt to trick users into visiting scam websites.
...
+    If users’ decisions last longer than the current session,
+    user agents should remind users that their past decision still applies.
+* **Ask questions people can understand.**
+    The associated risks should be obvious from the question for a typical user
+    and be directly connected to the feature's or API's utility.
+    This usually means we can't ask permission to accept fingerprinting risks,
```
hober: It would be nice to make this clause linkable, and to link to it from other places in the document that touch on fingerprinting. Basically, one of the many reasons we need to be reluctant to expand the fingerprinting surface of the web platform is precisely because it's difficult to secure meaningful user consent about it.

index.bs
```diff
@@ -158,57 +158,75 @@ This is often used to attempt to trick users into visiting scam websites.
...
+Being confronted with difficult or annoying questions leads
+to decision fatigue, habitual responses,
+annoyance from frequent interruptions, and regrettable decisions.
+
+When user involvement is unavoidable,
+empower users to make good decisions.
+Use these principles:
+
+* **Limit the consequences from a bad decision by addressing risk at the API level first.**
+    Try not to ask users questions where a "bad" decision can lead to severe consequences,
```
```suggestion
    Do not ask users questions where a "bad" decision can lead to severe consequences,
```

index.bs
```suggestion
```

index.bs
```diff
@@ -158,57 +158,75 @@ This is often used to attempt to trick users into visiting scam websites.
...
+    like a compromised device or serious personal data disclosure.
+    If you have to, make it harder for abusive sites to trick users into seeing the question.
+    If users’ decisions last longer than the current session,
```
```suggestion
* **Provide constant feedback about persistent decisions.**
    If users’ decisions last longer than the current session,
```

index.bs
```suggestion
    For example, this usually means we can't ask permission to accept fingerprinting risks,
```

+1 to Tess here, but this really is just an example.

index.bs
```diff
@@ -158,57 +158,75 @@ This is often used to attempt to trick users into visiting scam websites.
...
+    People need to easily understand who they're dealing with,
+    what the decision is about, and how the website will use any information they might share.
+* **Let people make decisions when *they* are ready.**
+    Websites should not be able to force the decision moment on the user,
+    while they're trying to do something else.
+    Instead, people should be able to choose the decision moment,
+    for example by clicking on a dedicated button like for the file chooser.
+
+If you can’t adhere to all of the above principles,
+then there's a "design smell" and asking the user isn't the right mitigation.
```
```suggestion
If you can’t adhere to all of the above principles,
that is likely an indication that asking the user isn't the right way to manage permission.
```

index.bs
```diff
@@ -158,57 +158,75 @@ This is often used to attempt to trick users into visiting scam websites.
...
+When user involvement is unavoidable,
+empower users to make good decisions.
+Use these principles:
+
+* **Limit the consequences from a bad decision by addressing risk at the API level first.**
+    Try not to ask users questions where a "bad" decision can lead to severe consequences,
+    like a compromised device or serious personal data disclosure.
+    If you have to, make it harder for abusive sites to trick users into seeing the question.
+    If users’ decisions last longer than the current session,
+    user agents should remind users that their past decision still applies.
```
martinthomson, line 208:
```suggestion
    user agents should remind users that their past decision still applies.
    Provide a way to retract any ongoing permission,
    including APIs to inform sites of the change in status.
```

index.bs
```diff
@@ -158,57 +158,75 @@ This is often used to attempt to trick users into visiting scam websites.
...
+    This usually means we can't ask permission to accept fingerprinting risks,
+    as they are subtle, technical, and commonly have nothing to do with the capability itself.
+* **Provide sufficient context.**
+    People need to easily understand who they're dealing with,
+    what the decision is about, and how the website will use any information they might share.
+* **Let people make decisions when *they* are ready.**
+    Websites should not be able to force the decision moment on the user,
+    while they're trying to do something else.
+    Instead, people should be able to choose the decision moment,
+    for example by clicking on a dedicated button like for the file chooser.
```
martinthomson: This one is a bit tricky.

I think that the advice is slightly different here.  If the permission requires -- or might require -- an immediate decision, only activate the permission in response to a user engagement gesture (like a click or keypress).  Any modal dialog, like the file chooser, is an example of that.  Note that we want to allow different browsers to make different choices with respect to modal interaction patterns.

As a general design principle, I agree with the sentiment you have: sites should not be able to force decisions on users, but sometimes that is the best way to manage an interaction.  For instance, I don't think that `getUserMedia` handling as a modal is such a terrible thing, given how it is typically used.

I'm sorry, but I don't have a solid suggestion for this item, but I wanted to flag it for further thought.

index.bs
```diff
@@ -3031,7 +3047,7 @@ you may not need to expose a list to script at all.
 An API which invokes a User-Agent-provided device picker could suffice.
 Such an API:
 - keeps the user in control,
-- doesn't expose any device information without the user's [consent](#consent),
+- [helps users understand](#user-decisions) what they are deciding about,
```
```suggestion
- [helps users understand](#user-decisions) their decisions,
```

index.bs
```diff
@@ -3041,6 +3057,7 @@ the fact that there are devices are available to be picked.
 This does expose one bit of fingerprinting data about the user's environment
 to websites,
 so it isn't quite as safe as an API which doesn't have such a feature.
+Ensure that such an API also provides sufficient information to [help users make good decisions](#user-decisions).
```
martinthomson: Is this an API or an interface?

```suggestion
Ensure that such an API ensures that it is possible to present sufficient information
on user interfaces that users are able to [make good decisions](#user-decisions).
```

index.bs
mharbach: Fair point. We could talk about untrusted 3P content or something like that instead? Or did you have something else in mind, maybe?

index.bs
mharbach: Got it. I added a sentence. Wdyt?

index.bs
mharbach: Huh, thanks for catching that. Updated to "If users are asked".

index.bs
mharbach: great idea. Since I'm not as familiar with the structure of documents like this, what's the best way to make this linkable?

index.bs
mharbach: Thanks for the suggestion. I replaced "mitigation" with "approach" instead, as I don't want to frame it as a decision about managing permission but really about managing risk. The following sentence hopefully communicates that. Wdyt?

index.bs
mharbach: good catch. I like your edit.

index.bs
mharbach: Thank you. Made this change.

index.bs
mharbach: Added this.

index.bs
mharbach: Thanks for your thoughts on this, Martin. Two thoughts in return:

1) I'm not convinced that a simple click or keypress is enough to activate a permission prompt (or really any decision prompt), given how they cheap these are to manufacture for a site with malicious intent. That is why I vaguely wrote "choose the decision moment" and then mentioned a "dedicated button" in the example. In my mind, we ought to recommend something that is a stronger signal of user intent than just any click or keypress.

2) Re: showing a modal as a good way to manage some interactions -- I agree that we can observe examples where this is actually working well today, and video conferencing sites are an example of that. A mapping site can be another, where the user already sees a map on page load and that is enough contextual information to react well to a prompt. At the same time, many VC sites actually don't have this behavior anymore and do display context information before a user's click leads to the permission prompt. I struggle to find a justification to have a kind of carve-out here that allows for interruptive decision UIs to be shown at a site's discretion. In my mind, chances are just so high that this can lead to moments without sufficient context, as we can already observe for geolocation, notification, and FedCM prompts. And on the flip site, encouraging API shapes that in turn encourage sites to provide context information for a user decision seems like a helpful thing.

index.bs
mharbach: make sense. added the prefix.

index.bs
mharbach: That is what I had before but softened it. Let's go with the stronger version.

index.bs
mharbach: Yeah, we probably don't need this if the previous sentence starts with "Do not".

index.bs
martinthomson: I don't see it as necessary to highlight this: it's enough to say that it is easy to retract audio output permission if a site is found to be abusing it AND that browsers have a number of mechanisms for doing just that.  Automatic silencing based on engagement metrics is just one of the more advanced options.

You did miss an opportunity to explain how asking for forgiveness is appropriate only when there is no permanent harm.  The fullscreen API used to be another example, but that's not something that is widely agreed nowadays due to the risk of it being used for takeover attacks.

index.bs
martinthomson: I only mentioned click or keypress as a bare minimum.  I don't think that a specific button adds anything when rendered below the line of death.  As you say, the site has total control of what is shown there and can manufacture clicks.  That undermines my point perhaps, but there has to be a different line through all of this.

This point about modal decisions is one where we should recognize that even if the browser doesn't treat the interaction as modal, a site can always escalate.  They can stop all actions until a decision is made, just as Google Meat does.

Either way, I don't think that this is concrete enough to be actionable.  And we try to make our design principles carry actionable advice.  What do you think are the guiding principles by which we determine that a site and browser jointly ensure that people are ready to make a decision?  I don't think that this is entirely something that a browser can do.

index.bs
```diff
@@ -158,57 +158,79 @@ This is often used to attempt to trick users into visiting scam websites.
...
-
-By asking for consent,
-we can inform the user of what capabilities the web page does or doesn't have,
-reinforcing their confidence that <a href="#safe-to-browse">the web is safe</a>.
-However, the <a href="#priority-of-constituencies">user benefit</a>
-of a new feature must justify the additional burden on users
-to decide whether to grant permission for each feature
-whenever it's requested by a Web page.
+Using such a feature should only be possible
+if the user’s expectation matches the feature’s consequences
```
martinthomson: We have inconsistent apostrophes now.  Just an observation, really.

index.bs
```diff
@@ -158,57 +158,79 @@ This is often used to attempt to trick users into visiting scam websites.
...
+</p>
+
+<h4 id="user-decisions">Help users make good decisions</h4>
+
+Empower good decision-making by limiting risk at the API level and
+providing clarity, context, and user control.
+Asking for user approval has been a common way to handle remaining risk.
+However, most people can't make good decisions on
+the questions they're asked (e.g., permission prompts, FedCM prompts).
+This is because they lack crucial context and information.
```
```suggestion
many of the questions they're asked
because they lack crucial context and information.
```

index.bs
```diff
@@ -581,8 +603,7 @@ if feature detection were available for the feature,
 then you should not support feature detection.
 
 Detecting the availability of a feature does not imply
-detecting whether <a href="#consent">consent</a> to use the feature
-has been granted.
+detecting whether <a href="#userr-decisions">the user has been asked</a> to use the feature.
```
```suggestion
detecting whether [the user has been asked](#user-decisions) to use the feature.
```

Dropping the extra 'r' mostly.

martinthomson (commented):
I wonder if this is a case where the principle needs to acknowledge that there are limits to what a browser can do to provide the necessary context ahead of a decision.  In that way, the best we can expect a browser to achieve is to make it possible - and maybe even easy - to build applications that promote these principles.  There are obvious cases of abuse that browsers can cut back on (popups, etc...), but the platform is too powerful to have all responsibility fall on a browser.

index.bs
mharbach: I agree that this is not quite actionable enough.

> What do you think are the guiding principles by which we determine that a site and browser jointly ensure that people are ready to make a decision?

In my mind, the underlying principle to strive for is that the user's intention and the site's intention are aligned. I.e. either the user already has a goal that plausibly includes access to a capability or the site has done sufficient work to convince the user that the capability is needed. Now that is not necessarily more actionable either, as the browser knows very little about the site's and the user's intent. To approximate them, the best we have been able to come up with is the PEPC concept. If the site places a PEPC on the page, they intent for the capability to be used in the moment when user comes across the PEPC button. JS API calls are a weaker signal, because they only tell us that the developer needed the capability in that moment in the control flow of their application, but that can be disconnected from the journey embodied by what the page looks like at that moment (or in other words, we have heard developers say that they call the JS API without thinking about when the permissions prompt is going to be shown to the user). For the user, we know they intent to use the capability because they just clicked on a button that says something like "use $capability?". This generates a clear signal of aligned intentions. This can of course also happen in other circumstances, but I'm not sure we can get a clear signal for those.

index.bs
mharbach: Thanks for flagging. I'll fix them.

index.bs
mharbach: Fair. I changed the text to use your statement and also added a sentence about asking for forgiveness. Thanks for that idea!

index.bs
mharbach: accepted the suggestion. thanks!

index.bs
mharbach: ah yes. thanks!

index.bs
```diff
@@ -161,57 +161,81 @@ This is often used to attempt to trick users into visiting scam websites.
...
+Only ask the user for confirmation in exceptional circumstances.
+If a capability can be easily reverted or causes only mild annoyance without the chance for permanent harm, 
+afford users an option to stop capability access instead.
+
+<p class="example">
+For example, web pages don't need to ask for permission to play sounds.
+It is easy to stop audio output if a site is found to be abusing it
+and browsers have a number of mechanisms for doing just that.
+Similarly, PWAs wanting to use [Window Control Overlay](https://wicg.github.io/window-controls-overlay/) don't need to ask for permission,
+as the user simply toggles this mode on when they need it.
```
martinthomson: I'm going to suggest that we don't cite a WICG specification here.  It's better to avoid citing stuff that is still in incubation, if only because it depends on details that might change.

```suggestion
```

index.bs
```diff
@@ -161,57 +161,81 @@ This is often used to attempt to trick users into visiting scam websites.
 If this feature was proposed today, it would probably not proceed.
 </p>
 
-<h3 id="consent">Ask users for meaningful consent</h3>
+<h3 id="user-intent">Design for user intent</h3>
```
martinthomson, line 161: We somewhat failed to adhere to our own standards for principles with this one.  We're supposed to provide a simple, positive summary.  How about:

```suggestion

Ensure that it is possible to have the outcome of asking users
for permission to use exceptional capabilities
meet their expectations.

```

That wording is deliberate.  I'll suggest a note below that might make the reasoning clearer.

index.bs
```diff
@@ -161,57 +161,81 @@ This is often used to attempt to trick users into visiting scam websites.
...
+and browsers have a number of mechanisms for doing just that.
+Similarly, PWAs wanting to use [Window Control Overlay](https://wicg.github.io/window-controls-overlay/) don't need to ask for permission,
+as the user simply toggles this mode on when they need it. 
+</p>
+
+<h4 id="user-decisions">Help users make good decisions</h4>
+
+Empower good decision-making by limiting risk at the API level and
+providing clarity, context, and user control.
+Asking for user approval has been a common way to handle remaining risk.
```
```suggestion

Asking for user approval has been a common way to handle remaining risk.
```

index.bs
```diff
@@ -161,57 +161,81 @@ This is often used to attempt to trick users into visiting scam websites.
...
+    Instead, people should be able to choose the decision moment,
+    for example by clicking on a dedicated button like for the file chooser.
+
+If you can't adhere to all of the above principles,
+that is likely an indication that asking the user isn't the right approach.
+Instead, change the feature or API to address any remaining risk.
 
 In your specification, the [=request permission to use=] and [=prompt the user to choose=] algorithms from [[permissions]] are good ways to ask for consent.
 
+If you ask a user to make a decision and they say no, the website shouldn't be able to do anything the user believes they just refused access to.
```
```suggestion
If you ask a user to make a decision and they say no,
the website shouldn't be able to do anything
the user believes they just refused access to. 
```

Just wrapping.

index.bs
```diff
@@ -161,57 +161,81 @@ This is often used to attempt to trick users into visiting scam websites.
...
+    what the decision is about, and how the website will use any information they might share.
+* **Let people make decisions when *they* are ready.**
+    Websites should not be able to force the decision moment on the user,
+    while they're trying to do something else.
+    Instead, people should be able to choose the decision moment,
+    for example by clicking on a dedicated button like for the file chooser.
+
+If you can't adhere to all of the above principles,
+that is likely an indication that asking the user isn't the right approach.
+Instead, change the feature or API to address any remaining risk.
```
martinthomson, line 225:
```suggestion

<div class=note>
A browser is responsible for managing requests for permission.
Browsers will not be able to understand
precisely how a user's interaction with a site
might result in a reasonable expectation of a permission request.
The web platform provides sites with many ways
of both effectively communicating purpose and intent,
or to employ deception and trickery.

The responsibility for clearly communicating intent is therefore shared
between the design of APIs,
user interface design in browsers,
and how websites are constructed.
It is preferable if respect for user intent
is a natural outcome
of following the simple, obvious, or default usage pattern for the API.
This principle only asks that API designs make it possible 
to construct sites that respect user intent.
</div>

```

I'm not sure if this is the best place for this, or the best version of that text, but it's what I'm presently thinking.

I don't know if it is worth highlighting the way in which the geolocation JS API failed in this regard, because sites found it most natural to ask for location the instant that someone arrives on the site.  That subverts intent and can be annoying. What do you think?

martinthomson (commented):
I think we have a pretty good shared understanding now.  I've made some more suggestions that might get this closer to done.

index.bs
mharbach: Thank you for this suggestion, Martin. I think it is an important contribution to make the principle clearer and easier to understand.

Wdyt about making your last sentence a bit stronger, maybe like so:

This principle only asks that API designs **encourage** to construct sites that respect user intent.

index.bs
mharbach: Thanks for fixing that oversight of mine. I would love it if this statement could be a bit stronger, similar to my comment on the note you added at the end. Wdyt about something like this:

Use your API design to encourage websites to meet user expectations when accessing exceptional capabilities.

index.bs
mharbach: @martinthomson What shall we do with this part?

mharbach (commented):
Thanks for your continued support on this, Martin. I like your suggestions.

index.bs
martinthomson: I was trying to be more circumspect, but that's by basic intent, so I'd be OK with that.

index.bs
martinthomson: That works for me.

Oh, and to be clear, it's a pre-existing error, so it's OK if you fail to fix it.

index.bs
martinthomson: I think that the addition I proposed mostly covers that.  I did ask about what you thought about using geolocation as an example of a design that failed to generate the right incentives.  That would be the only thing I can see us adding, maybe.

index.bs
```diff
@@ -161,57 +161,82 @@ This is often used to attempt to trick users into visiting scam websites.
...
-- be able to choose to give or refuse that permission **effectively**.
-
-If a feature is powerful enough to require user consent,
-but it's impossible to explain to a typical user what they are consenting to,
-that's a signal that you may need to reconsider the design of the feature.
-
-If a permission prompt is shown,
-and the user doesn't grant permission,
-the Web page should not be able to do anything
-that the user believes they have refused consent for.
```
dbaron, line 180: For what it's worth, I think this principle got lost in the new text, but it might be worth adding back.  In particular, I think the point here is that if we're designing a permission prompt, the thing we're asking the user about should be something that the browser can effectively deny.   For example, asking the user for camera permission meets this requirement because if the user never grants camera permission, web pages don't have access to the user's camera.  On the other hand, I think there have been a substantial number of suggestions for permission prompts in the privacy or tracking spaces (that weren't actually implemented) that didn't meet this criterion, because they would have been a permission prompt that only gated one of many ways that a web page could get certain information.

mharbach: Thanks for flagging that. The principle is actually still there (lines 231-233) but slightly rephrased. Pasting here for easy reference:

```
If you ask a user to make a decision and they say no,
the website shouldn't be able to do anything
the user believes they just refused access to. 
```

Does that address your concern?

dbaron: Indeed it does; somehow I missed that when looking for it.

index.bs
mharbach: Thank you. Added this in.

index.bs
mharbach: Ah yes, okay. That works for me. I'll remove this part then. I don't think we need to go into the geolocation example. The text of the note seems clear enough to me.

index.bs
mharbach: Okay great. I'll add it in. Thanks for the suggestion! As noted above, I don't think we need to get into the geolocation example. Your text seems clear enough to me already.

index.bs
```diff
@@ -161,57 +161,100 @@ This is often used to attempt to trick users into visiting scam websites.
 If this feature was proposed today, it would probably not proceed.
 </p>
 
-<h3 id="consent">Ask users for meaningful consent</h3>
+<h3 id="user-intent">Design for user intent</h3>
```
jyasskin: Preserve the anchor using https://speced.github.io/bikeshed/#id-gen

```suggestion
<h3 id="user-intent" oldids="consent">Design for user intent</h3>
```

index.bs
```diff
@@ -161,57 +161,100 @@ This is often used to attempt to trick users into visiting scam websites.
...
+Then, in the best case, people don't need to get involved in allowing API access
+and thus can't make decisions they regret later. 
+Only ask the user for confirmation in exceptional circumstances.
+If a capability can be easily reverted or causes only mild annoyance without the chance for permanent harm, 
+afford users an option to stop capability access instead.
+
+<p class="example">
+For example, web pages don't need to ask for permission to play sounds.
+It is easy to stop audio output if a site is found to be abusing it
+and browsers have a number of mechanisms for doing just that.
```
jyasskin:
```suggestion
It is easy to stop audio output if a site is found to be abusing it,
and browsers have a number of mechanisms for doing just that.
```

index.bs
```diff
@@ -161,57 +161,100 @@ This is often used to attempt to trick users into visiting scam websites.
...
-By asking for consent,
-we can inform the user of what capabilities the web page does or doesn't have,
-reinforcing their confidence that <a href="#safe-to-browse">the web is safe</a>.
-However, the <a href="#priority-of-constituencies">user benefit</a>
-of a new feature must justify the additional burden on users
-to decide whether to grant permission for each feature
-whenever it's requested by a Web page.
+Using such a feature should only be possible
+if the user's expectation matches the feature's consequences
+(e.g., personal information used, state changed).
```
jyasskin: I'm stumbling over the (e.g.) bit for some reason. Maybe

```suggestion
if the user's expectation matches the feature's consequences
(e.g., the personal information it reveals or the state it changes).
```

or

```suggestion
if the user's expectation matches the feature's consequences.
For example, the personal information it reveals or the state it changes).
```

index.bs
```diff
@@ -161,57 +161,100 @@ This is often used to attempt to trick users into visiting scam websites.
...
+It is easy to stop audio output if a site is found to be abusing it
+and browsers have a number of mechanisms for doing just that.
+</p>
+
+<h4 id="user-decisions">Help users make good decisions</h4>
+
+Empower good decision-making by limiting risk at the API level and
+providing clarity, context, and user control.
+
+Asking for user approval has been a common way to handle remaining risk.
```
jyasskin:
```suggestion
Asking for user approval has been a common way to handle the remaining risk.
```

?

index.bs
```diff
@@ -161,57 +161,100 @@ This is often used to attempt to trick users into visiting scam websites.
...
+However, most people can't make good decisions on
+many of the questions they're asked
```
jyasskin: I stumbled over "most ... many". Perhaps:

```suggestion
However, people often can't make good decisions on
the questions they're asked  
```

index.bs
```diff
@@ -161,57 +161,100 @@ This is often used to attempt to trick users into visiting scam websites.
...
+Use these principles:
+
+* **Limit the consequences from a bad decision by addressing risk at the API level first.**
+    Do not ask users questions where a "bad" decision can lead to severe consequences,
+    like a compromised device or serious personal data disclosure.
+* **Provide constant feedback about persistent decisions.**   
+    If users' decisions last longer than the current session,
+    user agents should remind users that their past decision still applies.
+    Provide a way to retract any ongoing permission,
+    including APIs to inform sites of the change in status.
```
jyasskin: The way to retract permission doesn't itself include the APIs. So ...

```suggestion
    Provide a way to retract any ongoing permission.,
    APIs should include a way for sites to learn of the change in status.
```

index.bs
```diff
@@ -161,57 +161,100 @@ This is often used to attempt to trick users into visiting scam websites.
...
+to decision fatigue, habitual responses,
+annoyance from frequent interruptions, and regrettable decisions.
+
+When user involvement is unavoidable,
+empower users to make good decisions.
+Use these principles:
+
+* **Limit the consequences from a bad decision by addressing risk at the API level first.**
+    Do not ask users questions where a "bad" decision can lead to severe consequences,
+    like a compromised device or serious personal data disclosure.
```
jyasskin: I'm a little concerned about this absolute statement, when the ubiquitous "download this file" and "fill in this form" features violate it. But principles can always be violated with a good reason, so I don't object to keeping it.

index.bs
```diff
@@ -161,57 +161,100 @@ This is often used to attempt to trick users into visiting scam websites.
...
+* **Provide constant feedback about persistent decisions.**   
+    If users' decisions last longer than the current session,
+    user agents should remind users that their past decision still applies.
+    Provide a way to retract any ongoing permission,
+    including APIs to inform sites of the change in status.
+* **Ask questions people can understand.**
+    The associated risks should be obvious from the question for a typical user
+    and be directly connected to the feature's or API's utility.
```
jyasskin: I stumbled over this sentence. Perhaps:

```suggestion
    When a typical user reads the question about a feature,
    they should immediately think of the associated risks.
    The risks should also be directly connected to the feature's or API's utility.
```

Although this also isn't perfect.

index.bs
```diff
@@ -161,57 +161,100 @@ This is often used to attempt to trick users into visiting scam websites.
...
+    For example, this usually means we can't ask permission to accept fingerprinting risks,
+    as they are subtle, technical, and commonly have nothing to do with the capability itself.
+* **Provide sufficient context.**
+    People need to easily understand who they're dealing with,
+    what the decision is about, and how the website will use any information they might share.
```
jyasskin: "how the website will use any information they might share" isn't usually something the UA can ensure the user is told, especially with our historical resistance to include website-provided strings in browser UI.

I'm ok with keeping the text anyway because I don't have any suggestions for how to fix it.

index.bs
```diff
@@ -161,57 +161,100 @@ This is often used to attempt to trick users into visiting scam websites.
...
+Instead, change the feature or API to address any remaining risk.
+
+<div class=note>
+A browser is responsible for managing requests for permission.
+Browsers will not be able to understand
+precisely how a user's interaction with a site
+might result in a reasonable expectation of a permission request.
+The web platform provides sites with many ways
+of both effectively communicating purpose and intent,
+or to employ deception and trickery.
```
jyasskin, line 234: This isn't parallel...

```suggestion
The web platform provides sites with many ways
either to communicate purpose and intent,
or to employ deception and trickery.
```

index.bs
```diff
@@ -161,57 +161,100 @@ This is often used to attempt to trick users into visiting scam websites.
...
+* **Provide sufficient context.**
+    People need to easily understand who they're dealing with,
+    what the decision is about, and how the website will use any information they might share.
+
+If you can't adhere to all of the above principles,
+that is likely an indication that asking the user isn't the right approach.
+Instead, change the feature or API to address any remaining risk.
+
+<div class=note>
+A browser is responsible for managing requests for permission.
```
jyasskin, line 228: Yay markdown:

```suggestion
<div class=note>

A browser is responsible for managing requests for permission.
```

index.bs
```diff
@@ -161,57 +161,100 @@ This is often used to attempt to trick users into visiting scam websites.
...
+of following the simple, obvious, or default usage pattern for the API.
+This principle only asks that API designs encourage
+to construct sites that respect user intent.
+</div>
 
 In your specification, the [=request permission to use=] and [=prompt the user to choose=] algorithms from [[permissions]] are good ways to ask for consent.
 
+If you ask a user to make a decision and they say no,
+the website shouldn't be able to do anything
+the user believes they just refused access to.
```
jyasskin: The implication of this sentence wasn't clear to me, although that's a pre-existing problem and doesn't need to be fixed in this PR. It could benefit from an example like

```suggestion
If you ask a user to make a decision and they say no,
the website shouldn't be able to do anything
the user believes they just refused access to.
For example, if the user agent asks permission to reveal a piece of information via one API,
the same information shouldn't be available with no permission via a different API.
```

index.bs
```diff
@@ -814,7 +855,8 @@ that site can deny or restrict the user's access to the services it provides.
 People who make use of assistive technologies
 are often [vulnerable members of society](https://www.w3.org/2001/tag/doc/ethical-web-principles/#noharm);
 their use of assistive technologies is [sensitive information](https://www.w3.org/TR/security-privacy-questionnaire/#sensitive-data) about them.
-If an API provides access to this information,
+If an API provides access to this information
+without [asking the user](#user-decisions),
```
jyasskin: I'm not sure it helps to ask the user in this case, although I also don't object to this change.

jyasskin (commented):
I love this change overall. Here are a bunch of optional nitpicks:

index.bs
mharbach: Fair point and I think these features do violate the principle for good reasons. These seem to be table-stakes-type features that enable a large number of use cases. Maybe if they didn't exist today, we might be hesitant to ship them, given what we know about abuse on the web. But yes, I still believe this is a good principle to keep.

index.bs
mharbach: Yes, it is something that we haven't ever solved for on the web platform, but other platforms have and I hope that we can start doing more of it in the future. So this is indeed more aspirational.

index.bs
mharbach: I can imagine features that reveal assistive technology being used that are beneficial to their users, so why wouldn't asking them to enable such a feature not be helpful?

index.bs
```diff
@@ -157,57 +157,104 @@ This is often used to attempt to trick users into visiting scam websites.
...
+
+* **Limit the consequences from a bad decision by addressing risk at the API level first.**
+    Do not ask users questions where a "bad" decision can lead to severe consequences,
+    like a compromised device or serious personal data disclosure.
+* **Provide constant feedback about persistent decisions.**   
+    If users' decisions last longer than the current session,
+    user agents should remind users that their past decision still applies.
+    Provide a way to retract any ongoing permission.
+    APIs should include a way for sites to learn of the change in status.
+* **Ask questions people can understand.**
```
jyasskin: @hober made a [request](https://github.com/w3ctag/design-principles/pull/584#discussion_r2341460179) to make this linkable. It's possible to use an `<a id=stuff>`, but probably [better to use](https://speced.github.io/bikeshed/#self-link)

```suggestion
<li id="ask-understandable-questions"> **Ask questions people can understand.**
```

Because of https://github.com/speced/bikeshed/issues/2184, you'll also have to change the other `*`s to `<li>`s and surround the list with `<ul>`. We can also do this in a separate PR after merging yours.

index.bs
jyasskin: I guess it helps, but it's also not good enough because it allows feature designers to take the shortcut of revealing AT and hoping developers add the special case and don't abuse the information. Instead, we should try to design systems that work for everyone with a single code path.

Also, I think the extra phrase isn't accurate here: if an API provides access to this information, it might be revealed, even if the API asks the user. It's still better to ask than silently leak information, but maybe it'd be better to just leave this section alone and let the rest of the document imply that?

jyasskin (approved):

index.bs
mharbach: Fair. I'll remove the addition.

index.bs
mharbach: Ah yes. Thanks for digging this up. Implemented your proposed change.

index.bs
```diff
@@ -157,57 +157,106 @@ This is often used to attempt to trick users into visiting scam websites.
 If this feature was proposed today, it would probably not proceed.
 </p>
 
-<h3 id="consent">Ask users for meaningful consent</h3>
+<h3 id="user-intent" oldids="consent">Design for user intent</h3>
+
+Use your API design to encourage websites to
+meet user expectations when accessing exceptional capabilities.
```
```suggestion
meet user expectations when accessing potentially unsafe capabilities.
```

martinthomson (approved):

PR #585, opened 2025-09-17
Clarify when details should go on events vs targets

jakearchibald:
I'm proposing this following discussions on `ProgressEvent` in https://github.com/webmachinelearning/translation-api/issues/61.

I wanted to provide reasoning and clarity for this rule, assuming I understand its intent correctly.

***
<a href="https://pr-preview.s3.amazonaws.com/jakearchibald/design-principles/pull/585.html" title="Last updated on Oct 22, 2025, 10:15 AM UTC (fb77d03)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/585/9fc15ba...jakearchibald:fb77d03.html" title="Last updated on Oct 22, 2025, 10:15 AM UTC (fb77d03)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2595,7 +2595,6 @@ Having state on the {{Event/target}} object can used to determine the current st
 without waiting for the next event.
 This is particularly useful if there's a final state,
 where there will be no further events.
-</div>
 
 It's usually not necessary to create new subclasses of {{Event}},
 but they can be used to provide information relating to how the state change occurred.


index.bs
```diff
@@ -2554,12 +2554,30 @@ as it implies that it's possible to dispatch an event asynchronously.
 All events are dispatched synchronously.
 What is more often implied by "asynchronous event" is to defer firing an event.
 
-<h3 id="state-and-subclassing">Use plain {{Event}}s for state</h3>
+<h3 id="state-and-subclassing">Put state on {{Event/target}} objects rather than {{Event}}s</h3>
```
jakearchibald, line 2588: The previous heading seemed… incomplete.

index.bs
```diff
@@ -2554,12 +2554,42 @@ as it implies that it's possible to dispatch an event asynchronously.
...
+An example where this is handled inconsistently is {{XMLHttpRequest}},
+where some request state is on the {{XMLHttpRequest|target}},
+such as {{XMLHttpRequest/readyState}}.
+Whereas other state,
+such as {{ProgressEvent/lengthComputable}} and {{ProgressEvent/total}},
+are only available via a {{ProgressEvent}} object via a {{XMLHttpRequestEventTarget/progress}} event.
+
+If the developer gets access to the {{XMLHttpRequest}} object after the first {{XMLHttpRequestEventTarget/progress}} event,
+the current {{ProgressEvent/lengthComputable}} and {{ProgressEvent/total}} values are inaccessible to them.
+</div>
```
jakearchibald: @lukewarlow done!

index.bs
```diff
@@ -1615,7 +1615,7 @@ Unlike dictionaries, interfaces:
     (i.e., one can check if  it is an `instanceof` a particular class on the global scope),
 
 Defining an interface also exposes it on the global scope, allowing for the specification of static methods.
-For example, the `canParse()` static method of the URL interface. 
+For example, the `canParse()` static method of the URL interface.
```
jakearchibald: Unrelated change, but it's something many editors do automatically. I can remove it, but I imagine others will run into the same thing.

martinthomson: You are not the only one.  We should maybe lint for trailing whitespace.

index.bs
```diff
@@ -2554,12 +2554,45 @@ as it implies that it's possible to dispatch an event asynchronously.
...
+such as {{XMLHttpRequest/readyState}} and {{XMLHttpRequest/status}},
+whereas other state,
+such as {{ProgressEvent/lengthComputable}} and {{ProgressEvent/total}},
+is only available on {{ProgressEvent}} objects via a {{XMLHttpRequestEventTarget/progress}} event.
+
+If the developer gets access to the {{XMLHttpRequest}} object after the first {{XMLHttpRequestEventTarget/progress}} event,
+the current {{ProgressEvent/lengthComputable}} and {{ProgressEvent/total}} values are inaccessible to them.
+
+Instead, all state should have been on the {{XMLHttpRequest}} object,
+and the {{XMLHttpRequestEventTarget/progress}} event should have been a simple {{Event}}.
```
annevk: I'm not sure I agree with this principle. This assumes that tracking this information is free. I don't think it necessarily is. And with the current design only those interested in the information pay for having it tracked.

(This doesn't mean that FetchObserver/FetchMonitor can't expose the information directly, because obtaining such an object could be considered the opt-in.)

index.bs
```diff
@@ -2554,12 +2554,45 @@ as it implies that it's possible to dispatch an event asynchronously.
 All events are dispatched synchronously.
 What is more often implied by "asynchronous event" is to defer firing an event.
 
-<h3 id="state-and-subclassing">Use plain {{Event}}s for state</h3>
+<h3 id="state-and-subclassing">Put state on {{Event/target}} objects rather than {{Event}}s</h3>
 
-Where possible, use a plain {{Event}} with a specified `type`,
-and capture any state information in the {{Event/target}} object.
+Where possible, rather than put state on {{Event}} objects, put state on the {{Event/target}}
+and use events to signal updates to that state.
```
```suggestion
Put state on the {{Event/target}}
and use events to signal updates to that state,
rather than having state on {{Event}} objects.
```

index.bs
```diff
@@ -2554,12 +2554,45 @@ as it implies that it's possible to dispatch an event asynchronously.
...
 
-It's usually not necessary to create new subclasses of {{Event}}.
+<div class="note">
```
martinthomson: Does this need to be a note?

index.bs
```diff
@@ -2554,12 +2554,45 @@ as it implies that it's possible to dispatch an event asynchronously.
...
+This means the {{Event/target}} object can used to determine the current state,
```
```suggestion
Having state on the {{Event/target}} object can used to determine the current state,
```

index.bs
```diff
@@ -2554,12 +2554,45 @@ as it implies that it's possible to dispatch an event asynchronously.
...
+without waiting for the next \'update\' event.
```
```suggestion
without waiting for the next event.
```

index.bs
```diff
@@ -2554,12 +2554,45 @@ as it implies that it's possible to dispatch an event asynchronously.
...
+
+This is particularly useful if there's a final state,
+where there will be no further \'update\' events.
```
```suggestion
This is particularly useful if there's a final state,
where there will be no further events.
```

index.bs
```diff
@@ -2554,12 +2554,45 @@ as it implies that it's possible to dispatch an event asynchronously.
...
+<div class="example">
+Properties on {{HTMLInputElement}},
+such as {{HTMLInputElement/value}},
+provide the state of the input.
+Properties on {{InputEvent}},
+such as {{InputEvent/inputType}},
+describe the nature of an update to the state.
+</div>
+
+<div class="note">
```
```suggestion
<div class="example">
```

index.bs
```diff
@@ -2554,12 +2554,45 @@ as it implies that it's possible to dispatch an event asynchronously.
...
+An example where this is handled inconsistently is {{XMLHttpRequest}}.
+Some state is on the {{XMLHttpRequest|target}},
```
```suggestion
An example where this is handled inconsistently is {{XMLHttpRequest}},
where some state is on the {{XMLHttpRequest|target}},
```

index.bs
```diff
@@ -2554,12 +2554,45 @@ as it implies that it's possible to dispatch an event asynchronously.
...
+such as {{XMLHttpRequest/readyState}} and {{XMLHttpRequest/status}},
+whereas other state,
```
```suggestion
and other state,
```

index.bs
```suggestion
```

I tend to think that you can economise here a little and rely on implying the better design, rather than fully spelling it out.

martinthomson (approved):
This is *way* better than what we had.  Thanks for doing this.  I've made a few suggestions, but feel free to dismiss them.

index.bs
jakearchibald: I did it to separate the rule from the explanation of the rule, since the explanation is non-normative. But this spec doesn't have such a strict divide between normative and non-normative. But I don't mind either way, happy to go with whatever you think.

index.bs
jakearchibald: Fair. I didn't know if counter-examples were true examples.

index.bs
jakearchibald: That's fair.

jakearchibald:
I've removed the XHR counter-example based on @annevk's feedback. I couldn't immediately think of another counter-example.

Happy for this to land once others are.

jakearchibald:
My assumption was that any rules in this doc can be broken if it can be demonstrated that the alternative is better for users/developers in a given case.

@annevk would it help if I added a note recognising the case where the event listener can be used as an opt-in to internal processing? Although, as we see in the fetch case, there are other ways to handle the opt-in.

index.bs
```diff
@@ -2581,6 +2581,13 @@ such as {{InputEvent/inputType}},
 describe the nature of an update to the state.
 </div>
 
+In some exceptional cases, where maintaining state on an object is expensive,
+the state may be placed on the {{Event}} instances,
+allowing {{EventTarget/addEventListener}} and {{EventTarget/removeEventListener}} to signal interest/disinterest in this state.
+Although, other patterns should be considered,
+such as retuning an {{EventTarget}} from a function,
+where the function call is a similar signal of interest.
```
martinthomson: I'm not sure that I'm following this sentence.  I don't know what "retuning" means in this context.

Do you mean that people might want to consider adding a function to EventTarget to get the state instead?  Or are you saying that you might add a function that flips the EventTarget into a mode that tracks the state, so that the cost is only paid when the state is needed?

It would really help if there was a concrete example of this problem.  I'm having a little trouble coming up with one.

annevk: Calling the function returns the EventTarget, at which point the implementation will have to track the state (if it's exposed on the target instead of the event).

jakearchibald: Yeah, so instead of something like:

```js
self.pageDataUsage.addEventListener('change', () => {
  console.log(self.pageDataUsage.bytesReceived);
});
```

…where `self.pageDataUsage.bytesReceived` will need to be continually updated so it can be synchronously accessed by any code running in the doc/worker, you'd do something like:

```js
self.pageDataUsage.addEventListener('change', (event) => {
  console.log(event.bytesReceived);
});
```

…where the data is only available to listeners of the `change` event, so if there are no listeners, the underlying operation to get `bytesReceived` doesn't need to happen. Or, alternatively, "return an {{EventTarget}} from a function":

```js
const pageDataUsage = await self.monitorDataUsage();

pageDataUsage.addEventListener('change', () => {
  console.log(pageDataUsage.bytesReceived);
});
```

In this case, it's the call to `monitorDataUsage()` that signals interest, and GC of `pageDataUsage` will signal disinterest.

The second model gives you an object representing current state that can be passed around.

jakearchibald: `retuning` is a typo in the PR. I'll fix that once we figure out the right wording.

index.bs
xiaochengh: I'm not sure if I understand this paragraph...

Or if it's completely accurate. There are a few events that are fired asynchronously, e.g. `slotchange`. The state on the event target might have been updated when the event listener is triggered.

index.bs
```diff
@@ -2585,12 +2585,61 @@ as it implies that it's possible to dispatch an event asynchronously.
...
+  <pre highlight="js">
+self.dataUsage.addEventListener('change', () => {
+  console.log(self.dataUsage.bytesReceived);
+});
+  </pre>
+
+  Where `bytesReceived` exists on the `dataUsage` {{EventTarget}},
+  rather than the {{Event}}.
+
+  If updating the state on this object is too expensive, this pattern may be considered:
```
```suggestion
  If maintaining the state on this object is too expensive, this pattern may be considered:
```

This approach still assumes that the dataUsage.bytesReceived is maintained, but only when you ask for it.  A truly expensive API might use:

```js
self.addEventListener('dataUsage', e => {
  console.log(e.bytesReceived);
});
```

No need to include that though.

jakearchibald: Wouldn't that be equally expensive as the function example? I guess `removeEventListener` is a more immediate signal of disinterest vs waiting for GC on the object you get back from the function.

index.bs
```diff
@@ -2585,12 +2585,61 @@ as it implies that it's possible to dispatch an event asynchronously.
...
+<h3 id="state-and-subclassing">Put state on {{Event/target}} objects rather than {{Event}}s</h3>
 
-Where possible, use a plain {{Event}} with a specified `type`,
-and capture any state information in the {{Event/target}} object.
+Put state on the {{Event/target}}
+and use events to signal updates to that state,
+rather than having state on {{Event}} objects.
 
-It's usually not necessary to create new subclasses of {{Event}}.
+<div class="note">
```
```suggestion
```
I'm going to say that this is useful text.

index.bs
```diff
@@ -2585,12 +2585,61 @@ as it implies that it's possible to dispatch an event asynchronously.
...
+Having state on the {{Event/target}} object can used to determine the current state,
+without waiting for the next event.
+This is particularly useful if there's a final state,
+where there will be no further events.
+</div>
```
```suggestion
```

martinthomson (approved):

martinthomson:
So am I.  Thanks Jake.

PR #590, opened 2025-09-18
Add principles for task sources

### LINKED ISSUE #38: Guidance on re-using/creating task sources
Tasks created by API specs come from a specific [task source](http://w3c.github.io/html/webappapis.html#task-source). Tasks from different tasks sources may be placed in different task queues. User agents run tasks one after the other but are free to assign different priorities to different task queues (and are free to group tasks from different task sources together).

HTML defines a few [generic task sources](http://w3c.github.io/html/webappapis.html#generic-task-sources). Some specs reuse them. WebRTC [reuses the networking task source](https://w3c.github.io/webrtc-pc/#dfn-networking-task-source) for instance. Other specs create additional task sources such as HTML which also defines the [timer task source](http://w3c.github.io/html/webappapis.html#timer-task-source) or the Generic Sensor API which defines a [sensor task source](https://w3c.github.io/sensors/#sensor-task-source).

Question is: when is it good practice to reuse an existing task source? And when is it good practice to create a new one?

Background for this question is the Presentation API. I suggested to use the networking task source because the tasks this specification creates are network-related. But is it a good idea? That task source would then be used by the Presentation API, WebSockets and WebRTC. Tasks created by WebSockets do not seem to need to run in any specific order compared to tasks created by WebRTC or the Presentation API. The networking task source is also used for image fetching, which seems unrelated as well.

Wouldn't it be preferable to create dedicated task sources when possible, and let user agents assign priorities as they want?

marcoscaceres:
closes #38 

This pull request adds new documentation guidance to the `index.bs` file, advising spec authors to prefer existing task sources rather than creating new ones, unless there is a clear, spec-driven reason. This helps improve interoperability, predictability, and performance scheduling across the platform.

Guidance on task source usage:

* Added a new section recommending the use of generic or preexisting task sources (such as the DOM manipulation, user interaction, networking, navigation, and rendering task sources) when queuing tasks, instead of minting new task sources without a strong, testable need.
* Provided examples and rationale for when a new task source might be justified, such as when isolation from generic sources is required for correct feature behavior.
* Encouraged the use of HTML's wrapper algorithms (`queue a global task`, `queue an element task`) and advised against specifying a particular event loop unless necessary.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2327,6 +2327,34 @@ See also:
 
 * [Command–query separation on Wikipedia](https://en.wikipedia.org/wiki/Command–query_separation)
 
+<h3 id="">Prefer existing task sources; mint new ones only for clear, spec-driven reasons</h3>
+Specs use task sources to group related work so UAs can schedule them sensibly.
+A task is queued on a task source,
+and each task source is associated with a task queue for an event loop.
+Order is guaranteed within a source, while the UA chooses which queue to service next.
+
+Use the generic or preexisting task sources whenever possible.
+HTML defines [[html##generic-task-sources|generic task sources]] meant for broad reuse across specs.
+This includes, but is not limited to:
+the [=DOM manipulation task source=],
+[=user interaction task source=],
+[=networking task source=],
+[=navigation and traversal task source=],
+and [=rendering task source=].
+If your feature matches one of these categories,
+queue on that source rather than creating a new one.
+Doing so improves interoperability, predictability, and performance scheduling across the platform.
+
+Only mint a new task source when you have a concrete, testable need.
+For example when the feature's timing/execution model must be isolated
+from the generic task sources
+(e.g., media elements get their own media element event task source to align with decoder/timeline semantics).
+
+When writing a spec, prefer HTML's wrapper algorithms
+([=queue a global task=], [=queue an element task=])
+and avoid relying on a particular event loop.
+And only be explicit about the event loop when appropriate (see [[#event-design]]).
+
 <h2 id="event-design">Event Design</h2>
 
 <h3 id="one-time-events">Use promises for one time events</h3>


index.bs
```diff
@@ -2327,6 +2327,32 @@ See also:
 
 * [Command–query separation on Wikipedia](https://en.wikipedia.org/wiki/Command–query_separation)
 
+<h3 id="prefer-existing-task-sources">Prefer existing task sources; mint new ones only for clear, spec-driven reasons</h3>
+Specs use task sources to group related work so UAs can schedule them sensibly.
```
```suggestion
Specs use [=task sources=] to group related work (or "[=tasks=]") so UAs can schedule them sensibly.
```

index.bs
```diff
@@ -2327,6 +2327,32 @@ See also:
 
 * [Command–query separation on Wikipedia](https://en.wikipedia.org/wiki/Command–query_separation)
 
+<h3 id="prefer-existing-task-sources">Prefer existing task sources; mint new ones only for clear, spec-driven reasons</h3>
+Specs use [=task sources=] to group related work (or "[=tasks=]") so UAs can schedule them sensibly.
+A task is queued on a task source,
+and each task source is associated with a task queue for an event loop.
```
```suggestion
and each task source is associated with a task queue for an [=event loop=].
```

index.bs
```diff
@@ -2327,6 +2327,32 @@ See also:
...
+Doing so improves interoperability, predictability, and performance scheduling across the platform.
+
+Only mint a new task source when you have a concrete, testable need.
+For example, when the feature's timing/execution model must be isolated
+from the generic task sources
+(e.g., media elements get their own media element event task source to align with decoder/timeline semantics).
+
+When writing a spec, prefer HTML's wrapper algorithms
+([=queue a global task=], [=queue an element task=])
+and avoid relying on a particular event loop.
```
```suggestion
and avoid relying on a particular [=event loop=].
```

index.bs
```diff
@@ -2327,6 +2327,32 @@ See also:
...
+
+Only mint a new task source when you have a concrete, testable need.
+For example, when the feature's timing/execution model must be isolated
+from the generic task sources
+(e.g., media elements get their own media element event task source to align with decoder/timeline semantics).
+
+When writing a spec, prefer HTML's wrapper algorithms
+([=queue a global task=], [=queue an element task=])
+and avoid relying on a particular [=event loop=].
+And only be explicit about the event loop when appropriate (see [[#event-design]]).
```
```suggestion
And only be explicit about the event loop when appropriate (see also [[#event-design]]).
```

alice:
Apologies for the drive-by comment, and this comment applies as much to a number of other principles which have been added in the last few years as it does to this, but:

We wrote a [style guide](https://github.com/w3ctag/process/blob/main/style-guide.md) for TAG design principles which includes [this guidance](https://github.com/w3ctag/process/blob/main/style-guide.md#get-to-the-point):

> When writing guidance, _begin with your main point_. Follow up with brief reasoning, clarifications and caveats in later paragraphs.

(emphasis added)

As an example, I think this guidance should begin with (some version of) the sentence "Use the generic or preexisting task sources whenever possible", rather than with an explanation of background with no indication of where in the text the actual guidance will be given.

index.bs
```diff
@@ -2329,6 +2329,32 @@ See also:
 
 * [Command–query separation on Wikipedia](https://en.wikipedia.org/wiki/Command–query_separation)
 
+<h3 id="prefer-existing-task-sources">Prefer existing task sources; mint new ones only for clear, spec-driven reasons</h3>
```
martinthomson: Alice's guidance isn't too hard to follow.

```suggestion
<h3 id="prefer-existing-task-sources">Prefer existing task sources</h3>

Use existing [=task sources=] for dispatching work where possible.

Only create a new [=task source=] when there is a clear reason to do so.
```

index.bs
```diff
@@ -2329,6 +2329,32 @@ See also:
...
+HTML defines [[html#generic-task-sources|generic task sources]] meant for broad reuse across specs.
+This includes, but is not limited to:
+the [=DOM manipulation task source=],
+the [=user interaction task source=],
+and the [=networking task source=].
+If your feature matches one of these categories,
+queue on that source rather than creating a new one.
+Doing so improves interoperability, predictability, and performance scheduling across the platform.
+
+Only mint a new task source when you have a concrete, testable need.
```
martinthomson, line 2350:
```suggestion
Adding a task source has performance and complexity costs
that all browser implementations pay.
Only mint a new task source when you have a concrete, testable need.
```

PR #592, opened 2025-09-18
Event names should use the present tense

### LINKED ISSUE #280: Guidance for naming events
The only [guidance we currently have about naming events](https://w3ctag.github.io/design-principles/#casing-rules) is that if they consist of multiple words, the words should be concatenated, but there is no guidance on how these words should be selected or ordered.

One very prominent pattern is that most event names end in verbs, and these verbs are in the present tense: `mousemove` (not `mousemoved`), `open` (not `opened`), `load` (not `loaded`), `change` (not `changed`) and so on. Instead, event names with verbs in the past tense like `appinstalled` and `connected` (and its composites, e.g. `gamepadconnected`) are outliers and thus should be considered antipatterns.

Other patterns I observed, which we may or may not want to mention:
- Nouns come before verbs (e.g. `mousemove`, not `movemouse`, `hashchange`, not `changehash`, `timeupdate` not `updatetime`)
- When `before` or `after` are part of the name, they are the first word (e.g. `beforeunload`, `afterscriptexecute`, `beforeinstallprompt`). When `start` or `end` are part of the name, they are the last word (e.g. `transitionstart`, `animationend`, `dragstart`, `audioend`, `touchstart`, `selectionstart`)

[List of all events, for anyone who wants to find more patterns](https://developer.mozilla.org/en-US/docs/Web/Events)

xiaochengh:
Part of #280

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/592.html" title="Last updated on Jun 1, 2026, 5:13 AM UTC (a85c984)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/592/41fee41...a85c984.html" title="Last updated on Jun 1, 2026, 5:13 AM UTC (a85c984)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2397,6 +2397,58 @@ Similarly, add [=event handler IDL attributes=]
 to {{WindowEventHandlers}} rather than {{Window}}.
 </p>
 
+<h3 id="events-are-for-notification">Name events in a consistent pattern</h3>
+
+New event names should follow the format
+
+[before] + {([noun] + verb) | (noun + preposition)} + [start | end] 
+
+Event names should be centered around a verb in the present tense,
+optionally following a noun as the subject or object of the verb.
+It is also allowed to use a preposition in the place of the verb,
+in which case, however, the noun must be included.
+
+This ensures that "on" + event name makes sense.
+
+<div class="example">
+Most event names use the present tense instead of the past tense.
+For example, {{load}} vs. {{loaded}}, {{resize}} vs. {{resized}}.
+
+Events like {{keydown}}, {{keyup}}, {{mouseover}}
+use a preposition instead of a verb.
+
+{{DOMContentLoaded}} is considered an anti-pattern.
+New events should not be named in this format.
+</div>
+
+<!-- before, after, start, end or nothing -->
+
+duration related: start/end
+animationstart, animationend
+
+intervention needed: before, or nothing since most events
+are dispatched before the actual change
+copy, cut, paste
+beforeunload, beforeinput (due to conflicts with unload, input)
+
+are beforeprint/afterprint antipatterns?
+
+When the event is related to a time-consuming procedure
+where both the start and end are of interest,
+there should be a pair of events
+whose names finish with words "start" and "end".
+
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2398,6 +2397,58 @@ Similarly, add [=event handler IDL attributes=]
...
+
+duration related: start/end
+animationstart, animationend
+
+intervention needed: before, or nothing since most events
+are dispatched before the actual change
+copy, cut, paste
+beforeunload, beforeinput (due to conflicts with unload, input)
+
+are beforeprint/afterprint antipatterns?
```
jyasskin: Maybe exceptions, and the whole situation they're meant to address is something of an antipattern. The idea is that you have to actually change the page content to set it up for printing, and then undo that after printing is finished. For that purpose, the `after` seems important.

index.bs
```diff
@@ -2398,6 +2397,58 @@ Similarly, add [=event handler IDL attributes=]
...
+
+When the event is related to a time-consuming procedure
+where both the start and end are of interest,
+there should be a pair of events
+whose names finish with words "start" and "end".
+
+Examples: animationstart, animationend
+Anipattern: beforeprint, afterprint
```
bkardell: antipattern?

csarven (approved):

index.bs
```diff
@@ -2558,6 +2559,24 @@ Similarly, add [=event handler IDL attributes=]
...
 
+<h3 id="name-events-consistently">Name events in a consistent pattern</h3>
+
+In new event names, verbs should be in the present tense.
+Do not use the past tense.
+
+<div class="example">
+Almost all event names use the present tense instead of the past tense.
+For example, {{load}} instead of <code>loaded</code>,
+{{resize}} instead of <code>resized</code>.
```
jyasskin: Probably

```suggestion
For example, {{load}} instead of <code>loaded</code> and
{{resize}} instead of <code>resized</code>.
```

jyasskin (approved):

PR #596, opened 2025-09-21
Recommend registries, and give some guidance on how to define them.

### LINKED ISSUE #583: Guidance on registries?
Hi TAG, in the context of the Digital Credentials API, we've gotten a [detailed issue with arguments against the use of a registry](https://github.com/w3c-fedid/digital-credentials/issues/345) for protocol elements. The FedID WG will consider these points when deciding whether/how to use registries for our specs. I'm sending the general discussion over to TAG to see if there's existing guidance. Thanks!

jyasskin:
Here is a "define registries" principle, based on the [discussion in Hong Kong last week](https://github.com/w3ctag/meetings/blob/gh-pages/2025/09-Hong-Kong/09-15-minutes.md#when-and-how-are-registries-appropriate). It will fix #583 once it's polished and merged.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/design-principles/pull/596.html#registries" title="Last updated on Dec 18, 2025, 9:39 PM UTC (91b96c3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/596/031a3de...jyasskin:91b96c3.html" title="Last updated on Dec 18, 2025, 9:39 PM UTC (91b96c3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3793,6 +3793,49 @@ can be either required or optional for implementations to recognize.
 For extension points that can't just be ignored when their extension isn't recognized,
 the registry should include at least 1 required entry.
 
+<div class="example" heading="Link Relations" id="example-link-relation-registry">
+
+Link relations are currently registered in a 3-part registry.
+[[rfc8288#procedure|RFC 8288]] defines the [[IANA-RELATIONS inline]] registry
+which accepts entries that "reference a freely available, stable specification."
+[[MFREL inline]] defines a second registry
+to which anyone with a wiki account can add.
+This registry has a column for specifications,
+but nobody enforces that the column is filled in
+or that the specifications meet any particular quality requirements.
+[[html#linkTypes]] then repeats a subset of these relation types
+to define how they work in web browsers.
+
+In an ideal world, [[RFC8288]] and [[MFREL inline]] would probably be a single registry,
+with [[html#linkTypes|HTML]] defining browser behavior.
+The current split is probably due to mistakes in the historical registry requirements.
+
+The idea of <{a/rel}> as list of externally-defined keywords
+dates to at least [[rfc1866 inline obsolete]], in 1995, which said
+
+> The REL attribute gives the relationship(s) described by the hyperlink.
+> The value is a whitespace separated list
+> of relationship names. The semantics of link
+> relationships are not specified in this document.
+
+The [[html401 inline]], in 1999,
+defined a list of [known link relations](https://www.w3.org/TR/html401/types.html#type-links)
+but continued to encourage authors to invent their own.
+
+The IETF eventually defined the IANA registry in [[rfc4287 inline]], in 2005,
+and [[rfc5988 inline obsolete]], in 2010.
+These required specifications and expert review,
+but by this time too many custom link relations were in use
+to get them all specified and registered.
+Even if the registry had been in place from the beginning,
+we believe this is a case where implementers are too diverse
+to even know what kind of specification to expect for registered elements.
+The modern [[html#linkTypes|HTML]] table
+is a good way to align browser implementations
+on the subset of link relations that they react to.
+
+</div>
+
 See [[qaframe-spec#extensions]] and [[RFC6709]] for
 more guidance on how to design extensibility.
 
--- a/index.bs
+++ b/index.bs
@@ -3734,50 +3734,65 @@ Default to defining either
 an IANA registry governed by [[RFC8126]] or a [[w3c-process#registries|W3C Registry]],
 but another similar mechanism is also acceptable.
 
-Registries are helpful because they
+The place in a specification that dispatches to an extension's implementation
+is known as an <dfn export>extension point</dfn>.
+To maximize the chance that a specification with extension points is interoperably implemented,
+each extension point should
 
-*   reduce name collisions,
-1. Help readers find the name for a desired purpose.
-1. Direct readers to a specification for each name
-    (if the registry requires this).
+* define how implementations can negotiate which extensions are acceptable,
+* define what to do with unrecognized extensions
+    (see [[RFC6709#section-4|section 4 of RFC6709]]),
+* define what interface extensions are supposed to implement, and
+* link to a specification of each extension that is
+    detailed enough to support interoperable implementations of the extension,
+    as is required by
+    the IETF's [[RFC8126#section-4.6|Specification Required]] registration policy.
 
-Items in a registry generally need to implement a defined interface
-in order to plug into the specification they're extending.
-The definition of the registry must be clear about this interface,
-and it should require new entries to include a specification that follows it.
-This roughly corresponds to
-the IETF's [[RFC8126#section-4.6|Specification Required]] registration policy.
-It's difficult to be confident in an interface definition
-before that interface has several implementations,
-so any new registry should start with at least 2-3 entries defined.
+Registries help because they
+
+* give each extension a unique name to use
+    in negotiation and
+    in recognizing what extension is in use,
+* provide a place to link to extension specifications
+    (if the registry requires this), and
+* help readers find the name for a desired purpose.
 
 It is tempting to additionally require that registry entries be
 "good" in some way beyond what is needed to achieve interoperability.
-Use a permissive registration policy instead of doing this
-unless the feature includes
-some way to enforce that no implementation can use an unregistered name.
-The IETF has found that
-when it's too hard to add entries to a registry,
-implementers will often simply use names without registering them.
+Whether this can succeed depends on the ecosystem and the expected implementers.
+Implementers are most likely to be willing to
+navigate a demanding registration process and
+constrain their implementations to match strict registration requirements
+when they are a small set, wealthy, and generally-aligned,
+as in the case of web browser engines.
+The more diverse or constrained implementers become,
+the less you can expect them to consistently work to register extensions.
 
-When a feature's main specification defines some initial registry entries,
-each one can be either required or optional for implementations to recognize.
-If implementations need to pick an understood registry entry to send to their recipients,
-then the registry should include at least 1 required entry.
-If implementations can be
-divided into distinct [[spec-variability#subdivision-profile|profiles]]
-that tend not to need to communicate with other profiles,
-it can be sufficient for each profile to require a registry entry
-even if the main specification leaves them all optional.
+At the limit, implementers might not even be willing to specify their extensions.
+If the specification authors consider this likely, it may be worth allowing
+[[rfc8126#section-4.4|first-come-first-served registrations without a specification]]
+just to reduce the risk of name collisions,
+although the registry should still encourage full specifications.
 
-It can be tempting to use URLs instead of registered strings to identify extensions.
-This has the benefit of making it very easy to extend the feature,
+In the case of a registry that doesn't require specifications,
+it can be tempting to identify extensions with URLs or URIs instead of registered strings.
+This has the effect of defining a [[rfc8126#section-4.3|hierarchical registration policy]]
+and making it very easy to extend the feature,
 by just picking a URL from a domain that the extender controls.
-However, there's no way to ensure that
-these URLs point to a specification for the extension they name,
-or even that the domain stays under the control of the entity that defined the extension.
-URLs are appropriate for a few kinds of completely-permissionless extension,
+This clearly loses the interoperability benefits of requiring a specification,
+and in the case of DNS-based URLs,
+it also risks that the entity that defined the extension may lose control of its domain.
+URIs are appropriate for a few kinds of very-low-coordination extension,
 but most of the time a WG-managed permissive registry table will work better.
+
+Because an [=extension point=] defines an interface,
+and it's difficult to be confident in an interface definition
+before that interface has several implementations,
+any new registry should start with at least 2-3 entries defined.
+Each of these initial registry entries
+can be either required or optional for implementations to recognize.
+For extension points that can't just be ignored when their extension isn't recognized,
+then the registry should include at least 1 required entry.
 
 See [[qaframe-spec#extensions]] and [[RFC6709]] for
 more guidance on how to design extensibility.
--- a/index.bs
+++ b/index.bs
@@ -3708,6 +3708,80 @@ and makes it clear when the state described by the flags is reset.
 
 <!-- TODO: add examples -->
 
+<h3 id="registries">Use registries to allow constrained extensions outside the original standards track</h3>
+
+When you expect a feature to need to be extended over time,
+and it would be inappropriate to
+manage that change through the same standards process that defined the original feature,
+then a registry is usually the right design.
+
+[[spec-variability#variability|"Variability complicates interoperability."]]
+When a feature needs optional components,
+it is easiest to manage the complication if
+a single group is in charge of the whole design.
+This implies that most optional features should be defined
+in the same specification that defined the original feature.
+
+Sometimes new extensions are expected to be needed in the future.
+In this case, the right plan is often to
+have the original standards body continue or restart
+a working group to produce an update to the feature's full specification.
+
+However, if extensions need to be defined through
+a different process than updating the full specification,
+a registry is usually the right way to manage the known extensions.
+Default to defining either
+an IANA registry governed by [[RFC8126]] or a [[w3c-process#registries|W3C Registry]],
+but another similar mechanism is also acceptable.
+
+Registries are helpful because they
+
+1. Reduce name collisions.
+1. Help readers find the name for a desired purpose.
+1. Direct readers to a specification for each name
+    (if the registry requires this).
+
+Items in a registry generally need to implement a defined interface
+in order to plug into the specification they're extending.
+The definition of the registry must be clear about this interface,
+and it should require new entries to include a specification that follows it.
+This roughly corresponds to
+the IETF's [[RFC8126#section-4.6|Specification Required]] registration policy.
+It's difficult to be confident in an interface definition
+before that interface has several implementations,
+so any new registry should start with at least 2-3 entries defined.
+
+It is tempting to additionally require that registry entries be
+"good" in some way beyond simply being specified.
+Use a permissive registration policy instead of doing this
+unless the feature includes
+some way to enforce that no implementation can use an unregistered name.
+The IETF has found that
+when it's too hard to add entries to a registry,
+implementers will often simply use names without registering them.
+
+When a feature's main specification defines some initial registry entries,
+each one can be either required or optional for implementations to recognize.
+If implementations need to pick an understood registry entry to send to their recipients,
+then the registry should include at least 1 required entry.
+If implementations can be
+divided into distinct [[spec-variability#subdivision-profile|profiles]]
+that tend not to need to communicate with other profiles,
+it can be sufficient for each profile to require a registry entry
+even if the main specification leaves them all optional.
+
+It can be tempting to use URLs instead of registered strings to identify extensions.
+This has the benefit of making it very easy to extend the feature,
+by just picking a URL from a domain that the extender controls.
+However, there's no way to ensure that
+these URLs point to a specification for the extension they name,
+or even that the domain stays under the control of the entity that defined the extension.
+URLs are appropriate for a few kinds of completely-permissionless extension,
+but most of the time a WG-managed permissive registry table will work better.
+
+See [[qaframe-spec#extensions]] for
+more guidance on how to design extensibility.
+
 <h3 id="implementability">Resolving tension between interoperability and implementability</h3>
 
 <!--
--- a/index.bs
+++ b/index.bs
@@ -3793,6 +3793,28 @@ can be either required or optional for implementations to recognize.
 For extension points that can't just be ignored when their extension isn't recognized,
 the registry should include at least 1 required entry.
 
+<div class=example heading="Well-Known URIs" id="example-well-known">
+
+The [Well-Known URI registry](https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml),
+defined by [[RFC8615]],
+ensures that protocols don't collide when allocating a path in all web servers.
+
+Part of the purpose of HTTP and similar URL schemes
+is to allow new protocols to be defined on top of them,
+without requiring an update to the core HTTP specification for each new protocol.
+When these protocols need to retrieve information about an entire origin,
+the only effect on HTTP as a whole would come from potential name collisions.
+Since the registry itself mitigates that risk,
+there isn't a need for the whole HTTP community
+to agree on a new HTTP revision for each registration.
+
+The registry's requirement for a specification isn't too much of a burden
+because these cross-origin protocols
+tend to be defined by an agreement between the participating origins,
+which is documented in a specification.
+
+</div>
+
 <div class="example" heading="Link Relations" id="example-link-relation-registry">
 
 Link relations are currently registered in a 3-part registry.
--- a/index.bs
+++ b/index.bs
@@ -3749,12 +3749,11 @@ each extension point should
 
 Registries help because they
 
-* give each extension a unique name to use
-    in negotiation and
-    in recognizing what extension is in use,
-* provide a place to link to extension specifications
+* create a place where extension developers can coordinate
+    the allocation of identifiers for their extensions,
+* provide a place for implementers to find extension specifications
     (if the registry requires this), and
-* help readers find the name for a desired purpose.
+* help readers discover when existing extensions address their needs.
 
 It is tempting to additionally require that registry entries be
 "good" in some way beyond what is needed to achieve interoperability.
@@ -3762,12 +3761,12 @@ Whether this can succeed depends on the ecosystem and the expected implementers.
 Implementers are most likely to be willing to
 navigate a demanding registration process and
 constrain their implementations to match strict registration requirements
-when they are a small set, wealthy, and generally-aligned,
+when they are a small set, well-resourced, and generally-aligned,
 as in the case of web browser engines.
 The more diverse or constrained implementers become,
 the less you can expect them to consistently work to register extensions.
 
-At the limit, implementers might not even be willing to specify their extensions.
+At the limit, implementers might not even be willing to document their extensions.
 If the specification authors consider this likely, it may be worth allowing
 [[rfc8126#section-4.4|first-come-first-served registrations without a specification]]
 just to reduce the risk of name collisions,
@@ -3782,12 +3781,13 @@ This clearly loses the interoperability benefits of requiring a specification,
 and in the case of DNS-based URLs,
 it also risks that the entity that defined the extension may lose control of its domain.
 URIs are appropriate for a few kinds of very-low-coordination extension,
+such as the definition of an XML namespace [[XML-NAMES]],
 but most of the time a WG-managed permissive registry table will work better.
 
 Because an [=extension point=] defines an interface,
 and it's difficult to be confident in an interface definition
 before that interface has several implementations,
-any new registry should start with at least 2-3 entries defined.
+any new registry should start with more than one entry defined.
 Each of these initial registry entries
 can be either required or optional for implementations to recognize.
 For extension points that can't just be ignored when their extension isn't recognized,
--- a/index.bs
+++ b/index.bs
@@ -3724,15 +3724,14 @@ in the same specification that defined the original feature.
 
 Sometimes new extensions are expected to be needed in the future.
 It is often most appropriate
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3567,6 +3567,80 @@ and makes it clear when the state described by the flags is reset.
...
+When a feature needs optional components,
+it is easiest to manage the complication if
+a single group is in charge of the whole design.
+This implies that most optional features should be defined
+in the same specification that defined the original feature.
+
+Sometimes new extensions are expected to be needed in the future.
+In this case, the right plan is often to
+have the original standards body continue or restart
+a working group to produce an update to the feature's full specification.
```
```suggestion
It is often most appropriate
to have revise the specification to define an extension.
This can reduce or avoid any need to precisely define requirements for extensions.
```

index.bs
```diff
@@ -3567,6 +3567,80 @@ and makes it clear when the state described by the flags is reset.
...
+This has the benefit of making it very easy to extend the feature,
+by just picking a URL from a domain that the extender controls.
+However, there's no way to ensure that
+these URLs point to a specification for the extension they name,
+or even that the domain stays under the control of the entity that defined the extension.
+URLs are appropriate for a few kinds of completely-permissionless extension,
+but most of the time a WG-managed permissive registry table will work better.
+
+See [[qaframe-spec#extensions]] for
+more guidance on how to design extensibility.
```
```suggestion
See [[qaframe-spec#extensions]] and [[RFC6709]] for
more guidance on how to design extensibility.
```

index.bs
```diff
@@ -3567,6 +3567,80 @@ and makes it clear when the state described by the flags is reset.
...
+In this case, the right plan is often to
+have the original standards body continue or restart
+a working group to produce an update to the feature's full specification.
+
+However, if extensions need to be defined through
+a different process than updating the full specification,
+a registry is usually the right way to manage the known extensions.
+Default to defining either
+an IANA registry governed by [[RFC8126]] or a [[w3c-process#registries|W3C Registry]],
+but another similar mechanism is also acceptable.
```
martinthomson: What other mechanism do you think might be appropriate?  What criteria do you think apply to the selection of process?

index.bs
```diff
@@ -3567,6 +3567,80 @@ and makes it clear when the state described by the flags is reset.
...
+
+Registries are helpful because they
+
+1. Reduce name collisions.
```
martinthomson: I think that you have a bullet list with commas, not periods...

```suggestion
*   reduce name collisions,
```

index.bs
```diff
@@ -3567,6 +3567,80 @@ and makes it clear when the state described by the flags is reset.
...
+1. Help readers find the name for a desired purpose.
+1. Direct readers to a specification for each name
+    (if the registry requires this).
+
+Items in a registry generally need to implement a defined interface
+in order to plug into the specification they're extending.
+The definition of the registry must be clear about this interface,
+and it should require new entries to include a specification that follows it.
```
martinthomson: The extension point that is governed by a registry needs to be fully-specified.  That includes a definition of what to do if an item is not understood (is there a negotiation mechanism?  is there feature detection? is the extension ignored when it is not understood? or, does it cause a failure of some sort?)

RFC 6709 goes into some detail about some of the extension models that might apply.  Those are for protocols, but the same principles broadly apply to APIs as well.

Then, the requirements for what it takes to define the extension need to be clearly defined, as you say here.

index.bs
```diff
@@ -3567,6 +3567,80 @@ and makes it clear when the state described by the flags is reset.
...
+This roughly corresponds to
+the IETF's [[RFC8126#section-4.6|Specification Required]] registration policy.
+It's difficult to be confident in an interface definition
+before that interface has several implementations,
+so any new registry should start with at least 2-3 entries defined.
+
+It is tempting to additionally require that registry entries be
+"good" in some way beyond simply being specified.
```
```suggestion
"good" in some way beyond what is needed to achieve interoperability.
```

index.bs
```diff
@@ -3567,6 +3567,80 @@ and makes it clear when the state described by the flags is reset.
...
+Use a permissive registration policy instead of doing this
+unless the feature includes
+some way to enforce that no implementation can use an unregistered name.
+The IETF has found that
+when it's too hard to add entries to a registry,
+implementers will often simply use names without registering them.
```
martinthomson: This isn't quite the angle I'd use here.  The extent to which you can add tighter constraints varies based on the ecosystem you are deploying into.  If your implementers are browsers, then you might imagine that a tighter set of restrictions might be respected and therefore it would be more reasonable to at least try it on.

If you are looking at websites implementing, then you can almost guarantee that registrations won't happen, even if the constraints are very lightweight.

Getting that nuance in might be a good idea.

index.bs
```diff
@@ -3567,6 +3567,80 @@ and makes it clear when the state described by the flags is reset.
...
+
+When a feature's main specification defines some initial registry entries,
+each one can be either required or optional for implementations to recognize.
+If implementations need to pick an understood registry entry to send to their recipients,
+then the registry should include at least 1 required entry.
```
martinthomson: Extensions for optional features - ones that are simply ignored if not understood - don't need this.

You should say that where the functioning of a specification depends on choosing and using an registered value, then you have three choices: 

* One party implements all the options.
* You mandate a baseline that all parties implement.
* Risk interoperability failure if the intersection of options between entities is empty.

And that's for a two-party negotiation.  There are many things that the W3C documents where there are more than two potential actors involved.

index.bs
```diff
@@ -3567,6 +3567,80 @@ and makes it clear when the state described by the flags is reset.
...
+even if the main specification leaves them all optional.
+
+It can be tempting to use URLs instead of registered strings to identify extensions.
+This has the benefit of making it very easy to extend the feature,
+by just picking a URL from a domain that the extender controls.
+However, there's no way to ensure that
+these URLs point to a specification for the extension they name,
+or even that the domain stays under the control of the entity that defined the extension.
+URLs are appropriate for a few kinds of completely-permissionless extension,
+but most of the time a WG-managed permissive registry table will work better.
```
martinthomson: XML namespaces use URIs (not URLs).  They are not using URIs for their ability to resolve to something; most XML namespace URIs, if they are resolved, point to nothing useful.  You seem to imply that they could point to something; some do, but that's never been the goal.

They use URIs because the domain into which they are being deployed does not benefit from centralized coordination.  They need identifiers and a way to generate them without coordination or - save us all - namecoin-style consensus.  Indeed, it would be a mistake to require centralized coordination for those (doubly so for a wasteful distributed consensus).  XML is a file format that is used by many people for many reasons.  There is no value in asking people to coordinate identifier allocation.  The same applies to JSON-LD, for similar reasons [^1].

[^1]: Even if you don't think that JSON-LD is solving a problem that matters, this choice is not one that is really in contention, to my mind.

But the question we need to answer here is different.  If you are specifying an extension to a protocol, why would you use a registry rather than letting the proverbial thousand flowers bloom?  That all comes down to how you conceive of the interoperability goals of the extension point.

An optional, ignorable extension might benefit from a registry if there is a greater need to have different implementations agree on what the meaning of the extension is.  In TLS, we have a registry for extensions, not just because there is a limited space available, but because the goal of defining an extension is to enable a shared understanding of the space.

In comparison, the IPTC metadata format uses URLs, because they don't see a need to constrain the space of options.  There, the risk that people define multiple identifiers that are close, or even identical, in meaning is something that is tolerated so that there is a low-coordination way of adding metadata for limited or proprietary uses.

An extension point that determines everything that follows, like TLS ciphersuites or digital credential protocols, benefits more from coordination than something that can be readily ignored.  Those are places where mandating implementation of a core by all (or a subset by browsers) is necessary.

index.bs
```diff
@@ -3567,6 +3567,80 @@ and makes it clear when the state described by the flags is reset.
...
+
+When a feature's main specification defines some initial registry entries,
+each one can be either required or optional for implementations to recognize.
+If implementations need to pick an understood registry entry to send to their recipients,
+then the registry should include at least 1 required entry.
+If implementations can be
+divided into distinct [[spec-variability#subdivision-profile|profiles]]
+that tend not to need to communicate with other profiles,
+it can be sufficient for each profile to require a registry entry
+even if the main specification leaves them all optional.
```
martinthomson: That's not interoperability in the sense that we should care about here.

https://www.w3.org/TR/ethical-web-principles/#oneweb

martinthomson (changes requested):

index.bs
annevk: XML namespaces use strings that by convention often look like URLs and might parse as one, if we're going to quibble about details.

index.bs
jyasskin: My thought here was that, if we were writing this several years ago I would have put just the IANA process here. When the W3C process arrived, I would have wanted it to be written in such a way that the W3C process were also acceptable. I don't know of another process that would fit now, but there's also no reason that only 2 organizations would be acceptable.

On the other hand, maybe this is an extensibility point for the design principles, and we can just update the design principles instead of defining a registry of registry mechanisms with criteria for adding new ones.

```suggestion
Default to defining either
an IANA registry governed by [[RFC8126]] or a [[w3c-process#registries|W3C Registry]].
```

index.bs
jyasskin: Works for me.

index.bs
jyasskin: I've rearranged some things to add this in, and to defer to RFC 6790 for details. How does it look?

index.bs
jyasskin: Good point. I replaced the tail of this paragraph with this discussion, and added a paragraph saying (roughly) that if this is likely, then it can justify a FCFS-with-no-spec policy.

index.bs
jyasskin: I'm happy to drop this sentence, although DIDs might be an example where there are many specifications (protocols) that have fields in which a DID could appear, and each of those specifications needs to mandate at least 1 DID Method to be implemented, but not all of those specifications need to agree on the same Method. I don't think that would violate the One Web principle, although it also wouldn't be ideal.

index.bs
jyasskin: I tried to incorporate optional features.

I feel like we should only endorse one of your three choices: 
* having 1 party implement all options means that party gets a veto on new options ... maybe it's saying that all values have to be marked as required for one side of the protocol so that the other side can make a free choice ... and in that case, values can only be added in a revision of the main spec, so it's not a registry anymore.
* the baseline is equivalent to requiring at least 1 entry
* We shouldn't endorse a plan that leads to interoperability failures.

index.bs
jyasskin: After some of the rewrites above, the paragraph about URIs/URLs wound up fitting in as a thing one might do if the alternative is a specification-less registry. They're [hierarchical in the IETF sense](https://www.rfc-editor.org/rfc/rfc8126#section-4.3), but they also tend to risk a DNS entry being bought out from under the original creator. And they're usually more bytes than really necessary, but that's petty and I didn't find a convenient way to work it in.

The [IPTC example](https://iptc.org/std/photometadata/specification/IPTC-PhotoMetadata#xmp-namespaces-and-identifiers) seems like a case where they should have defined the main terms in the main spec, or possibly done a hierarchical space where each "Recommended/used Prefix" is controlled by a particular entity.

Or are you thinking that we should call out the case of enabling local/proprietary extensions as a time when URLs are actually the right way to go?

jyasskin (commented):
Sorry for the long delay. How's it look now?

matatk (commented):
This is great.

Do you think that some concrete examples might help?

In terms of positive examples, there are a few that seem to work well, including: link types/`rel` attribute values, and Well-known URIs.

But are there any non-contentious, and preferably not recent, examples of where a registry was implemented, but wasn't a good idea, or was decided against?

index.bs
```diff
@@ -3708,6 +3708,137 @@ and makes it clear when the state described by the flags is reset.
 
 <!-- TODO: add examples -->
 
+<h3 id="registries">Use registries to allow constrained extensions outside the original standards track</h3>
```
martinthomson: Is this really the right title now that you have done the work to change the focus?

index.bs
```diff
@@ -3708,6 +3708,137 @@ and makes it clear when the state described by the flags is reset.
...
+
+When you expect a feature to need to be extended over time,
```
martinthomson, line 3713: This needs a punchy "Use an appropriate mechanism to manage the future extensibility of your work." Or similar.

index.bs
```diff
@@ -3708,6 +3708,137 @@ and makes it clear when the state described by the flags is reset.
...
+then a registry is usually the right design.
+
+[[spec-variability#variability|"Variability complicates interoperability."]]
+When a feature needs optional components,
+it is easiest to manage the complication if
+a single group is in charge of the whole design.
+This implies that most optional features should be defined
+in the same specification that defined the original feature.
+
+Sometimes new extensions are expected to be needed in the future.
```
martinthomson: This paragraph isn't really about the expectation that extensions might be needed, but about keeping control within the group that defined the original feature.  To that end, it belongs with the previous paragraph.  I'd still use a new paragraph, but use a bridging statement, not something like what you have here, which implies the situation is very different.

index.bs
```diff
@@ -3708,6 +3708,137 @@ and makes it clear when the state described by the flags is reset.
...
+It is often most appropriate
+to have to revise the specification to define an extension.
+This can reduce or avoid any need to precisely define requirements for extensions.
+
+However, if extensions need to be defined through
```
martinthomson: If the substance of this section, the thing introducing it shouldn't start with "However,".

> Where there is strong reason to believe that a feature will be used *and extended* by a diverse set of users, including those who might not be involved in its original creation, a registry is usually the right way to manage extensibility.
>
> A registry is a central location for identifying and managing extensions. Registries can be defined as a [[...|W3C Registry]], where the W3C as a whole, often delegated to a working group, is responsible for the maintenance of the registry. <abbr title="Internet Assigned Numbers Authority">IANA</abbr> might also be asked to establish and maintain registries, following the procedures in [[RFC8126]].

index.bs
```diff
@@ -3708,6 +3708,137 @@ and makes it clear when the state described by the flags is reset.
...
+an IANA registry governed by [[RFC8126]] or a [[w3c-process#registries|W3C Registry]].
+
+The place in a specification that dispatches to an extension's implementation
+is known as an <dfn export>extension point</dfn>.
+To maximize the chance that a specification with extension points is interoperably implemented,
+each extension point should
+
+* define how implementations can negotiate which extensions are acceptable,
+* define what to do with unrecognized extensions
+    (see [[RFC6709#section-4|section 4 of RFC6709]]),
```
martinthomson: These two points are an "or" situation.  If you negotiate, then you might never have to deal with unrecognized extensions, except to the extent they might be mentioned in negotiation.  Or, you might not negotiate, in which case you might need to deal with unrecognized extension in different places.

index.bs
```diff
@@ -3708,6 +3708,137 @@ and makes it clear when the state described by the flags is reset.
...
+a different process than updating the full specification,
+a registry is usually the right way to manage the known extensions.
+Default to defining either
+an IANA registry governed by [[RFC8126]] or a [[w3c-process#registries|W3C Registry]].
+
+The place in a specification that dispatches to an extension's implementation
+is known as an <dfn export>extension point</dfn>.
+To maximize the chance that a specification with extension points is interoperably implemented,
+each extension point should
+
```
martinthomson, line 3739:
```suggestion

* define how extensions are uniquely identified,
```

index.bs
```diff
@@ -3708,6 +3708,137 @@ and makes it clear when the state described by the flags is reset.
...
+* link to a specification of each extension that is
+    detailed enough to support interoperable implementations of the extension,
+    as is required by
+    the IETF's [[RFC8126#section-4.6|Specification Required]] registration policy.
+
+Registries help because they
+
+* give each extension a unique name to use
+    in negotiation and
+    in recognizing what extension is in use,
```
```suggestion
* give each extension a unique identifier to use
    either in negotiation or
    in recognizing when an extension is in use,
```

index.bs
```diff
@@ -3708,6 +3708,137 @@ and makes it clear when the state described by the flags is reset.
...
+* provide a place to link to extension specifications
```
```suggestion
* provide a place for implementers to find extension specifications
```

index.bs
```diff
@@ -3708,6 +3708,137 @@ and makes it clear when the state described by the flags is reset.
...
+    (if the registry requires this), and
+* help readers find the name for a desired purpose.
```
```suggestion
* help readers discover when existing extensions address their needs.
```

index.bs
```diff
@@ -3708,6 +3708,137 @@ and makes it clear when the state described by the flags is reset.
...
+
+It is tempting to additionally require that registry entries be
+"good" in some way beyond what is needed to achieve interoperability.
+Whether this can succeed depends on the ecosystem and the expected implementers.
+Implementers are most likely to be willing to
+navigate a demanding registration process and
+constrain their implementations to match strict registration requirements
+when they are a small set, wealthy, and generally-aligned,
```
```suggestion
when they are a small set, well-resourced, and generally-aligned,
```

index.bs
martinthomson: Or


```suggestion
* create a place where extension developers can coordinate
    the allocation of identifiers for their extensions,
```

With the above point about unique identifiers, above.

index.bs
```diff
@@ -3708,6 +3708,137 @@ and makes it clear when the state described by the flags is reset.
...
+as in the case of web browser engines.
+The more diverse or constrained implementers become,
+the less you can expect them to consistently work to register extensions.
+
+At the limit, implementers might not even be willing to specify their extensions.
```
```suggestion
At the limit, implementers might not even be willing to document their extensions.
```

index.bs
```diff
@@ -3708,6 +3708,137 @@ and makes it clear when the state described by the flags is reset.
...
+
+In the case of a registry that doesn't require specifications,
+it can be tempting to identify extensions with URLs or URIs instead of registered strings.
+This has the effect of defining a [[rfc8126#section-4.3|hierarchical registration policy]]
+and making it very easy to extend the feature,
+by just picking a URL from a domain that the extender controls.
+This clearly loses the interoperability benefits of requiring a specification,
+and in the case of DNS-based URLs,
+it also risks that the entity that defined the extension may lose control of its domain.
+URIs are appropriate for a few kinds of very-low-coordination extension,
```
martinthomson, line 3783:
```suggestion
URIs are appropriate for a few kinds of very-low-coordination extension,
such as the definition of an XML namespace [[XML-NAMES]],
```

index.bs
```diff
@@ -3708,6 +3708,137 @@ and makes it clear when the state described by the flags is reset.
...
+but most of the time a WG-managed permissive registry table will work better.
+
+Because an [=extension point=] defines an interface,
+and it's difficult to be confident in an interface definition
+before that interface has several implementations,
+any new registry should start with at least 2-3 entries defined.
```
```suggestion
any new registry should start with more than one entry defined.
```

index.bs
```diff
@@ -3708,6 +3708,137 @@ and makes it clear when the state described by the flags is reset.
...
+Each of these initial registry entries
+can be either required or optional for implementations to recognize.
+For extension points that can't just be ignored when their extension isn't recognized,
+the registry should include at least 1 required entry.
```
```suggestion
Specifications that include extension points for mandatory features,
such as a core function that can have an alternative implementation
that is selected by identifying the extension,
must include at least one entry that all implementations are required to support.
```

I think that you are referring here to DC and the protocol choice.  But this also applies to TLS key exchange methods or cipher suites.  You don't have a functional specification if you don't at least one thing that everyone supports.

index.bs
```diff
@@ -3708,6 +3708,137 @@ and makes it clear when the state described by the flags is reset.
...
+
+<div class="example" heading="Link Relations" id="example-link-relation-registry">
+
+Link relations are currently registered in a 3-part registry.
+[[rfc8288#procedure|RFC 8288]] defines the [[IANA-RELATIONS inline]] registry
```
martinthomson, line 3830: Interesting.  The RFC 8288 link goes to the actual section.  The same is not true for https://datatracker.ietf.org/doc/html/rfc8288#procedure though that is where I would expect many RFC links to use.  (Failing that: https://www.rfc-editor.org/rfc/rfc8288.html#procedure)

index.bs
```diff
@@ -3708,6 +3708,137 @@ and makes it clear when the state described by the flags is reset.
...
+[[MFREL inline]] defines a second registry
+to which anyone with a wiki account can add.
+This registry has a column for specifications,
+but nobody enforces that the column is filled in
+or that the specifications meet any particular quality requirements.
+[[html#linkTypes]] then repeats a subset of these relation types
+to define how they work in web browsers.
+
+In an ideal world, [[RFC8288]] and [[MFREL inline]] would probably be a single registry,
+with [[html#linkTypes|HTML]] defining browser behavior.
```
martinthomson, line 3841: I sort of agree with this statement, though this entire example makes me a little uncomfortable.  It is not exactly a simple and straightforward example that illustrates the principles on hand.  Instead, it's a complicated situation that has evolved over time, mostly as an object lesson in how NOT to do registries.

So, while I appreciate the history lesson, I'm not sure that we're doing our readers the favour that they came here for.  We're unloading a lot of nuance on them instead.

Do we have a simple, relatively old, straightforward registry that we can point to?  I can point to a lot of IETF work that fits that neatly, but I'm less able to point to a good example from the W3C.  That might say something, of course.

index.bs
jyasskin: I wound up merging and shortening it.

index.bs
jyasskin: I'm happy to start this without the "However". But do you think registries are obscure enough that we need to define them here, instead of leaving the definition behind the two links?

index.bs
jyasskin: Sure.

index.bs
jyasskin: Yeah, for some reason, [specref](https://www.specref.org/?q=rfc8288) and the webref crawler have RFC8288 under httpwg.org. That site defines different heading IDs from rfc-editor.org or the Datatracker. Because webref knows of some heading IDs, this linking syntax is requiring that I use an ID that it understands. I could switch to a plain markdown link since I have another [[reference]] elsewhere in the document, but I feel like it's not particularly bad to have the link go to httpwg.org.

index.bs
jyasskin: I think it might be. We start by saying that "the original standards track" is often the right way to handle extensions, and we end by saying that URLs are sometimes acceptable when it's important not to constrain extensibility, but the main point of the section is still "use registries".

index.bs
jyasskin: I tend to think that "appropriate" hides the advice, but here's an attempt to be more precise.

index.bs
jyasskin: Tweaked and done.

index.bs
jyasskin: I'm thinking more of some arguments I had with the DID and VC communities about some empty registries, but TLS key-exchange and cipher suites are also important. I rewrote this somewhat differently to try to capture the requirement better.

index.bs
jyasskin: I added the Well-Known registry, which overlaps with Web protocols fairly often and has a less-complicated history than Media Types.

I also added a note that sometimes registries get complicated, and that's still ok.

index.bs
martinthomson: Would "Provide support for future extensions to features" work better?  It doesn't include the word "registry", but I think it matches the text better.

index.bs
```diff
@@ -3708,6 +3708,165 @@ and makes it clear when the state described by the flags is reset.
...
+and it's difficult to be confident in an interface definition
+before that interface has several implementations,
+any new registry should start with more than one entry defined.
+
+Consider the case of two implementations
+that implement non-overlapping subsets
+of the defined extensions for a given extension point.
+If the extension point cannot simply be treated as empty in that case,
+you must ensure that implementations always overlap
+by defining the registry to always include at least 1 required entry.
```
martinthomson: This is still a bit subtle.


```suggestion
If operating the feature requires that implementations
choose an extension that they both support,
implementations that support non-overlapping subsets cannot interoperate.
To avoid such failures, the registry must include at least 1 entry
that all implementations are required to support.
```

index.bs
```diff
@@ -3708,6 +3708,165 @@ and makes it clear when the state described by the flags is reset.
...
+
+The [Well-Known URI registry](https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml),
+defined by [[RFC8615]],
+ensures that protocols don't collide when allocating a path in all web servers.
+
+Part of the purpose of HTTP and similar URL schemes
+is to allow new protocols to be defined on top of them,
+without requiring an update to the core HTTP specification for each new protocol.
+When these protocols need to retrieve information about an entire origin,
+the only effect on HTTP as a whole would come from potential name collisions.
```
martinthomson: If you want to be truly pedantic content negotiation could save people from name collisions.  Most well-known resources have a defined media type.

index.bs
```diff
@@ -3708,6 +3708,165 @@ and makes it clear when the state described by the flags is reset.
...
+Since the registry itself mitigates that risk,
+there isn't a need for the whole HTTP community
+to agree on a new HTTP revision for each registration.
```
martinthomson, line 3814: Some HTTP people don't like it when protocols are layered on top of HTTP.

You might get a better reception with:

```suggestion
HTTP does not constrain how people define resources on their servers,
but well-known allows specifications to define functions or information
that all servers can publish in a common format.
Specifications other than HTTP itself can define well-known resources
that every HTTP server can make available,
with the only challenge being avoiding a name collision.
A registry avoids name conflicts
and ensures that new resources can be defined
without revising the core HTTP specification.
```

index.bs
```diff
@@ -3708,6 +3708,165 @@ and makes it clear when the state described by the flags is reset.
...
+tend to be defined by an agreement between the participating origins,
+which is documented in a specification.
+
+</div>
+
+<div class="example" heading="Link Relations" id="example-link-relation-registry">
+
+Registries can get complicated,
+especially when implementers work around mistakes in the registration requirements.
+They're still usually better than providing no place to list extensions.
```
martinthomson, line 3825: Is this exemplary text, or could it go ahead of the example?

martinthomson (approved):
The approval stands.  The well-known text activated a few antibodies, but it's a good example.

(FWIW, I think `.well-known` is abused more than it is used correctly, but I am still learning to let go.)

index.bs
jyasskin: I'd like to try to include a clue to the guidance we're giving, so folks don't see the heading and immediately have to ask "how?", or worse, assume that their semi-specified extension scheme does what we're suggesting. Our two suggestions are roughly "if possible, only support future extensions by updating the main spec" and "otherwise use a registry" ... so maybe ...
```suggestion
<h3 id="registries">Plan for future extensibility through main spec updates if possible; otherwise use a registry</h3>
```

index.bs
jyasskin: It can go ahead of the example. Done.

index.bs
jyasskin: This is true, but content negotiation itself is contentious, and using it here just pushes the burden to the Media Types registry. (I think it usually hasn't worked out. Maybe that's because the `Accept*` headers aren't sophisticated enough to capture client preferences, but if server-driven negotiation were the right way to do things, folks would probably have developed a better `Accept` system by now.)

I think you don't feel strongly, but I'll point this out in Slack to see if the rest of the group leans one way or the other.

index.bs
jyasskin: I feel like those HTTP people are in the rough, given the prevalence of REST, JSON-RPC, and .well-known itself. If folks on the TAG are skeptical, or if the original is incorrect in some way, I can take this rewording (although I think I see a way to simplify it a bit), but I don't want to be less direct just because we might get criticized for the stronger stance.

index.bs
martinthomson: WFM

index.bs
martinthomson: I very much don't feel strongly.  It was just a thought that maybe you hadn't considered.

index.bs
martinthomson: A lighter edit then, as an alternative:

```suggestion
Part of the purpose of HTTP and similar URL schemes
is to allow new capabilities to be defined that use the core protocol,
without requiring an update to the core specification.
When information is needed about an entire origin,
the only need for coordination as a whole comes from avoiding the possibility of name collisions.
Since the registry itself mitigates that risk,
there isn't a need for the whole HTTP community
to agree on a new HTTP revision for each registration.
```

PR #597, opened 2025-09-22
Handle non-fully-active documents

marcoscaceres:
This pull request adds new guidance for handling cases where documents or execution contexts are no longer fully active, such as when an iframe is removed or navigated away. The update clarifies how specifications should address the lifecycle and cleanup of asynchronous operations and UI elements associated with destroyed or inactive contexts.

**Guidance for handling destroyed contexts and non-fully-active documents:**

* Added a new section (`#handle-destroyed-contexts`) recommending that specs explicitly define behavior when the associated document is not fully active at call time (e.g., throw or reject with `InvalidStateError`).
* Specified that ongoing async operations should be aborted and rejected with `AbortError` if the context becomes non-fully-active or the execution context is destroyed, and that any UI should be dismissed in such cases.
* Advised that queued tasks should check if the document is still fully active before proceeding, and to avoid relying on garbage collection for cleanup semantics.
* Included cross-references to related guidance for the BFCache and detached iframes.

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/597.html#handle-non-fully-active-documents" title="Last updated on Jan 27, 2026, 4:16 AM UTC (3ed3057)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/597/6ec7e11...3ed3057.html" title="Last updated on Jan 27, 2026, 4:16 AM UTC (3ed3057)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -929,6 +929,44 @@ it is not always *sufficient*
 to protect users from invasive behaviours.
 [Designing for user intent](#user-intent) and [helping users to make good decisions](#user-decisions) is also important.
 
+<h3 id="handle-destroyed-contexts">Handle non-fully-active documents (and destroyed execution contexts)</h3>
+
+When an <{iframe}> (or similar) is removed or navigated,
+its document stops being [=Document/fully active=]
+and its [=ECMAScript/execution contexts=] can get destroyed.
+However, references to objects from that document
+can remain reachable from other documents
+or from other realms.
+
+Specifications should say what happens both at call time
+and after an operation has begun:
+otherwise background work may keep running after a page is gone,
+UI can hang, and user agents diverge in abort/cleanup behavior.
+
+If the [=relevant global object=]'s [=associated Document=]
+is not [=Document/fully active=] at call time,
+then [=throw=], or return [=a promise rejected with=],
+with {{InvalidStateError}}.
+
+If, after starting an asynchronous operation,
+that [=associated Document=] becomes non-[=Document/fully active=]
+or the [=relevant global object=]'s [=ECMAScript/execution context=] is destroyed,
+abort the operation and reject with {{AbortError}}.
+If the operation presents UI (e.g., a picker, chooser, or prompt),
+dismiss it when the context becomes non-[=Document/fully active=]
+and settle with {{AbortError}}.
+
+When [=queue a task|queueing=] subsequent work
+(e.g., for an event, settling a promise, a time-based operation),
+include an early exit:
+if the [=associated Document=] is not [=Document/fully active=],
+gracefully terminate the [=task=].
+Do not rely on garbage collection to define these semantics.
+
+See also: [[#support-non-fully-active]] for the BFCache case;
+most of the same principles apply to detached iframes
+and other torn-down [=All Execution Contexts/realms=].
+
 <h3 id="support-non-fully-active">Support non-fully active BFCached documents</h3>
 
 Specify how your feature behaves
--- a/index.bs
+++ b/index.bs
@@ -996,6 +996,7 @@ even though BFCache may suspend a document
 rather than tearing it down.
 
 <h3 id="support-non-fully-active">Support non-fully active BFCached documents</h3>
+<h3 id="support-non-fully-active">Support non-fully active BFCached documents</h3>
 
 Specify how your feature behaves
 in non-[=Document/fully active=] BFCached (Back/forward cached) documents if possible.
--- a/index.bs
+++ b/index.bs
@@ -933,10 +933,9 @@ to protect users from invasive behaviours.
 
 Make sure that your API behaves predictably
 when its [=relevant global object=]'s [=associated Document=]
-is no longer [=Document/fully active=]:
-failing fast if the document is already non-[=Document/fully active=]
-and aborting cleanly if it becomes non-[=Document/fully active=]
-mid-operation.
+is no longer [=Document/fully active=].
+Fail fast when the document is already non-[=Document/fully active=]
+and abort cleanly if it becomes non-[=Document/fully active=] mid-operation.
 
 When an <{iframe}> (or similar) is removed or navigated,
 its [=associated Document=] can stop being [=Document/fully active=].
@@ -950,9 +949,10 @@ This situation can arise when browser-mediated UI
 (e.g., authentication, payment, or credential presentation)
 is shown on behalf of an <{iframe}>
 and the <{iframe}> is removed or navigated
-while a user is interacting with it.
+while a user is interacting with that interface.
 
-Specifications should define behavior
+Specifications should define how UI actions
+interact with non-[=Document/fully active=] documents
 both at call time
 and after an operation has begun.
 Doing so helps ensure that background work stops
@@ -984,7 +984,7 @@ settle a promise, or perform time-based processing),
 specifications should include an early-exit check
 at the point the [=task=] runs:
 if the [=associated Document=] is not [=Document/fully active=],
-gracefully terminate the [=task=]
+then gracefully terminate the [=task=]
 without further observable effects.
 Specifications must not rely on garbage collection
 to define these semantics.
--- a/index.bs
+++ b/index.bs
@@ -962,21 +962,21 @@ and that user agents behave consistently.
 
 As general rules:
 
-* If, at call time,
-the [=relevant global object=]'s [=associated Document=]
-is not [=Document/fully active=],
-then the operation must fail immediately:
-either [=throw=] a {{InvalidStateError}} {{DOMException}},
-or return [=a promise rejected with=] {{InvalidStateError}} {{DOMException}}.
+  * If, at call time,
+    the [=relevant global object=]'s [=associated Document=]
+    is not [=Document/fully active=],
+    then the operation must fail immediately:
+    either [=throw=] a {{InvalidStateError}} {{DOMException}},
+    or return [=a promise rejected with=] {{InvalidStateError}} {{DOMException}}.
 
-* If, after starting an asynchronous operation,
-the [=associated Document=] becomes non-[=Document/fully active=],
-the operation must be aborted
-and failure signaled with {{AbortError}}.
-If the operation presents user-facing UI
-(e.g., a picker, chooser, or prompt),
-dismiss it when the document becomes non-[=Document/fully active=]
-and settle with {{AbortError}}.
+  * If, after starting an asynchronous operation,
+    the [=associated Document=] becomes non-[=Document/fully active=],
+    the operation must be aborted
+    and failure signaled with {{AbortError}}.
+    If the operation presents user-facing UI
+    (e.g., a picker, chooser, or prompt),
+    dismiss it when the document becomes non-[=Document/fully active=]
+    and settle with {{AbortError}}.
 
 When [=queue a task|queueing=] subsequent work
 (for example, to dispatch an event,
--- a/index.bs
+++ b/index.bs
@@ -929,43 +929,71 @@ it is not always *sufficient*
 to protect users from invasive behaviours.
 [Designing for user intent](#user-intent) and [helping users to make good decisions](#user-decisions) is also important.
 
-<h3 id="handle-destroyed-contexts">Handle non-fully-active documents (and destroyed execution contexts)</h3>
+<h3 id="handle-non-fully-active-documents">Handle non-fully-active documents</h3>
+
+Make sure that your API behaves predictably
+when its [=relevant global object=]'s [=associated Document=]
+is no longer [=Document/fully active=]:
+failing fast if the document is already non-[=Document/fully active=]
+and aborting cleanly if it becomes non-[=Document/fully active=]
+mid-operation.
 
 When an <{iframe}> (or similar) is removed or navigated,
-its document stops being [=Document/fully active=]
-and its [=ECMAScript/execution contexts=] can get destroyed.
+its [=associated Document=] can stop being [=Document/fully active=].
+In many cases, user agents will tear down the document’s JavaScript
+[=ECMAScript/execution contexts=] rather than keeping them suspended.
 However, references to objects from that document
 can remain reachable from other documents
-or from other realms.
+or from other [=ECMAScript/realms=].
 
-Specifications should say what happens both at call time
-and after an operation has begun:
-otherwise background work may keep running after a page is gone,
-UI can hang, and user agents diverge in abort/cleanup behavior.
+This situation can arise when browser-mediated UI
+(e.g., authentication, payment, or credential presentation)
+is shown on behalf of an <{iframe}>
+and the <{iframe}> is removed or navigated
+while a user is interacting with it.
 
-If the [=relevant global object=]'s [=associated Document=]
-is not [=Document/fully active=] at call time,
-then [=throw=], or return [=a promise rejected with=],
-with {{InvalidStateError}}.
+Specifications should define behavior
+both at call time
+and after an operation has begun.
+Doing so helps ensure that background work stops
+on non-[=Document/fully active=] documents,
+that user-facing UI is gracefully dismissed,
+and that user agents behave consistently.
 
-If, after starting an asynchronous operation,
-that [=associated Document=] becomes non-[=Document/fully active=]
-or the [=relevant global object=]'s [=ECMAScript/execution context=] is destroyed,
-abort the operation and reject with {{AbortError}}.
-If the operation presents UI (e.g., a picker, chooser, or prompt),
-dismiss it when the context becomes non-[=Document/fully active=]
+As general rules:
+
+* If, at call time,
+the [=relevant global object=]'s [=associated Document=]
+is not [=Document/fully active=],
+then the operation must fail immediately:
+either [=throw=] a {{InvalidStateError}} {{DOMException}},
+or return [=a promise rejected with=] {{InvalidStateError}} {{DOMException}}.
+
+* If, after starting an asynchronous operation,
+the [=associated Document=] becomes non-[=Document/fully active=],
+the operation must be aborted
+and failure signaled with {{AbortError}}.
+If the operation presents user-facing UI
+(e.g., a picker, chooser, or prompt),
+dismiss it when the document becomes non-[=Document/fully active=]
 and settle with {{AbortError}}.
 
 When [=queue a task|queueing=] subsequent work
-(e.g., for an event, settling a promise, a time-based operation),
-include an early exit:
+(for example, to dispatch an event,
+settle a promise, or perform time-based processing),
+specifications should include an early-exit check
+at the point the [=task=] runs:
 if the [=associated Document=] is not [=Document/fully active=],
-gracefully terminate the [=task=].
-Do not rely on garbage collection to define these semantics.
+gracefully terminate the [=task=]
+without further observable effects.
+Specifications must not rely on garbage collection
+to define these semantics.
 
-See also: [[#support-non-fully-active]] for the BFCache case;
-most of the same principles apply to detached iframes
-and other torn-down [=All Execution Contexts/realms=].
+See also: [[#support-non-fully-active]] for the BFCache case.
+Many of the same principles apply to detached iframes
+and other documents that have become non-[=Document/fully active=],
+even though BFCache may suspend a document
+rather than tearing it down.
 
 <h3 id="support-non-fully-active">Support non-fully active BFCached documents</h3>
 
--- a/index.bs
+++ b/index.bs
@@ -996,7 +996,6 @@ even though BFCache may suspend a document
 rather than tearing it down.
 
 <h3 id="support-non-fully-active">Support non-fully active BFCached documents</h3>
-<h3 id="support-non-fully-active">Support non-fully active BFCached documents</h3>
 
 Specify how your feature behaves
 in non-[=Document/fully active=] BFCached (Back/forward cached) documents if possible.
--- a/index.bs
+++ b/index.bs
@@ -943,7 +943,7 @@ In many cases, user agents will tear down the document’s JavaScript
 [=ECMAScript/execution contexts=] rather than keeping them suspended.
 However, references to objects from that document
 can remain reachable from other documents
-or from other [=ECMAScript/realms=].
+or from other parts of the user agent.
 
 This situation can arise when browser-mediated UI
 (e.g., authentication, payment, or credential presentation)
@@ -951,7 +951,7 @@ is shown on behalf of an <{iframe}>
 and the <{iframe}> is removed or navigated
 while a user is interacting with that interface.
 
-Specifications should define how UI actions
+Specifications must define how calls and UI actions
 interact with non-[=Document/fully active=] documents
 both at call time
 and after an operation has begun.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -866,6 +866,44 @@ it is not always *sufficient*
...
+or from other realms.
+
+Specifications should say what happens both at call time
+and after an operation has begun:
+otherwise background work may keep running after a page is gone,
+UI can hang, and user agents diverge in abort/cleanup behavior.
+
+If the [=relevant global object=]'s [=associated Document=]
+is not [=Document/fully active=] at call time,
+then [=throw=], or return [=a promise rejected with=],
```
```suggestion
then [=throw=] or return [=a promise rejected with=]
```

index.bs
```diff
@@ -866,6 +866,44 @@ it is not always *sufficient*
...
+with {{InvalidStateError}}.
+
+If, after starting an asynchronous operation,
+that [=associated Document=] becomes non-[=Document/fully active=]
+or the [=relevant global object=]'s [=ECMAScript/execution context=] is destroyed,
```
```suggestion
or the [=relevant global object=]'s [=ECMAScript/execution context=] is destroyed by script,
```

index.bs
```diff
@@ -866,6 +866,44 @@ it is not always *sufficient*
 to protect users from invasive behaviours,
 and seeking [meaningful consent](#consent) is also important.
 
+<h3 id="handle-destroyed-contexts">Handle non-fully-active documents (and destroyed execution contexts)</h3>
```
jyasskin: Everywhere else we use "non-fully active" ... which is kinda wrong, but allows [=fully active=] to work. We should be consistent about the hyphenation across this whole document.

index.bs
```diff
@@ -866,6 +866,44 @@ it is not always *sufficient*
...
+When [=queue a task|queueing=] subsequent work
+(e.g., for an event, settling a promise, a time-based operation),
+include an early exit:
+if the [=associated Document=] is not [=Document/fully active=],
+gracefully terminate the [=task=].
+Do not rely on garbage collection to define these semantics.
+
+See also: [[#support-non-fully-active]] for the BFCache case;
+most of the same principles apply to detached iframes
+and other torn-down [=All Execution Contexts/realms=].
```
jyasskin: I feel like these sections need to be integrated. Is there a set of instructions that works for all non-fully-active documents, with small extensions that could live in subsections for the bfcache and iframe cases?

index.bs
```diff
@@ -866,6 +866,44 @@ it is not always *sufficient*
...
 and seeking [meaningful consent](#consent) is also important.
 
+<h3 id="handle-destroyed-contexts">Handle non-fully-active documents (and destroyed execution contexts)</h3>
+
+When an <{iframe}> (or similar) is removed or navigated,
+its document stops being [=Document/fully active=]
+and its [=ECMAScript/execution contexts=] can get destroyed.
+However, references to objects from that document
+can remain reachable from other documents
+or from other realms.
```
jyasskin: As in other PRs, [sections should start with the main, positive, advice](https://github.com/w3ctag/process/blob/main/style-guide.md#get-to-the-point). Maybe:

> Avoid doing work, and usually cancel ongoing work, on non-[=fully active=] documents and destroyed execution contexts.

(With a link to the specification terms that should be used to detect a destroyed execution context.)

index.bs
```diff
@@ -866,6 +866,44 @@ it is not always *sufficient*
...
+UI can hang, and user agents diverge in abort/cleanup behavior.
+
+If the [=relevant global object=]'s [=associated Document=]
+is not [=Document/fully active=] at call time,
+then [=throw=], or return [=a promise rejected with=],
+with {{InvalidStateError}}.
+
+If, after starting an asynchronous operation,
+that [=associated Document=] becomes non-[=Document/fully active=]
+or the [=relevant global object=]'s [=ECMAScript/execution context=] is destroyed by script,
```
jyasskin: This should link to the algorithm to patch for every asynchronous operation.

index.bs
```diff
@@ -866,6 +866,44 @@ it is not always *sufficient*
...
+abort the operation and reject with {{AbortError}}.
```
jyasskin: This doesn't seem to always be the right advice. E.g. https://html.spec.whatwg.org/multipage/images.html#update-the-image-data just waits for the document to become fully active again.

index.bs
```diff
@@ -866,6 +866,44 @@ it is not always *sufficient*
...
+If the operation presents UI (e.g., a picker, chooser, or prompt),
+dismiss it when the context becomes non-[=Document/fully active=]
+and settle with {{AbortError}}.
+
+When [=queue a task|queueing=] subsequent work
+(e.g., for an event, settling a promise, a time-based operation),
+include an early exit:
+if the [=associated Document=] is not [=Document/fully active=],
+gracefully terminate the [=task=].
```
jyasskin: This paragraph seems to be saying how to accomplish the previous paragraph. Do folks struggle enough with figuring out how to abort an operation that we need to spend words on it? Is this even always the right strategy?

index.bs
```diff
@@ -866,6 +866,44 @@ it is not always *sufficient*
 to protect users from invasive behaviours,
 and seeking [meaningful consent](#consent) is also important.
 
+<h3 id="handle-destroyed-contexts">Handle non-fully-active documents (and destroyed execution contexts)</h3>
+
+When an <{iframe}> (or similar) is removed or navigated,
+its document stops being [=Document/fully active=]
+and its [=ECMAScript/execution contexts=] can get destroyed.
```
jyasskin: Is a "destroyed" execution context a concept that can appear in specifications? If not, we should only mention the spec-visible conditions that cause this to happen in implementations.

jyasskin (changes requested):

index.bs
jyasskin: Also please try not to use suggested changes to edit your own PRs. They're pretty noisy to everyone else subscribed to the CL. If you can't edit locally, there's always https://github.com/w3ctag/design-principles/edit/more_doc_state/index.bs?pr=%2Fw3ctag%2Fdesign-principles%2Fpull%2F597.

marcoscaceres: Yeah, that's equally as annoying tho: because it means I need to spending additional time replicating what is suggested there locally. 

Best I can do is group the commits, but not being able to make changes in this UI would be disruptive to my workflow.

index.bs
```suggestion
<h3 id="handle-destroyed-contexts">Handle non-fully active documents (and destroyed execution contexts)</h3>
```

index.bs
marcoscaceres: I might say "As a general rule, ..." and maybe mention that special case. 

The intent here is that if the developer destroys the execution context (e..g, `iframe.remove()`), they generally intended (or should be interpreted by the UA) as them wanting to abort all ongoing operations. 

The case you are talking about is different: when it's a user initiated action that forces a document into the BFCache, and hence the document becomes non-fully active. In that case, it makes sense to continue/resume loading things like images or other media (i.e., user needs VS developer wants).

index.bs
marcoscaceres: Right, but this is a special case... document cannot be made non-fully active by script in way that is recoverable to make them fully active again (i.e., AFAIK, only user action can put a document into a recoverable state, but never via script).

index.bs
marcoscaceres: My experience is that yes, and also because there are the tricky edge cases, even implementers might have not thought of these cases. These are where you get weird situations around unspecified behavior.

index.bs
marcoscaceres: Yeah, it's basically:

1. Removing an iframe. 
2. Navigating an iframe.

Both destroy the script execution context, causing all the destructors to run... and some of those would resolve pending promises with AbortError, and tear down UIs (e.g., payment sheet, a credential chooser, and so on).

marcoscaceres: I wonder if we should instead say "its [=ECMAScript/execution contexts=] can get garbage collected". 

"garbage collected" is not exported or defined in ECMAScript, but at least it makes the language consistent. 

```suggestion
and its [=ECMAScript/execution contexts=] gets garbage collected.
```

martinthomson: @jyasskin, WDYT?

index.bs
```diff
@@ -868,6 +866,44 @@ it is not always *sufficient*
 to protect users from invasive behaviours,
 and seeking [meaningful consent](#consent) is also important.
 
+<h3 id="handle-destroyed-contexts">Handle non-fully active documents (and destroyed execution contexts)</h3>
+
+When an <{iframe}> (or similar) is removed or navigated,
+its document stops being [=Document/fully active=]
```
```suggestion
its [=document=] stops being [=Document/fully active=]
```

index.bs
```diff
@@ -868,6 +866,44 @@ it is not always *sufficient*
...
+and its [=ECMAScript/execution contexts=] can get destroyed.
+However, references to objects from that document
```
marcoscaceres, line 944:
```suggestion
However, references to javascript objects from that document
```

index.bs
```diff
@@ -868,6 +866,44 @@ it is not always *sufficient*
 to protect users from invasive behaviours,
 and seeking [meaningful consent](#consent) is also important.
 
+<h3 id="handle-destroyed-contexts">Handle non-fully active documents (and destroyed execution contexts)</h3>
+
```
```suggestion
<h3 id="handle-destroyed-contexts">Handle non-fully active documents (and contexts that have been garbage collected)</h3>

Always handle what happens when a [=document=] becomes [=connected/disconnected=] and garbage collection can leave an API in potentially invalid state.

```

index.bs
marcoscaceres: Added suggestion above.

index.bs
```diff
@@ -868,6 +866,44 @@ it is not always *sufficient*
...
+and after an operation has begun:
+otherwise background work may keep running after a page is gone,
+UI can hang, and user agents diverge in abort/cleanup behavior.
+
+If the [=relevant global object=]'s [=associated Document=]
+is not [=Document/fully active=] at call time,
+then [=throw=] or return [=a promise rejected with=]
+with {{InvalidStateError}}.
+
+If, after starting an asynchronous operation,
```
```suggestion
As a general rule, if, after starting an asynchronous operation,
```

index.bs
```diff
@@ -868,6 +866,44 @@ it is not always *sufficient*
...
+that [=associated Document=] becomes non-[=Document/fully active=]
+or the [=relevant global object=]'s [=ECMAScript/execution context=] is destroyed by script,
+abort the operation and reject with {{AbortError}}.
+If the operation presents UI (e.g., a picker, chooser, or prompt),
+dismiss it when the context becomes non-[=Document/fully active=]
+and settle with {{AbortError}}.
```
```suggestion
and settle any pending promises with an {{AbortError}}.
```

index.bs
marcoscaceres: I hold that they are different things. The BFCache is not observable (or is not supposed to be observable by script).... what we are describing here is observable and actually caused by scripts doing "fun things" (i.e., holding references to things they probably shouldn't be holding onto).

index.bs
jyasskin: "Garbage collected" doesn't seem quite right for a thing that can still have active references from JS. @annevk, is there a defined term for this kind of document/context?

jyasskin (commented):
I do still think that the two non-fully-active sections need to be unified in some way. It seems like we have some things that should be true of all non-fully-active documents, and then some things for bfcached documents and some other things for "released" documents (by whatever name).

I checked with @fergald, who said that in Chromium, it's true that there can't be JS references into bfcached documents, but he was surprised that someone from Safari would agree with that statement.

rakina:
> I do still think that the two non-fully-active sections need to be unified in some way. It seems like we have some things that should be true of all non-fully-active documents, and then some things for bfcached documents and some other things for "released" documents (by whatever name).

I agree with this, or at least make the "Support non-fully active BFCached documents" and "Handle non-fully active documents (and destroyed execution contexts) " explicitly mutually exclusive.  There is already some mention of other non-fully active cases not being covered in the former. Maybe these should be combined and point to the separate BFCache doc for the BFCache case, and the new guidance for the rest. As it reads right now it's unclear if all the suggestions in the new section should apply to BFCached non-fully active case too, especially things like

> "If, after starting an asynchronous operation, that [ associated Document ](https://html.spec.whatwg.org/multipage/nav-history-apis.html#concept-document-window) becomes non- [ fully active ](https://html.spec.whatwg.org/multipage/document-sequences.html#fully-active) or the [ relevant global object ](https://html.spec.whatwg.org/multipage/webappapis.html#concept-relevant-global) ’s [ execution context ](https://tc39.es/ecma262/multipage/executable-code-and-execution-contexts.html#sec-execution-contexts) is destroyed by script, abort the operation and reject"

For the BFCache case, I think there are cases where we don't want to abort and just continue after restore.

fergald:
> For the BFCache case, I think there are cases where we don't want to abort and just continue after restore.

Absolutely. There would be significant breakage of pages if BFCache restores involved aborting all ongoing async operations.

index.bs
```diff
@@ -868,6 +866,44 @@ it is not always *sufficient*
 to protect users from invasive behaviours,
 and seeking [meaningful consent](#consent) is also important.
 
+<h3 id="handle-destroyed-contexts">Handle non-fully active documents (and destroyed execution contexts)</h3>
```
```suggestion
<h3 id="handle-destroyed-contexts">Handle non-fully active documents</h3>
```

index.bs
```diff
@@ -925,8 +865,74 @@ once per API call ([transient consuming](https://html.spec.whatwg.org/#activatio
...
+and seeking [meaningful consent](#consent) is also important.
+
+<h3 id="handle-non-fully-active-documents">Handle non-fully-active documents</h3>
+
+Make sure that your API behaves predictably
+when its [=relevant global object=]'s [=associated Document=]
+is no longer [=Document/fully active=]:
+failing fast if the document is already non-[=Document/fully active=]
+and aborting cleanly if it becomes non-[=Document/fully active=]
+mid-operation.
```
```suggestion
is no longer [=Document/fully active=].
Fail fast when the document is already non-[=Document/fully active=]
and abort cleanly if it becomes non-[=Document/fully active=] mid-operation.
```

index.bs
```diff
@@ -925,8 +865,74 @@ once per API call ([transient consuming](https://html.spec.whatwg.org/#activatio
...
+[=ECMAScript/execution contexts=] rather than keeping them suspended.
+However, references to objects from that document
+can remain reachable from other documents
+or from other [=ECMAScript/realms=].
+
+This situation can arise when browser-mediated UI
+(e.g., authentication, payment, or credential presentation)
+is shown on behalf of an <{iframe}>
+and the <{iframe}> is removed or navigated
+while a user is interacting with it.
```
```suggestion
while a user is interacting with that interface.
```

index.bs
```diff
@@ -925,8 +865,74 @@ once per API call ([transient consuming](https://html.spec.whatwg.org/#activatio
...
+
+Specifications should define behavior
```
```suggestion
Specifications should define how UI actions
interact with non-[=Document/fully active=] documents
```

index.bs
```diff
@@ -925,8 +865,74 @@ once per API call ([transient consuming](https://html.spec.whatwg.org/#activatio
...
+    dismiss it when the document becomes non-[=Document/fully active=]
+    and settle with {{AbortError}}.
+
+When [=queue a task|queueing=] subsequent work
+(for example, to dispatch an event,
+settle a promise, or perform time-based processing),
+specifications should include an early-exit check
+at the point the [=task=] runs:
+if the [=associated Document=] is not [=Document/fully active=],
+gracefully terminate the [=task=]
```
```suggestion
then gracefully terminate the [=task=]
```

martinthomson (commented):
Some editorial fixes only.  Nothing much of substance; it reads well to me.

marcoscaceres:
@fergald, @rakina, I tried to make the distinction between the two sections more clear (and how they relate). Can you take another look?

index.bs
```diff
@@ -928,6 +928,72 @@ it is not always *sufficient*
...
+Specifications should define how UI actions
+interact with non-[=Document/fully active=] documents
+both at call time
+and after an operation has begun.
+Doing so helps ensure that background work stops
+on non-[=Document/fully active=] documents,
+that user-facing UI is gracefully dismissed,
+and that user agents behave consistently.
+
+As general rules:
```
rakina, line 963: I think it would be a bit more clear to explicitly mention that this is for the destroyed / detached iframe case (maybe in the section title too?). Also is there a way to differentiate if we're in the destroyed fully active case vs BFCached fully active case?

marcoscaceres: The differentiation happens because the developer/script has explicitly removed the frame or changed the frame's `src`... so I guess the differentiation is "the developer did x" vs "the user navigated back/forward".

index.bs
```diff
@@ -928,6 +928,72 @@ it is not always *sufficient*
...
+without further observable effects.
+Specifications must not rely on garbage collection
+to define these semantics.
+
+See also: [[#support-non-fully-active]] for the BFCache case.
+Many of the same principles apply to detached iframes
+and other documents that have become non-[=Document/fully active=],
+even though BFCache may suspend a document
+rather than tearing it down.
+
```
marcoscaceres, line 996:
```suggestion

<h3 id="support-non-fully-active">Support non-fully active BFCached documents</h3>
```

index.bs
```diff
@@ -928,6 +928,73 @@ it is not always *sufficient*
...
+<h3 id="support-non-fully-active">Support non-fully active BFCached documents</h3>
```
```suggestion
```

index.bs
```diff
@@ -928,6 +928,73 @@ it is not always *sufficient*
...
+Specifications should define how UI actions
+interact with non-[=Document/fully active=] documents
+both at call time
+and after an operation has begun.
+Doing so helps ensure that background work stops
+on non-[=Document/fully active=] documents,
+that user-facing UI is gracefully dismissed,
+and that user agents behave consistently.
+
+As general rules:
```
marcoscaceres, line 963: @rakina, how about: 

```suggestion
The following guidance applies to cases where a document
becomes non-[=Document/fully active=] because it is being detached,
navigated away from, or otherwise destroyed by the developer,
and not to cases where a document is intentionally kept suspended
by the user agent (e.g., via BFCache):
```

index.bs
```diff
@@ -928,6 +928,72 @@ it is not always *sufficient*
...
+Fail fast when the document is already non-[=Document/fully active=]
+and abort cleanly if it becomes non-[=Document/fully active=] mid-operation.
+
+When an <{iframe}> (or similar) is removed or navigated,
+its [=associated Document=] can stop being [=Document/fully active=].
+In many cases, user agents will tear down the document’s JavaScript
+[=ECMAScript/execution contexts=] rather than keeping them suspended.
+However, references to objects from that document
+can remain reachable from other documents
+or from other [=ECMAScript/realms=].
```
jyasskin: This is almost saying the same thing twice (except for the initial document shift mentioned at https://html.spec.whatwg.org/multipage/nav-history-apis.html#concept-document-window). We could remove the "or from other realms" part, but I'm ok keeping it if you think it'll prevent mistakes. You might also mean "or from other parts of a user agent", which connects this directly to the next paragraph.

If so:
```suggestion
can remain reachable from other documents
or from other parts of the user agent.
```

index.bs
```diff
@@ -928,6 +928,72 @@ it is not always *sufficient*
...
+
+This situation can arise when browser-mediated UI
+(e.g., authentication, payment, or credential presentation)
+is shown on behalf of an <{iframe}>
+and the <{iframe}> is removed or navigated
+while a user is interacting with that interface.
+
+Specifications should define how UI actions
```
jyasskin: A nit, but the general rules below use "must", and I think it's appropriate here too.
```suggestion
Specifications must define how UI actions
```

We could also say "need to" if we don't want the normative language.

jyasskin: The general rules also address calls that aren't associated with a UI action, so perhaps:


```suggestion
Specifications must define how calls and UI actions
```

index.bs
```diff
@@ -928,6 +928,72 @@ it is not always *sufficient*
...
+and that user agents behave consistently.
+
+As general rules:
+
+  * If, at call time,
+    the [=relevant global object=]'s [=associated Document=]
+    is not [=Document/fully active=],
+    then the operation must fail immediately:
+    either [=throw=] a {{InvalidStateError}} {{DOMException}},
+    or return [=a promise rejected with=] {{InvalidStateError}} {{DOMException}}.
```
jyasskin: I think we don't need to describe how to fail operations, but it's good to recommend an exception type. How about:

```suggestion
    then the operation must fail immediately.
    We suggest using an {{InvalidStateError}} {{DOMException}}.
```

index.bs
jyasskin: Is it right to say that these two points don't apply to user-driven navigations that might wind up in the bfcache? I feel like they do, and in that case we wouldn't need a change here. 

@rakina, is there a reason not to give this advice for operations on BFCached documents? In Chromium, it sounds like a document can't wind up in the BFCache if it has external references that would allow any of these situations to arise, but in Safari it might, and this advice seems good in that situation.

index.bs
```diff
@@ -928,6 +928,72 @@ it is not always *sufficient*
...
+Specifications must not rely on garbage collection
+to define these semantics.
+
+See also: [[#support-non-fully-active]] for the BFCache case.
+Many of the same principles apply to detached iframes
+and other documents that have become non-[=Document/fully active=],
+even though BFCache may suspend a document
+rather than tearing it down.
+
 <h3 id="support-non-fully-active">Support non-fully active BFCached documents</h3>
```
jyasskin, line 997: This section includes a note that "It is possible for a document to become non-fully active for other reasons ...". We should update that to point back to this section. I'm ok with that being a separate PR.

jyasskin (approved):
A bunch of nits, but this looks generally good. Thanks for bearing with me.

index.bs
marcoscaceres: @rakina, we are going to merge and follow up with any feedback you might have.

index.bs
rakina: Thanks, I think what I'm trying to figure out here is how a feature can spec something that behaves differently for the "detached, so can do destructive things" case vs "might be restored from bfcache later, so don't do destructive things". Maybe there is a "is detached" check on top of a "is fully active" check that they can do?

index.bs
rakina: Thanks and sorry for the delay! Good point on this being possible on Safari, hmm. The first point seems fine since when BFCached it shouldn't be possible to do calls when the global object is frozen due to bfcache (at least in Chromium). I think the second point is tricky:

> If, after starting an asynchronous operation, the [ associated Document ](https://html.spec.whatwg.org/multipage/nav-history-apis.html#concept-document-window) becomes non- [ fully active ](https://html.spec.whatwg.org/multipage/document-sequences.html#fully-active), the operation must be aborted and failure signaled with [ AbortError ](https://webidl.spec.whatwg.org/#aborterror). If the operation presents user-facing UI (e.g., a picker, chooser, or prompt), dismiss it when the document becomes non- [ fully active ](https://html.spec.whatwg.org/multipage/document-sequences.html#fully-active) and settle with  AbortError

We expect pages that have async tasks when going in bfcache to be frozen, but when we restore from bfcache we would expect all of these async tasks to be able to run normally. Like @fergald  [said](https://github.com/w3ctag/design-principles/pull/597#issuecomment-3748594080)  cancelling / aborting them as soon as we get to the non-fully active state will break lots of pages. In the [bfcache guide](https://w3ctag.github.io/bfcache-guide/#listen-fully-active) we recommend to e.g. drop some connections to "outside the document" but not trigger errors / drop things within the document itself.

This is why I think it would be useful to be able to differentiate on the spec text whether we're fully active & detached (so we can do terminal/disruptive things) or we're fully active and just frozen but might be restored later. WDYT?

PR #600, opened 2025-10-03
Avoid abbreviations for strings and enums

### LINKED ISSUE #599: Enums should avoid abbreviations
We should enhance the guidance for enums to state that abbreviations/acronyms should be avoided when possible.

marcoscaceres:
This pull request makes a minor update to the documentation for using strings as constant values. The change clarifies that abbreviations should be avoided, as they can make code harder to read.

Closes #599

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/600.html" title="Last updated on Oct 21, 2025, 4:51 AM UTC (2c16555)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/600/444a329...2c16555.html" title="Last updated on Oct 21, 2025, 4:51 AM UTC (2c16555)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3826,10 +3826,7 @@ Names take meaning from:
 API naming *must* be done in easily readable US English.
 Keep in mind that most web developers aren't native English speakers.
 Whenever possible, names should be chosen that use common vocabulary
-a majority of English speakers are likely to understand when first encountering the name.
-Avoid abbreviations,
-except in cases where an abbreviation is extremely common and easy to understand.
-For example `UIEvent`, where "`UI`" stands for user interface. 
+a majority of English speakers are likely to understand when first encountering the name. 
 
 <div class="example">
 
@@ -3840,6 +3837,9 @@ than `cardinality`.
 </div>
 
 Value readability over brevity.
+Avoid abbreviations,
+except in cases where an abbreviation is extremely common and easy to understand.
+For example `UIEvent`, where "`UI`" stands for user interface.
 Keep in mind, however, that
 the shorter name is often the clearer one.
 For instance,


index.bs
```diff
@@ -2206,6 +2206,7 @@ Use strings as the values for constants or sets of enumerated values.
 
 Strings make it easier for developers to inspect values
 and read code that uses those values.
+As such, avoid abbreviations as they can make code harder to read.
```
martinthomson: Can we not just cite https://w3ctag.github.io/design-principles/#naming-is-hard ?

If that doesn't mention abbreviation, then it can be updated to do that.

index.bs
```diff
@@ -3798,6 +3798,8 @@ API naming *must* be done in easily readable US English.
 Keep in mind that most web developers aren't native English speakers.
 Whenever possible, names should be chosen that use common vocabulary
 a majority of English speakers are likely to understand when first encountering the name.
+Abbreviations should be avoided,
+except in cases where an abbreviation is canonicaland easier to understand.
```
```suggestion
except in cases where an abbreviation is canonical and easier to understand.
```

index.bs
```diff
@@ -3798,6 +3798,8 @@ API naming *must* be done in easily readable US English.
 Keep in mind that most web developers aren't native English speakers.
 Whenever possible, names should be chosen that use common vocabulary
 a majority of English speakers are likely to understand when first encountering the name.
+Abbreviations should be avoided,
+except in cases where an abbreviation is canonical and easier to understand.
```
```suggestion
except in cases where an abbreviation is extremely common and easier to understand.
For example `UIEvent`, where "`UI`" stands for user interface. 
```

index.bs
```diff
@@ -3798,6 +3798,9 @@ API naming *must* be done in easily readable US English.
 Keep in mind that most web developers aren't native English speakers.
 Whenever possible, names should be chosen that use common vocabulary
 a majority of English speakers are likely to understand when first encountering the name.
+Abbreviations should be avoided,
+except in cases where an abbreviation is extremely common and easier to understand.
```
```suggestion
except in cases where an abbreviation is extremely common and easy to understand.
```

index.bs
```diff
@@ -2205,7 +2205,7 @@ is a request, rather than a guarantee.
 Use strings as the values for constants or sets of enumerated values.
 
 Strings make it easier for developers to inspect values
-and read code that uses those values.
+and read code that uses those values (see [[#naming-is-hard]]).
```
jyasskin: I think this fits better in the first paragraph:

```suggestion
and read code that uses those values.
```

index.bs
```diff
@@ -2205,7 +2205,7 @@ is a request, rather than a guarantee.
 Use strings as the values for constants or sets of enumerated values.
```
jyasskin, line 2201:
```suggestion
Use strings as the values for constants or sets of enumerated values.
See [[#naming-is-hard]] when choosing strings.
```

index.bs
```diff
@@ -3798,6 +3798,9 @@ API naming *must* be done in easily readable US English.
 Keep in mind that most web developers aren't native English speakers.
 Whenever possible, names should be chosen that use common vocabulary
 a majority of English speakers are likely to understand when first encountering the name.
+Abbreviations should be avoided,
```
jyasskin: https://github.com/w3ctag/process/blob/main/style-guide.md#use-the-active-voice

```suggestion
Avoid abbreviations,
```

index.bs
```diff
@@ -3798,6 +3798,9 @@ API naming *must* be done in easily readable US English.
...
+except in cases where an abbreviation is extremely common and easy to understand.
+For example `UIEvent`, where "`UI`" stands for user interface.
```
jyasskin: This is an example of "Value readability over brevity" in the next non-example paragraph. How about

> Value readability over brevity.
> Often this means to avoid abbreviations.
> Keep in mind, however, that

You could add the "UI" example into that example. It might be good to also have an example of when the abbreviation was a mistake.

index.bs
```diff
@@ -2203,6 +2203,8 @@ is a request, rather than a guarantee.
 <h3 id="string-constants">Use strings for constants and enums</h3>
 
 Use strings as the values for constants or sets of enumerated values.
+See [[#naming-is-hard]] when choosing strings.
+See [[#naming-is-hard]] when choosing strings.
```
```suggestion
```

jyasskin (approved):
This would fit better into the "Value readability over brevity." paragraph and the example that follows it, but this is good enough for me.

index.bs
marcoscaceres: I think `Intl` namespace was a mistake from ECMA was a mistake. Would other folks here agree?

PR #612, opened 2026-02-06
Disambiguate when to add global event handlers

### LINKED ISSUE #611: Clarify guidance for onfoo attributes
https://w3ctag.github.io/design-principles/#always-add-event-handlers says to add an onfoo attribute to the event handler of the element.

I am interpreting it such that if I have an HTMLFooElement that has an onfoo handler, I should add onfoo to `interface HTMLFooElement`.

However, https://github.com/WICG/fenced-frame/pull/144#discussion_r1560200386 seems to say that such event handlers should instead be added to the global event handler list (`interface mixin GlobalEventHandlers`, at least in blink).

I am not sure if I am misreading the guidance and/or the discussion thread but it would be great to clarify what is the right thing to do here.

cc'ing some people from an internal discussion and from that thread:
@domfarolino @domenic @dbaron

domfarolino:
Fixes https://github.com/w3ctag/design-principles/issues/611. I think when new events are introduced with the intention of being fired at HTML **or** SVG elements, as opposed to just generic non-DOM-associated EventTargets, the guidance is to add them to the [`GlobalEventHandlers`](https://html.spec.whatwg.org/multipage/webappapis.html#globaleventhandlers) mixin.

As https://github.com/w3ctag/design-principles/issues/611 notes, this is the advice I received by Domenic in https://github.com/WICG/fenced-frame/pull/144#discussion_r1560200386, and I think there is slightly ambiguous wording in the design principles that may mistakenly lead people away from this advice when designing new APIs that introduce new events aimed at HTML elements. See https://github.com/w3ctag/design-principles/issues/611#issuecomment-3862411678.

***
<a href="https://pr-preview.s3.amazonaws.com/domfarolino/design-principles/pull/612.html" title="Last updated on Feb 11, 2026, 2:14 PM UTC (c206262)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/612/31952d9...domfarolino:c206262.html" title="Last updated on Feb 11, 2026, 2:14 PM UTC (c206262)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2550,7 +2550,7 @@ it's important to continue to define event handler IDL attributes because:
 * they enable feature-detection for the supported events (see [[#feature-detect]])
 
 <p class="note">
-For consistency, if the event needs to be handled by HTML or SVG elements,
+For consistency, if the event needs to be handled by HTML, SVG, or MathML elements,
 add the [=event handler IDL attributes=]
 on the {{GlobalEventHandlers}} interface mixin,
 instead of directly on the relevant element interface(s).
--- a/index.bs
+++ b/index.bs
@@ -2550,7 +2550,7 @@ it's important to continue to define event handler IDL attributes because:
 * they enable feature-detection for the supported events (see [[#feature-detect]])
 
 <p class="note">
-For consistency, if the event needs to be handled by HTML and SVG elements,
+For consistency, if the event needs to be handled by HTML or SVG elements,
 add the [=event handler IDL attributes=]
 on the {{GlobalEventHandlers}} interface mixin,
 instead of directly on the relevant element interface(s).


domfarolino:
@zcorpan @annevk Can you confirm that this is right? In particular, https://github.com/w3ctag/design-principles/issues/611#issuecomment-3862411678.

annevk (commented):
I'm not sure that we necessarily want all SVG events to end up on that list, but maybe? The other direction does seem like it's usually desired.

cc @foolip

jyasskin (approved):

index.bs
```diff
@@ -2550,7 +2550,7 @@ it's important to continue to define event handler IDL attributes because:
 * they enable feature-detection for the supported events (see [[#feature-detect]])
 
 <p class="note">
-For consistency, if the event needs to be handled by HTML and SVG elements,
+For consistency, if the event needs to be handled by HTML or SVG elements,
```
jyasskin: @bkardell suggests:

```suggestion
For consistency, if the event needs to be handled by HTML, SVG, or MathML elements,
```

Which makes sense since events can propagate through MathML and so be caught there.

domfarolino:
That suggestion works for me @jyasskin, thanks! Btw @annevk, can you think of any specific guidance we could add (maybe in a follow-up) for when it would make sense to add element-specific handlers, instead of adding global ones?

annevk:
Element-specific seems almost never justified (due to bubbling), but I don't know how many events SVG and MathML define and whether we want to have those on the global list.

marcoscaceres:
I think @bkardell said none for MathML. I’m unsure about svg. Maybe we should take out mentions of those.

marcoscaceres:
Checked SVG’s IDL. It depends fully on html. https://www.w3.org/TR/SVG2/idl.html

marcoscaceres (approved):

bkardell:
MathML simply mixes in the global handlers: https://w3c.github.io/mathml-core/#dom-and-javascript

I just want to be sure that to the greatest extent possible if we ever say "SVG" we consider MathML too :)

PR #616, opened 2026-06-02
Create an "accessible by default" principle.

jyasskin:
The "Media Independence" and "[CSS]Content should be viewable and accessible by default" principles expressed specific applications of this goal, so I merged them.

@dbaron @atanassov, this heavily modifies sections you added, so I'd appreciate a check that I haven't lost anything essential.

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/pull/616.html#accessible-by-default" title="Last updated on Jul 8, 2026, 5:12 PM UTC (c95949e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/616/41fee41...c95949e.html" title="Last updated on Jul 8, 2026, 5:12 PM UTC (c95949e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -303,7 +303,7 @@ An inattentive developer should, by default,
 produce content that the widest possible set of users can interact with,
 whether or not they have the [[ethical-web-principles#multi|devices, software]],
 or [[ethical-web-principles#allpeople|abilities]] that the developer tested against.
-That is, the simplest ways to use HTML, CSS, and Javascript should produce content that is
+That is, the simplest ways to use HTML, CSS, and JavaScript should produce content that is
 [[wcag22#perceivable|perceivable]], [[wcag22#operable|operable]], and [[wcag22#robust|robust]].
 
 One of the main values of the Web is that it's extremely flexible:
--- a/index.bs
+++ b/index.bs
@@ -296,20 +296,40 @@ necessary.
 Use short-lived, temporary identifiers
 unless a persistent identifier is absolutely necessary.
 
-<h3 id="devices-platforms">Support the full range of devices and platforms (Media Independence)</h3>
-<!-- was "Media Independence" in the HTML Design Principles -->
+<h3 oldids="css-content-should-be-visible,devices-platforms"
+    id="accessible-by-default">Content should be accessible by default</h3>
 
-As much as possible,
-ensure that features on the web work across different input and output [devices,
-screen sizes, interaction modes, platforms, and media](https://www.w3.org/TR/ethical-web-principles/#multi).
+An inattentive developer should, by default,
+produce content that the widest possible set of users can interact with,
+whether or not they have the [[ethical-web-principles#multi|devices, software]],
+or [[ethical-web-principles#allpeople|abilities]] that the developer tested against.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -296,20 +296,40 @@ necessary.
...
+    id="accessible-by-default">Content should be accessible by default</h3>
 
-As much as possible,
-ensure that features on the web work across different input and output [devices,
-screen sizes, interaction modes, platforms, and media](https://www.w3.org/TR/ethical-web-principles/#multi).
+An inattentive developer should, by default,
+produce content that the widest possible set of users can interact with,
+whether or not they have the [[ethical-web-principles#multi|devices, software]],
+or [[ethical-web-principles#allpeople|abilities]] that the developer tested against.
+That is, the simplest ways to use HTML, CSS, and Javascript should produce content that is
```
matatk:
```suggestion
That is, the simplest ways to use HTML, CSS, and JavaScript should produce content that is
```

matatk (changes requested):
Reads very well to me; thanks @jyasskin!

My only change was to use what appears to be the standard capitalisation for JavaScript.

index.bs
```diff
@@ -296,20 +296,40 @@ necessary.
...
+and accessibility of the web platform,
+even when used by hasty developers.
+
+It should take sophistication and non-default attributes, properties, and arguments
+to create pages that exclude some users,
+rather than requiring sophistication to include everyone.
+
+<div class="example">
+For example, the default behavior of all layout systems in CSS will not lead to content being clipped,
+content overlapping other content, or content being unreachable by scrolling.
```
dbaron, line 328: I think it might be worth mentioning possible causes of such problems, which include both variation in screen size and variation in user settings like font sizes.  (That was more explicit in one of the two examples, the block/flex/grid one, that you merged into this one.)

dbaron (commented):
Two thoughts here (along with one other specific comment below):

One, I'm a little worried about the naming of "accessible by default" given the (typically narrower) way the word "accessibility" is generally used in the web technology space.  I'm not sure if I have any great answers here.

Second, given the breadth of the principle, I think it would be good to come up with some additional examples.  Ideally it would be nice to have a good example about designing for the (narrower definition of) accessibility by default, and maybe about something like designing for operating system independence by default.  I don't have any great ones to offer right now, though.

index.bs
bkardell: @dbaron could you suggest how you'd prefer to fix this?

index.bs
```diff
@@ -297,20 +297,40 @@ necessary.
 Use short-lived, temporary identifiers
 unless a persistent identifier is absolutely necessary.
 
-<h3 id="devices-platforms">Support the full range of devices and platforms (Media Independence)</h3>
-<!-- was "Media Independence" in the HTML Design Principles -->
+<h3 oldids="css-content-should-be-visible,devices-platforms"
+    id="accessible-by-default">Content should be accessible by default</h3>
```
jyasskin, line 301: @dbaron suggested perhaps:

```suggestion
    id="accessible-by-default">Content should be accessible and widely usable by default</h3>
```

I need to also go through and see if I need to propagate that anywhere else in the patch.

PR #621, opened 2026-08-19
Add new fragment directive guidance

tabatkins:
I'm proposing a new guidance, to always use ["fragment directives"](https://wicg.github.io/scroll-to-text-fragment/#the-fragment-directive) when defining new things that go in the fragment of a URL.

In [CSS Link Params](https://drafts.csswg.org/css-link-params/) I initally copied SVG's fragment directives when defining how to pass link Params directly in a URL, but it was pointed out in  <https://github.com/w3c/csswg-drafts/issues/14235> that I should instead be relying on fragment directives, as defined by the [Scroll To Text Fragment spec](https://wicg.github.io/scroll-to-text-fragment/).

The fragment directive concept was designed with careful attention to compat and how it should be exposed to various DOM APIs, and is generally a much better structure than the general fragment, which is unstructured and heavily used by authors, which makes it hard to *impose* any structure.

Text fragments are the only currently-shipping feature using fragment directives, but link params will be the second, and we should encourage further use of this feature instead of letting people unknowingly repeat the legacy mistakes of SVG and Media Fragments.

***
<a href="https://pr-preview.s3.amazonaws.com/tabatkins/design-principles/pull/621.html" title="Last updated on Sep 14, 2026, 8:42 PM UTC (ec7e8b3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/design-principles/621/87ed909...tabatkins:ec7e8b3.html" title="Last updated on Sep 14, 2026, 8:42 PM UTC (ec7e8b3)">Diff</a>


index.bs
```diff
@@ -3544,6 +3547,34 @@ due to security implications, and instead recommend enforcing strict MIME types
 
 New MIME types should have a specification and should be registered with the Internet Assigned Numbers Authority (IANA).
 
+<h3 id="fragment-directives">Add new URL fragment features as directives</h3>
+
+When defining new URL features that should pass information via the fragment,
```
```suggestion
When defining new URL features that passes information via the fragment,
```

marcoscaceres:
@tabatkins, do you know what the plan is to move this stuff to, presumedly, the URL spec and HTML?

index.bs
```diff
@@ -3544,6 +3547,34 @@ due to security implications, and instead recommend enforcing strict MIME types
...
+define a [=fragment directive=] to do so,
+rather than inventing a novel syntax in the normal fragment segment.
+
+New URL features are appropriate to define in the fragment when they do something purely on the client
+after the response has been received,
```
marcoscaceres, line 3557: Not following this ... maybe this could be simplified?

index.bs
```diff
@@ -3544,6 +3547,34 @@ due to security implications, and instead recommend enforcing strict MIME types
...
+rather than something that should be communicated to the server with the request.
+For example, the traditional use of the URL fragment to indicate the ID of an element that should be scrolled to,
+or the use of [=text directives=]
+to indicate a segment of text that should be scrolled to.
```
marcoscaceres: How about: 

```suggestion
For example, the traditional use of the URL fragment to indicate the ID of an element that a user agent scrolls to,
or the use of [=text directives=]
to indicate a segment of text that a user agent scrolls to and highlights.
```

index.bs
dandclark: Tense in the above suggestion is wrong, how about:
```suggestion
When defining new URL features that pass information via the fragment,
```

index.bs
```diff
@@ -3544,6 +3547,34 @@ due to security implications, and instead recommend enforcing strict MIME types
...
+
+The "standard" fragment syntax space is already dedicated to the "ID to scroll to" task,
+and sometimes used by applications to pass additional information.
+It is not meaningfully structured,
+and can't have structure <em>applied</em> to it by the user agent,
+as this could interfere with existing author usage.
+[=Fragment directives=] instead give a separate syntax space,
+structured identically to query parameters,
+without interfering with the existing uses of the fragment.
+
```
marcoscaceres, line 3571: It might be good to have a note that developers sometimes (ab)use hash space to do things (e.g., "#5" might be slide 5 in a presentation)... so consideration must be given to not stomp on what developers might be doing...

Might be good to add a note around that.

marcoscaceres (commented):
This is great, @tabatkins! Thank you. Some suggestions.

index.bs
tabatkins: This paragraph is meant to be a general intro to what stuff goes in the fragment, as opposed to in the query or what-have-you.

index.bs
tabatkins: This is already alluded to by "and sometimes used by applications to pass additional information".

jyasskin:
> @tabatkins, do you know what the plan is to move this stuff to, presumedly, the URL spec and HTML?

https://github.com/whatwg/html/issues/8282 has the plan to merge into HTML and links to a PR.

index.bs
```diff
@@ -3545,6 +3548,34 @@ due to security implications, and instead recommend enforcing strict MIME types
...
+Note: Examples of specifications that do this already are [[scroll-to-text-fragment]] and [[css-link-params]].
+Examples of older features that violate this and instead define something in the standard fragment
+are <a href="https://www.w3.org/TR/SVG2/linking.html#SVGFragmentIdentifiersDefinitions">SVG Fragment Identifiers</a>
```
jyasskin, line 3575: We should probably mention [XPointer](https://www.w3.org/TR/xptr-framework/), which recommended syntax like

> #xpointer(id('boy-blue')/horn[1])element(boy-blue/3)

(Which is, I think, where SVG fragments came from?)

@domfarolino, since you're upstreaming fragment directives, or @bokand since you designed this in the first place, do you know if the XPointer syntax was evaluated and found to be web-incompatible, or if fragment directives are better for some other reason? The TAG can always say "browsers decided to go this way instead, and it's best to be consistent", but if there's a technical reason it's better, that'll head off some complaints.

